PR #55, opened 2014-12-02
update distro page text

glensc:
it does list any distro that was listed at least once, it does not do
this based on current mapping. so there's chance the distro is listed
but at the moment there are zero project mappings

### DIFF
--- a/anitya/templates/distros.html
+++ b/anitya/templates/distros.html
@@ -42,8 +42,7 @@
 
 <div class="row">
   <p>
-    Here is the list of all the distributions having at least one project
-    mapped monitored by anitya.
+    Here is the list of all the distributions that anitya knows about.
   </p>
   {% if is_admin %}
   <p>


pypingou:
@glensc  I realized today that you did not sign the FPCA in the Fedora account system, could you please confirm that you gave us permission to use all your changes under the same license terms as the rest of the code in the project ?

Thanks

glensc:
sure, can you give me link (or pointers) where to sign FPCA?

pypingou:
You can sign the FPCA if you create a Fedora account at: https://admin.fedoraproject.org/accounts

But just saying here that your contributions are made under the same license as the project is also fine (and saves you from creating this account if you don't want to).

glensc:
I think i did not feel comfortable providing my telephone number, that's why it was not signed before. Filled it now.

pypingou:
This information is private and basically can ever be used by admins to contact you in case of (I guess legal, can't think of another reason) emergency.

PR #72, opened 2015-01-09
add url variable to how to test

glensc:
this makes it easier to test with own url:
write line of url yourself, and c&p the rest


pypingou:
Cool, thanks!

PR #74, opened 2015-01-09
Add api to query for a project by homepage.

ralphbean:
I'm thinking that we can use this when a new package gets added to Fedora.  We
can look up its upstream_url in pkgdb, and then query anitya to see if any
projects are registered with that homepage.  If there are not, then add one
with this new package if its backend is an easy one (pypi, rubygems, etc).

### DIFF
--- a/anitya/api.py
+++ b/anitya/api.py
@@ -106,9 +106,20 @@ def api_projects():
 
         /api/projects/?pattern=py*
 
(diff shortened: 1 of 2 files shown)


pypingou:
Sounds good :+1:

ralphbean:
Thanks @pypingou!

PR #83, opened 2015-02-20
Remove non-working script tags

tyll:
Remove script tags with sources to non-existent files. JQuery is already
included in master template.

### DIFF
--- a/anitya/templates/index.html
+++ b/anitya/templates/index.html
@@ -43,12 +43,6 @@
 {% endblock %}
 
 {% block jscript %}
-<script type="text/javascript"
-    src="{{ url_for('static', filename='jquery-1.10.2.js') }}">
-</script>
-<script type="text/javascript"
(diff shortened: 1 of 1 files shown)


ralphbean:
How about jQuery-UI?  Is it already included.

I'm not sure what we actually rely on and will wait for @pypingou to comment.

tyll:
On Fri, Feb 20, 2015 at 03:47:41PM -0800, Ralph Bean wrote:
> How about jQuery-UI?  Is it already included.
> 
> I'm not sure what we actually rely on and will wait for @pypingou to comment.

It is not in the repo and also not deployed on release-monitoring.org.
But if it is needed, the path needs to be adjusted to include js/, if it
should match the jquery javascript file that is already used.

ralphbean:
> It is not in the repo and also not deployed on release-monitoring.org

Sounds good to me.  :+1: here.

pypingou:
Thanks!

PR #89, opened 2015-02-22
Autocomplete distribution name when mapping a project

pypingou:
Fixes https://github.com/fedora-infra/anitya/issues/79


ralphbean:
:+1:

PR #90, opened 2015-02-23
Check existing projects while adding a new one

pypingou:
Fixes https://github.com/fedora-infra/anitya/issues/85

### DIFF
--- a/anitya/templates/project_new.html
+++ b/anitya/templates/project_new.html
@@ -119,7 +119,7 @@
 
   };
 
-  function show_project(_url) {
+  function show_project(_url, type) {
     $.ajax({
         url: _url,
         type: 'GET',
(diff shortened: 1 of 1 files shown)


puiterwijk:
Could you maybe put the html already in the template but then have it enabled (unhidden) by javascript, rather than creating the entire block in JS itself?
Makes it easier to modify the template I think.

:+1:.

anitya/templates/project_new.html
```diff
@@ -115,11 +125,52 @@
 
   };
 
+  function show_project(_url, type) {
```
puiterwijk: Maybe rename to show_similar_project?

pypingou: sure why not, but it's not used anywhere else than this template/page

puiterwijk:
Looks good to me. :shipit:

PR #91, opened 2015-02-23
Check now button available after an edit

pypingou:


pypingou:
Fixes https://github.com/fedora-infra/anitya/issues/65 as well

ralphbean:
Will this allow me to look at any project and manually tack ``?justedit=True`` in the url in order to get an edit button?

pypingou:
No the value is set in the session after having press 'Update' in the edit page

ralphbean:
Well, :+1: then.

PR #92, opened 2015-02-24
More information for the custom backend

pypingou:
Fixes https://github.com/fedora-infra/anitya/issues/71

### DIFF
--- a/anitya/static/css/cnucnu.css
+++ b/anitya/static/css/cnucnu.css
@@ -25,3 +25,7 @@
     padding-left: 2px;
     padding-right: 2px;
 }
+
+table input[type="text"] {
+    width: 300px;
+}
--- a/anitya/templates/project_new.html
+++ b/anitya/templates/project_new.html
@@ -6,11 +6,11 @@
 <div class="page-header">
   <h1>{{ context }} project</h1>
 </div>
-{% if context == 'Add' %}
+{% if context == 'Add' -%}
 <form method="POST" action="{{ url_for('new_project') }}" >
-{% elif context == 'Edit' %}
+{%- elif context == 'Edit' -%}
 <form method="POST" action="{{ url_for('edit_project', project_id=project.id) }}" >
-{% endif %}
+{%- endif %}
   <div class="row">
     <div class="col-md-6">
       <p> {{ context }} your project to anitya</p>
@@ -37,7 +37,7 @@
               <td id="example_txt"></td>
             </tr>
 
-            {% if context == 'Add' %}
+            {%- if context == 'Add' -%}
             <tr>
               {{ render_field_in_row(form.distro) }}
               <td></td>
@@ -46,7 +46,7 @@
               {{ render_field_in_row(form.package_name) }}
               <td></td>
             </tr>
-            {% endif %}
+            {%- endif -%}
           </table>
         </span>
       </div>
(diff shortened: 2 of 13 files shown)


tests/test_flask_admin.py
```diff
@@ -82,6 +82,7 @@ def test_add_distro(self):
                 sess['email'] = 'pingou@pingoured.fr'
 
             output = c.get('/distro/add', follow_redirects=True)
+            print output.data
```
puiterwijk: Is this something left over from testing, or should this now always be in the tests?

pypingou: Left over from testing indeed, but not a problem when ran with ``runtests.sh`` (cuts stdout unless the test breaks)

anitya/templates/project_new.html
```diff
@@ -119,10 +121,19 @@
     if (str == 'custom'){
       $('#regex_row').show();
       $('#version_url_row').show();
+      $('#regex').attr('placeholder', 'Leave empty for default regex');
     };
 
     $('#example_txt').html(examples[str]);
 
+    if (str == 'custom'){
```
puiterwijk: How about just appending this to the examples['custom'] string itself?

pypingou: because they are meant to be a list of projects for which the backend applies

anitya/templates/project_new.html
```diff
@@ -119,10 +121,19 @@
...
+      $('#example_txt').append(
+        '<br /> More information at <a href="https://release-monitoring.org'
```
puiterwijk: Maybe use url_for, so that people get local links if they have their own setup of Anitya?

anitya/templates/project_new.html
```diff
@@ -119,10 +121,19 @@
     if (str == 'custom'){
```
puiterwijk: I think this (and the if's above) should all be in the plugin code themselves.
As it stands now, you have a modular setup, but then still special-case all plugins in the core.

anitya/templates/project_new.html
```diff
@@ -119,10 +121,19 @@
...
+      $('#regex').attr('placeholder', 'Leave empty for default regex');
     };
 
     $('#example_txt').html(examples[str]);
 
+    if (str == 'custom'){
+      $('#example_txt').append(
+        '<br /> More information in the '
+        + '<a href="{{ url_for("about") }}#test-you-regex">'
+        + 'about#test-you-regex</a>'
```
puiterwijk: Maybe also make this use url_for?

anitya/static/css/cnucnu.css
```diff
@@ -25,3 +25,7 @@
     padding-left: 2px;
     padding-right: 2px;
 }
+
+table input[type="text"] {
```
puiterwijk: Is this also related to the more information change, or is this something else?

anitya/templates/project_new.html
pypingou: ?

anitya/static/css/cnucnu.css
pypingou: As the commit says, it makes the input fields a little larger, allowing, for example, to see the placeholder entirely.

anitya/templates/project_new.html
puiterwijk: I thought you previously displayed the full URL. Never mind the comment.

pypingou: I did, but url_for will not show the full url, it will just give /about, which I thought was even worse than just about#...

puiterwijk: Right, that's true.

puiterwijk:
:+1:

pypingou:
Thanks !

PR #93, opened 2015-02-24
Fix the drupal backends

pypingou:


anitya/lib/backends/drupal6.py
```diff
@@ -59,9 +62,13 @@ def get_versions(cls, project):
...
         '''
+        name = project.name
+        if 'drupal6:' in name.lower():
+            name = name[len('drupal6:'):].strip()
+
         url_template = 'https://www.drupal.org/project/%(name)s'
 
-        url = url_template % {'name': project.name}
+        url = url_template % {'name': name}
         regex = REGEX % {'name': project.name}
```
ralphbean: How about this line?  Should it be just ``name`` as well?

anitya/lib/backends/drupal7.py
```diff
@@ -59,6 +62,10 @@ def get_versions(cls, project):
             when the versions cannot be retrieved correctly
 
         '''
+        name = project.name
+        if 'drupal7:' in name.lower():
+            name = name[len('drupal7:'):].strip()
+
         url_template = 'https://www.drupal.org/project/%(name)s'
 
         url = url_template % {'name': project.name}
```
ralphbean: And here?

anitya/lib/backends/drupal6.py
pypingou: Yes indeed ² :)

ralphbean:
:+1:

pypingou:
Thanks @ralphbean

PR #96, opened 2015-02-24
Allow users to browse their own log

pypingou:
Should fix https://github.com/fedora-infra/anitya/issues/94


ralphbean:
Easy.  :+1:

pypingou:
Thanks

PR #103, opened 2015-02-25
On postgresql '_' is a special character that we should escape

pypingou:
Escaping it brings back the default behavior allowing the search for
mod_* and have this return mod_perl, mod_wsgi & co

Fixes https://github.com/fedora-infra/anitya/issues/86


ralphbean:
:+1:  :)

PR #105, opened 2015-02-25
More info when mapping a project fails

pypingou:


ralphbean:
Can we provide a link to the other project so people can go inspect it?

pypingou:
I thought about this but it'll appear escaped.

We could turn off autoescape for flash messages I guess

pypingou:
Hm, looking further, this will be hard as the error message is generated in the backend library which has thus not access to flask :-/

ralphbean:
Oh well, then.  :+1:

ralphbean:
Much better.  Thanks!  :+1:

PR #106, opened 2015-02-25
Fix paginations

pypingou:


anitya/templates/logs.html
```diff
@@ -35,11 +35,13 @@
...
         <li>
             {% if page > 1%}
-              <a href="{{ url_for('browse_logs') }}?page={{page - 1}}">
+              <a href="{{ url_for('browse_logs', page=page-1, project=project,
+                user=user, from_date=from_date) }}">
                 «
               </a>
             {% else %}
-              <a href="{{ url_for('browse_logs') }}?page={{page}}">
+              <a href="{{ url_for('browse_logs', page=page, project=project,
```
ralphbean: Why even have a back button at all if ``page <= 1``?

anitya/templates/logs.html
```diff
@@ -49,11 +51,13 @@
...
         <li>
             {% if page < total_page %}
-              <a href="{{ url_for('browse_logs') }}?page={{page + 1}}">
+              <a href="{{ url_for('browse_logs', page=page+1, project=project,
+                user=user, from_date=from_date) }}">
                 »
               </a>
             {% else %}
-              <a href="{{ url_for('browse_logs') }}?page={{page}}">
+              <a href="{{ url_for('browse_logs', page=page, project=project,
```
ralphbean: Same here, why even both with a forward button if ``page >= total_page``?

anitya/templates/updates.html
```diff
@@ -24,13 +24,13 @@
...
                 url_for('projects_updated',
-                        status=status, name=name, log=log) }}?page={{page - 1}}">
+                        status=status, name=name, log=log, page=page-1) }}">
                 «
               </a>
             {% else %}
               <a href="{{
                 url_for('projects_updated',
-                        status=status, name=name, log=log) }}?page={{page}}">
+                        status=status, name=name, log=log, page=page) }}">
```
ralphbean: Same here.

anitya/templates/logs.html
pypingou: Looks nicer :-)

anitya/templates/logs.html
pypingou: and same answer :)
but for this one I could take it down since it won't move the other buttons, but I thought we might as well be consistent and keep them both.

ralphbean:
Fair enough, :+1:.

pypingou:
Instead of keeping redirecting to the same page, I'm thinking I could just disable them keeping the UI clear but dropping the infinite loop

ralphbean:
:+1: after the revisions.  Thanks Pierre!

Issue #107, opened 2015-03-02
casing in googlecode urls


keszybz:
I added https://release-monitoring.org/project/5737/ for BioNetGen, but the check fails because the url has lowercase bionetgen:
bad: https://code.google.com/p/BioNetGen/downloads/list?sort=releasedate
good: https://code.google.com/p/bionetgen/downloads/list?sort=releasedate

Upstream camelcases the tarball names, so the camelcased name seems to be the right one to use. It would be really nice if I could use the standard googlecode pattern, but allow the name to be lowercase.

pypingou:
So this is an easyfix in the google backend

cicku:
Just a friendly note, Google Code will be shutdown in 2016.

pypingou:
Hi Christopher, thanks for the heads up, do you have a link for this?


-------- Message d'origine --------
De : Christopher Meng <notifications@github.com> 
Date :13/03/2015  05:52  (GMT+01:00) 
À : fedora-infra/anitya <anitya@noreply.github.com> 
Cc : Pierre-Yves Chibon <pingou@pingoured.fr> 
Objet : Re: [anitya] casing in googlecode urls (#107) 

Just a friendly note, Google Code will be shutdown in 2016.

cicku:
http://google-opensource.blogspot.com/2015/03/farewell-to-google-code.html

devel list has already advised.

Anyway, about this issue, is it ok to just use lower() to handle the case like bionetgen?

pypingou:
@cicku good idea yes

Aayush-Kasurde:
I would like fix this issue, Can anyone please help to fix it ?

Aayush-Kasurde:
Please see my PR - https://github.com/fedora-infra/anitya/pull/113 and provide comments.

pypingou:
This has been fixed by #114

PR #113, opened 2015-03-17
Added fix for issue #107

Aayush-Kasurde:
Converted Google project name to lower case using lower()

### DIFF
--- a/anitya/lib/backends/google.py
+++ b/anitya/lib/backends/google.py
@@ -56,9 +56,10 @@ class GoogleBackend(BaseBackend):
             when the versions cannot be retrieved correctly
 
         '''
+	projectname = project.name.lower()
         url = 'http://code.google.com/p/%(name)s/downloads/list'\
-            '?sort=releasedate' % {'name': project.name}
+            '?sort=releasedate' % {'name': projectname}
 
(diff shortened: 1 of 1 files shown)


pypingou:
Hi,

Thanks for the fix, couple of remarks:

a/ This will not fix the use case of #107 because now it will just take the ``project.name.lower()`` so it's as if the user had entered in anytia the project name in lower case. 
It only needs fixing in the URL not for the regex.

b/ The commit message is good, but you just inverted.
The first line should be : ``Converted Google project name to lower case using lower()`` and lower down you would have ``Fixes issue #107``
Otherwise the first thing you see when reading the log of the project is ``Added fix for issue_107`` which is not really descriptive of what the change is (unless you know by heart every issue of the project).

Aayush-Kasurde:
a/ Could you please elaborate more ?

b/ Totally Agree. I will take care of this in future.

pypingou:
See in #107 (the first post, the original issue) what is a good and a bad URL and see how the sources are named ;-)

cicku:
It's better to squash into one commit.

pypingou:
And it would allow to adjust the commit message :)

Also, you could change only one line, no real need to keep the projectname variable :)

PR #117, opened 2015-03-23
Allow projects to make insecure http calls

pypingou:
The idea is to have a per project settings to allow insecure http call, ie calls made over https where the certificate does not match the hostname.

This should fix issues such as: https://github.com/fedora-infra/anitya/issues/70


ralphbean:
LGTM!

pypingou:
Thanks for the review @ralphbean

PR #118, opened 2015-03-24
Update the GNOME backend to rely on the cache.json if present

pypingou:
fixes https://github.com/fedora-infra/anitya/issues/54


anitya/lib/backends/gnome.py
```diff
@@ -58,7 +84,12 @@ def get_versions(cls, project):
...
         '''
-        url = 'https://download.gnome.org/sources/%(name)s/' % {
-            'name': project.name}
 
-        return get_versions_by_regex(url, REGEX, project)
+        output = []
+        try:
+            # First try to get the version by using the cache.json file
+            output = use_gnome_cache_json(project)
+        except Exception:
```
ralphbean: Can we log a warning here about the details of the exception?  Otherwise, if the cache/json stuff breaks, we'll never know.

ralphbean:
Other than the request for a log statement, this looks good.  :+1:

PR #119, opened 2015-03-25
Include in the fedmsg message if the new version found is odd or not

pypingou:
By odd I mean that the new version found appears to be older than the
last version we had (ie: upstream released a version 2 after having
released a version 3).

This should help fixing https://github.com/fedora-infra/anitya/issues/66


ralphbean:
:+1:

PR #121, opened 2015-03-26
Strip leading v from versions before we compare them.

### LINKED ISSUE #110: Problems when the version numbering format changes
I updated protobuf on release-monitoring.org to the new upstream at GitHub, and anitya seems to be having issues dealing with the fact that version numbers now have a "v" in front of them.

https://release-monitoring.org/project/3715/ -- `Something strange occured, we found that this project has released a version "v3.0.0-alpha-2" while we had the latest version at "2.5.0"`

ralphbean:
Kinda fixes #110.

### DIFF
--- a/anitya/lib/backends/__init__.py
+++ b/anitya/lib/backends/__init__.py
@@ -46,6 +46,11 @@ def upstream_cmp(v1, v2):
 
     """
 
+    # Strip leading 'v' characters; turn v1.0 into 1.0.
+    # https://github.com/fedora-infra/anitya/issues/110
+    v1 = v1.lstrip('v')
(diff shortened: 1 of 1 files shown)


pypingou:
I didn't really want to go this road because I thought we could use this to teach upstreams to be consistent in their release naming, but ok :+1:

ralphbean:
> teach upstreams to be consistent in their release naming

But projects which have already switched format cannot rewrite their release history.  Ansible, for instance, already has github tags in two different formats and I don't think we can argue that they should go back and rename the old ones.

Plus, teaching all upstreams something is probably a sisyphean task.  :wink:

Issue #43, opened 2014-11-25
project search exclude


glensc:
wanted to search for "ruby" package (the ruby language), the pattern was modified to be "ruby*"
and they gave me 11 pages of rubygems, tried exclude with minus, that gave nothing:

https://release-monitoring.org/projects/search/?pattern=ruby*+-rubygems

please implement some advanced filtering or exact match

pypingou:
Ok I have two approached possible for this ticket:

* https://github.com/fedora-infra/anitya/compare/exact_match
We remove doing ``pattern*`` search by default. This will work and do exact match by default but this also means we are changing the default behavior.

* https://github.com/fedora-infra/anitya/compare/rework_search
If the pattern starts with a `!`, then do an exact match search, keeping the current default behavior.

ralphbean:
This might prove to be an interesting third option:  http://www.postgresql.org/docs/8.3/static/textsearch-intro.html#TEXTSEARCH-MATCHING

pypingou:
Well, isn't the question here about being able to do exact match search rather than text search?

ralphbean:
Yes.  But textsearch might be a more flexible method for specifying searches (and it might return results in a nicer order).

pypingou:
Ever played with textsearch? In sqlalchemy?

There seems to be some support: http://docs.sqlalchemy.org/en/latest/dialects/postgresql.html#full-text-search

pypingou:
So I guess there are two issues here:
* improve the search by using textsearch (``table.c.column.match(pattern)`` in sqlalchemy)
* allow perfect match search

pypingou:
I just did some testing of using ``match`` vs ``ilike`` when doing the search:

````
- Packages.package_name.match('python')
- Project.name.match('python')
  -> 451 projects found

- Packages.package_name.ilike('python%')
- Project.name.ilike('python%')
  -> 454 projects found
  
- Packages.package_name.ilike('python')
- Project.name.ilike('python')
  -> 454 projects found
````

ralphbean:
What is the first result?  is it 'python'?  :)

pypingou:
In both case it's not (after removing the order_by constraint)

ralphbean:
Bummer.  I was hoping textsearch would provide the "best" match first like xapian does.

PR #126, opened 2015-03-31
Run exact match search by default but add button to extend it

pypingou:
By default anitya will now do exact match searches on project and package name.
If there is only one result, the user will be redirected to its page.

If there are more than one result, the user will be presented with a button entitled ``Extend`` offering the possibility to extend the search.

### DIFF
--- a/anitya/lib/backends/custom.py
+++ b/anitya/lib/backends/custom.py
@@ -68,7 +68,7 @@ class CustomBackend(BaseBackend):
         if project.regex:
             regex = REGEX_ALIASES.get(project.regex, project.regex)
 
-        if '%{name}' in regex:
+        if '%(name)' in regex:
             regex = regex % {'name': project.name.replace('+', '\+')}
 
         return get_versions_by_regex(
--- a/anitya/lib/model.py
+++ b/anitya/lib/model.py
@@ -528,17 +528,31 @@ class Project(BASE):
 
         query1 = session.query(
             cls
-        ).filter(
-            Project.name.ilike(pattern)
         )
+
+        if '%' in pattern:
+            query1 = query1.filter(
+                Project.name.ilike(pattern)
+            )
+        else:
+            query1 = query1.filter(
+                Project.name == pattern
+            )
 
         query2 = session.query(
             cls
         ).filter(
             Project.id == Packages.project_id
-        ).filter(
-            Packages.package_name.ilike(pattern)
         )
+
+        if '%' in pattern:
+            query2 = query2.filter(
+                Packages.package_name.ilike(pattern)
+            )
+        else:
+            query2 = query2.filter(
+                Packages.package_name == pattern
+            )
 
         if distro is not None:
             query1 = query1.filter(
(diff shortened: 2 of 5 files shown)


pypingou:
Fixes https://github.com/fedora-infra/anitya/issues/43

ralphbean:
> If there are more than one result, the user will be presented with a button entitled Extend offering the possibility to extend the search.

If you know that there is more than one result, then we have already done the search.  I don't understand why we don't just include those results in the first page instead of asking the user to do a second operation and click 'Extend'.

ralphbean:
I can't think of another UI that has a search field that returns only exact matches by default, but then requires you to click an extend button to resubmit the search... to actually search.

I had a proposal in IRC to implement this by: making two subsequent queries -- one for exact matches, and one by a pattern -- and to return those two lists concatenated together as the results.  It covers all use cases I can think of with simpler interactions on the user's part.  Is there something wrong with this approach?

pypingou:
> If you know that there is more than one result, then we have already done the search. I don't 
> understand why we don't just include those results in the first page instead of asking the user to do a 
> second operation and click 'Extend'.

Search for ``arr`` might return you several results, but you will get even more if you search for ``arr*`` or ``*arr*``.

pypingou:
> I had a proposal in IRC to implement this by: making two subsequent queries -- one for exact matches, and one by a pattern -- and to return those two lists concatenated together as the results. It covers all use cases I can think of with simpler interactions on the user's part. Is there something wrong with this approach?

Well it does not solve the original question from @glensc on #43. It makes it easier to get the results but not able to do exact match.

ralphbean:
He didn't say that he wanted to be redirect directly to what he wanted.  He only said that he
> wanted to search for "ruby" package (the ruby language)

It does solve the problem because it puts the exact match *first* in the list where it is obvious and easy to click on.  If a user searches for 'nethack', then the nethack package ought show up first.. but it seems incorrect to me to ask the user to click an Extend button to also get the 'nethack-vultures' package to show up.  If they knew very precisely that they wanted to go directly to the 'nethack' package, they could have just typed it into the URL and not bothered searching to begin with.

Furthermore, the logic as currently implemented in this changeset where " If there is only one result, the user will be redirected to its page." makes it odd for novice users to search for *all* things with "ruby".  They'll go to search and type 'ruby' and then be redirected away from the search interface and over to the project for the ruby core project.  They'll have to know about the extended syntax.. go back to the search interface and type it again with "*ruby*".

If I go to most any other search interface out there on the internet (google, fedora-packages, etc..) and type 'ruby' I don't get back things that only exactly match 'ruby'.  I get *first* things that exactly match 'ruby' but then also things later in the results that have the string 'ruby' in them somewhere (i.e., rubygems-this and rubygems-that).

The approach as implemented here *exposes implementation details* about the underlying database to the user and asks them to adapt their behavior to that layer, instead of providing a search interface that is more familiar and has more in common with most other search interfaces.

ralphbean:
:+1:  Thanks Pierre!

pypingou:
Thanks @ralphbean

PR #128, opened 2015-04-01
When retrieving the latest version for the GNOME backend, try first cache.json

pypingou:
Assuming the odering of the version numbers on the cache.json file is always
correct, we first try it, then if we there are no cache.json, we get the list of
versions in a different way and rely on our sorting.

Should fix behavior such as:
https://bugzilla.redhat.com/show_bug.cgi?id=1034056#c6
where we went backward


pypingou:
I'm going to close this for now as I'm not quite fan of doing another http request for this and I think #129 might just have solved the problem we had here.

PR #129, opened 2015-04-01
Sort things consistently.

ralphbean:


pypingou:
:+1: for me.

The question I wonder is if this obsoletes #128 or not

ralphbean:
Well, #128 makes sense on its own anyways, no?

pypingou:
> Well, #128 makes sense on its own anyways, no?

Yes and no, it does extra http calls which are not needed if the sorting works (but are helpful if the sorting does not work).

Issue #131, opened 2015-04-04
github backend version check fails: version_url empty


piotr1212:
version check fails for projects with github backend, see carbon-c-relay for example:
https://release-monitoring.org/project/6115/
When I do a version check, the url https://github.com//tags is checked instead of https://github.com/grobian/carbon-c-relay/tags

In github.py it seems that the variable "version_url" is used to build the url. This variable is never set, there is no no box where this could be filled in.

pypingou:
So #134 fixes this, I'll try to cut a new release of anitya tonight when I get home.

PR #134, opened 2015-04-04
fix GitHub capitalization

piotr1212:
This should fix isse #131. It appears that commit a5a3ba878ff8f2f58b75f315fbb36d30ce33de85 broke github by changing the capitalizion in the github backend but not in all references to is.


pypingou:
Thanks!

PR #135, opened 2015-04-05
When querying a complete table with pagination, limit is always the set

pypingou:
Only the offset changes.

Fixes https://github.com/fedora-infra/anitya/issues/133

### DIFF
--- a/anitya/lib/model.py
+++ b/anitya/lib/model.py
@@ -167,7 +167,7 @@ class Distro(BASE):
                 page = None
 
         if page:
-            limit = page * 50
+            limit = 50
             offset = (page - 1) * 50
             query = query.offset(offset).limit(limit)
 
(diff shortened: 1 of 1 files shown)


anitya/lib/model.py
```diff
@@ -399,7 +399,7 @@ def all(cls, session, page=None, count=False):
                 page = None
 
         if page:
-            limit = page * 50
+            limit = 50
             offset = (page - 1) * 50
```
ralphbean: might as well make this ``offset = (page - 1) * limit`` to make it future-proof.

ralphbean:
This needs to be fixed in 4 or 5 other places in the file also.

Since that code is copied and pasted between all those methods, perhaps it can be reorganized into a single function?

ralphbean:
:+1:

pypingou:
Thanks for the review @ralphbean

PR #138, opened 2015-04-08
Adjust the default regex to ignore .asc in filename

pypingou:
These are used when tarball are GPG signed when released and can lead to
false positive in versions found.

Fixes https://github.com/fedora-infra/anitya/issues/136


laxathom:
:+1:

PR #140, opened 2015-04-11
Use vcr to mock out all http stuff in the tests.

### LINKED ISSUE #120: address the decay of the test suite
every time I go to fix an anitya bug, I try to run the tests and they are broken.  not because anitya is broken, but because the tests rely on querying actual upstreams about what versions are available... and those upstreams are always releasing new versions.

I went to try and write what would have been a five minute patch, but I've been fixing the test suite for 40 minutes now :(

----

So, we could do any number of things to fix this.  One of them is mocking.  We could mock ``requests.get()`` so that it returns a fake object with a lookup of all the pages we scrape **today** and we keep copies of all those pages in files in some test data directory.  The tests will then get a *real* copy of a web page that really existed, but it will be frozen in time and will never change.

ralphbean:
This should fix #120.

I'm in the process of packaging python-vcrpy for Fedora still.  There are a
couple of deps still to go.


pypingou:
Looking good, let's see what jenkins says with the proper deps installed :)

pypingou:
http://jenkins.cloud.fedoraproject.org/job/anitya/143/ looking good, :+1: for me!

PR #144, opened 2015-04-16
Feature/flag project

### LINKED ISSUE #87: Allow to propose projects to be deleted
I would like to be able to propose projects to be deleted for example after I accidently created a duplicate:
https://release-monitoring.org/project/5582/

oddshocks:
This should fix #87, and moooreee! :rainbow: 

- Adds 'Flag' button to project pages
  - Logged-in users can flag a project with a reason for doing so, such as "please delete this because blah blah blah"
- Adds an admin view to review, sort, and open/close flags


anitya/__init__.py
```diff
@@ -145,7 +145,8 @@ def log(session, project=None, distro=None, topic=None, message=None):
                              'project: %(project)s',
         'project.edit': '%(agent)s edited the project: %(project)s fields: '
                         '%(changes)s',
-
+        'project.flag': '%(agent)s flagged the project: %(project)s',
```
pypingou: Should we include the message in there?

anitya/admin.py
```diff
@@ -375,3 +375,116 @@ def browse_logs():
...
+        cnt_flags=cnt_flags,
+        total_page=total_page,
+        page=page,
+        project=project or '',
+        from_date=from_date or '',
+        user=user or ''
+    )
+
+
+@APP.route('/flags/<flag_id>/set/<state>', methods=['GET'])
```
pypingou: This should be ``POST`` not ``GET`` and have it CSRF protected otherwise we'll have a security hole :)

anitya/lib/__init__.py
```diff
@@ -262,3 +262,62 @@ def map_project(
...
+        message=dict(
+            agent=user_mail,
+            project=project.name,
+            reason=reason,
+        )
+    )
+    session.commit()
+    return project
+
+def set_flag_state(session, flag, state, user_mail):
```
pypingou: Missing an empty line :)

anitya/lib/__init__.py
```diff
@@ -262,3 +262,62 @@ def map_project(
...
+    """ Change the state of a ProjectFlag in the database.
+
+    """
+
+    flag.state = state
```
pypingou: What if the flag already had this state? Do we send another fedmsg?

anitya/lib/__init__.py
```diff
@@ -262,3 +262,62 @@ def map_project(
...
+            session,
+            topic='project.flag.set',
+            message=dict(
+                agent=user_mail,
+                flag=flag.id,
+                state=state,
+            )
+        )
+        session.add(flag)
+        session.commit()
```
pypingou: Should we send the fedmsg before saving the info to the DB?

anitya/lib/model.py
```diff
@@ -602,3 +602,129 @@ class ProjectVersion(BASE):
...
+
+    project_id = sa.Column(
+        sa.Integer,
+        sa.ForeignKey(
+            "projects.id",
+            ondelete="cascade",
+            onupdate="cascade")
+    )
+
+    project = sa.orm.relation('Project')
```
pypingou: I normally put the relation at the end of the table definition, first all the column, then the constraints, then the relations

anitya/lib/model.py
```diff
@@ -602,3 +602,129 @@ class ProjectVersion(BASE):
...
+            created_on=time.mktime(self.created_on.timetuple()),
+            updated_on=time.mktime(self.updated_on.timetuple()),
+        )
+        if detailed:
+            output['reason'] = self.reason
+
+        return output
+
+    @classmethod
+    def insert(cls, session, user, project, reason):
```
pypingou: Looks like this is not being used anywhere

anitya/lib/model.py
```diff
@@ -565,3 +565,129 @@ class ProjectVersion(BASE):
...
+        :kwarg count: a boolean to return the result of a COUNT query
+            if true, returns the data if false (default).
+
+        """
+        query = session.query(
+            cls
+        )
+
+        if project_name:
+            query = query.filter(cls.project_id == Project.id).filter(Project.name  == project_name)
```
pypingou: Does this line fit in 80 chars?

anitya/templates/flags.html
```diff
@@ -0,0 +1,149 @@
...
+  {% if refresh %}
+  <p>This page should refresh automatically every 5 seconds</p>
+  <a href="{{ url_for('browse_flags') }}?page={{page}}">
+    <button type="submit" class="btn btn-danger" id="checknow">
+      <span class="glyphicon glyphicon-remove-sign"></span>
+      Remove auto-refresh
+    </button>
+  </a>
+  {% else %}
+  <a href="{{ url_for('browse_flags') }}?page={{page}}&refresh=True">
```
pypingou: You cna pass ``page=page, refresh=True`` within the ``url_for()`` method that will add them automatically and nicely

anitya/templates/flags.html
```diff
@@ -0,0 +1,149 @@
...
+{% endblock %}
+
+
+{% block body %}
+
+<div class="page-header">
+  <h1>Flags</h1>
+  {% if refresh %}
+  <p>This page should refresh automatically every 5 seconds</p>
+  <a href="{{ url_for('browse_flags') }}?page={{page}}">
```
pypingou: See below :)

anitya/__init__.py
oddshocks: It's your choice. The reason might be kind of long, so I excluded it from the log message, figuring that an admin could check the reason from the Flags view itself. Do you think it'd be good to have it here, as well?

anitya/ui.py
```diff
@@ -430,6 +430,45 @@ def edit_project(project_id):
...
+            anitya.lib.flag_project(
+                SESSION,
+                project=project,
+                reason=form.reason.data,
+                user_mail=flask.g.auth.email,
+            )
+            flask.flash('Project flagged for admin review')
+            # TODO: what happens with justflag? Need to change something
+            # somewhere else for that to work?
+            flask.session['justflag'] = True
```
pypingou: What is this used for?

anitya/__init__.py
pypingou: I think I would like to have it in yes. It'll give the admin an idea what this flag is about when looking at the list of flags.

anitya/admin.py
oddshocks: My problem was that if I exclude `GET`, trying to toggle a flag state results in "The method is not allowed for the requested URL." :(

pypingou: Because you're using a link instead of a form, only forms can make POST requests :)

anitya/lib/__init__.py
oddshocks: Ahh, I see what you mean, I think. Because someone could just keep hitting the URL that sets the flag to the same state over and over again, right? I'll add a check for that now.

anitya/__init__.py
oddshocks: The admin can already see the reason when looking at the list of flags. The reason is listed there. I just excluded it from this line to avoid a large block of text when viewing the *logs*. Do you think the full reason should be listed on the Flags *and* Logs page? If so, I can fix that.

anitya/lib/model.py
oddshocks: I just added a json representation in case it's used somewhere in fedmsgs or if it would be needed in the future. Should I remove it?

anitya/ui.py
oddshocks: `justflag`? Apparently, nothing. I swear I took that from somewhere else (I didn't just make it up), but it looks like it's gone now. Either way, it doesn't do anything apparently so I've deleted both instances of it. :+1:

anitya/admin.py
oddshocks: So, do I have to switch to using a form, or...? What do I have to change to make POST work?

anitya/__init__.py
pypingou: Yes I think it would be useful.

As it is for admins only, we could still easily rework that later on :)

anitya/admin.py
pypingou: You have to use a form in the html and then mark this endpoint as accepting ``POST`` requests only.

anitya/lib/__init__.py
pypingou: yup :)

anitya/lib/model.py
pypingou: I meant the ``insert()`` method. It doesn't seem to be used anywhere.

The json representation will be useful for broadcasting fedmsg messages ;-)

anitya/admin.py
ralphbean: @pypingou by "form" here do you mean a ``<form>``?  or do you mean a ``wtform?``  Can you link to an example?

pypingou: Well, it should be ``<form...>`` in the html and probably ``wtform`` in the python code to do the csrf validation.

As for an example, the ``delete_distro`` endpoint is CSRF protected: https://github.com/fedora-infra/anitya/blob/master/anitya/admin.py#L104

oddshocks: OK, I'm still a bit confused.
- Do I need to wrap a `<form>` around the individual button, or the entire `for` loop in the template that renders X buttons?
- How do I set the form `action=` if there are many buttons that can have one of two actions? I assume this means I need to wrap a `<form>` around each button, answering my first question?
- Which part of that `delete_distro` function actually adds CSRF protection? `validate_on_submit`?
- If I need to use a form for these buttons to avoid a security hole, should I just use the already-existing (empty) `ConfirmationForm`, or something else?

pypingou: On Tue, Apr 21, 2015 at 02:15:52PM -0700, David Gay wrote:
>      * Do I need to wrap a <form> around the individual button, or the entire
>        for loop in the template that renders X buttons?

The simplest will be to wrap the form around the individual buttons, as
otherwise you will end-up with a form that has multiple submit button, which is
doable but a little more tricky.

>      * How do I set the form action= if there are many buttons that can have
>        one of two actions? I assume this means I need to wrap a <form> around
>        each button, answering my first question?

Pretty much :)
Otherwise you could go via a third endpoint which then redirects to the proper
one, or use a single endpoint to do the two actions depending on which submit
button was selected, but it's gets hairy.

>      * Which part of that delete_distro function actually adds CSRF
>        protection? validate_on_submit?

Yes, you add the ``{{ form.csrf_token }}`` in the html and then validate this
CSRF via the ``form.validate_on_submit()`` method.

>      * If I need to use a form for these buttons to avoid a security hole,
>        should I just use the already-existing (empty) ConfirmationForm, or
>        something else?

You can either create a new form (which gives you the added bonus of
validators), or use the ConfirmationForm that is empty and thus only provides
CSRF protection, nothing else.

oddshocks:
Alright, things seem to be working! Good to merge?

pypingou:
To be entirely feature complete, it needs some unit-tests :-)

I also woul like to give it a try locally to see if I missed anything in the review :)

oddshocks:
@pypingou, I've added a good set of tests for both viewing the `/flags/` view and attempting to set the flag state under various conditions. All of these tests appear to pass. Let me know what you think!

oddshocks:
Actually, there is one issue I'm seeing. In `test_flag_project`, we're getting 405's in these sort of blocks:
```
with anitya.app.APP.test_client() as c:                       
    with c.session_transaction() as sess:                     
        sess['openid'] = 'http://pingou.id.fedoraproject.org/'
        sess['fullname'] = 'Pierre-Yves C.'                   
        sess['nickname'] = 'pingou'                           
                                                              
    output = c.get('/flags/{0}/set/closed'.format(flag.id),   
                   follow_redirects=True)                     
                                                              
self.assertEqual(flag.state, 'closed')                        
```

I'm 99% sure this is because the CSRF protection makes it so you can't directly hit the URL. What do I have to change here to make this call in these tests?

pypingou:
I see a couple of things here:
* You're doing a GET request instead of a POST (cf ``c.get()``)
* You will need to retrieve the csrf token first and include it in the data sent to the URL, see for example: https://github.com/fedora-infra/anitya/blob/master/tests/test_flask_admin.py#L96-L115

oddshocks:
I took a whack at getting the CSRF token, though I'm not sure I did it right. I thought I'd have to do it for each of those `with anitya.app.APP.test_client() as c:` blocks, but when I run `nosetests`, I don't get a failure. :confused:

oddshocks:
I believe this is ready for final review.

pypingou:
@oddshocks github seems to say this PR conflicts with what's in the repo, could you see if you could rebase it?

anitya/lib/__init__.py
```diff
@@ -262,3 +262,70 @@ def map_project(
...
+
+
+def set_flag_state(session, flag, state, user_mail):
+    """ Change the state of a ProjectFlag in the database.
+
+    """
+
+    # Don't toggle the state or send a new fedmsg if the flag's
+    # state wouldn't actually be changed.
+    if flag.state == state: return
```
pypingou: What about raising an exception here instead?

anitya/lib/model.py
```diff
@@ -565,3 +565,117 @@ class ProjectVersion(BASE):
...
+    reason = sa.Column(sa.Text, nullable=False)
+    user = sa.Column(sa.String(200), index=True, nullable=False)
+    state = sa.Column(sa.String(50), default='open', nullable=False)
+    created_on = sa.Column(sa.DateTime, default=datetime.datetime.utcnow)
+    updated_on = sa.Column(sa.DateTime, server_default=sa.func.now(),
+                           onupdate=sa.func.current_timestamp())
+
+    project = sa.orm.relation('Project', backref='flags')
+
+    def __init__(self, user, project, reason):
```
pypingou: You can drop this one, it's handled directly by SQLAlchemy :)

anitya/templates/flags.html
```diff
@@ -0,0 +1,149 @@
...
+        </button>
+      </div>
+    </form>
+  </div>
+
+</div>
+
+
+<table id="flag_list">
+<th></th><th>State</th><th>Created</th><th>Project</th><th>Reason</th><th>User</th>
```
pypingou: Shouldn't this be in a row ``<tr></tr>`` ?

anitya/templates/project_flag.html
```diff
@@ -0,0 +1,41 @@
...
+{% block body %}
+
+<div class="page-header">
+  <h1>Flag project</h1>
+</div>
+<form method="POST" action="{{ url_for('flag_project', project_id=project.id) }}" >
+  <div class="row">
+    <div class="col-md-12">
+      <p> Flag a project for admin review. Use this form to request that the project
+          be deleted, or otherwise changed in a way that can't be done normally.</p>
```
pypingou: ``can't be done by regular user`` no?

anitya/templates/project_flag.html
```diff
@@ -0,0 +1,41 @@
...
+        </span>
+      </div>
+    </div>
+  </div>
+  <button type="submit" class="btn btn-success" id="submit_btn">
+    <span class="glyphicon glyphicon-plus"></span>
+    Submit
+  </button>
+  {%- if context == 'Add' -%}
+  <a href="{{ url_for('projects') }}">
```
pypingou: Looking at the change in the ``ui.py`` file it seems that the context is always ``Flag``, is this swtich still used?

tests/test_flask_admin.py
```diff
@@ -407,6 +408,162 @@ def test_browse_logs(self):
...
+            self.assertEqual(output.status_code, 200)
+            self.assertTrue('<h1>Flags</h1>' in output.data)
+            self.assertTrue('geany' in output.data)
+
+            output = c.get('/flags?page=abc&limit=def&from_date=ghi')
+            self.assertEqual(output.status_code, 200)
+            self.assertTrue('<h1>Flags</h1>' in output.data)
+            self.assertTrue('geany' in output.data)
+
+            output = c.get('/flags?from_date=%s' % datetime.date.today())
```
pypingou: Just curious, if the from_date is tomorrow, then geany does not appear, correct?

pypingou:
Code wise, I have a couple of minor questions but this is looking really good, thanks @oddshocks !!

anitya/lib/model.py
oddshocks: Sorry, drop which thing?

pypingou: The ``__init__`` method :)

tests/test_flask_admin.py
oddshocks: Added a test for this :smile:

anitya/lib/model.py
oddshocks: Ah -- alright. Only added one because other classes had one. :)

pypingou: On Wed, May 20, 2015 at 11:12:20PM -0700, David Gay wrote:
>    In anitya/lib/model.py:
> 
>  > +        sa.ForeignKey(
>  > +            "projects.id",
>  > +            ondelete="cascade",
>  > +            onupdate="cascade")
>  > +    )
>  > +
>  > +    reason = sa.Column(sa.Text, nullable=False)
>  > +    user = sa.Column(sa.String(200), index=True, nullable=False)
>  > +    state = sa.Column(sa.String(50), default='open', nullable=False)
>  > +    created_on = sa.Column(sa.DateTime, default=datetime.datetime.utcnow)
>  > +    updated_on = sa.Column(sa.DateTime, server_default=sa.func.now(),
>  > +                           onupdate=sa.func.current_timestamp())
>  > +
>  > +    project = sa.orm.relation('Project', backref='flags')
>  > +
>  > +    def __init__(self, user, project, reason):
> 
>    Ah -- alright. Only added one because other classes had one. :)

Ah, then we should consider if we want to drop them as well, but maybe in
another PR :)

oddshocks:
This should be good to finally merge, with final approval.

pypingou:
I just tried to run the branch and I got:
```
  File ".../anitya/anitya/admin.py", line 447, in browse_flags
    total_page = int(ceil(cnt_flags / float(limit)))
UnboundLocalError: local variable 'cnt_flags' referenced before assignment
```

pypingou:
Second one:
```
  File ".../anitya/anitya/ui.py", line 469, in flag_project
    plugins=plugins,
NameError: global name 'plugins' is not defined
```

:-)

pypingou:
Nice job! :+1: for merging :)

PR #145, opened 2015-04-28
Fix mapping of a project on an existing package

pypingou:
Currently, when a mapping is removed, we do not delete the package as
there might be other project mapping to it (for example one packages
built from several sources).
This means that we can have packages not linked to any project.
So we should support mapping a project to this package and this is what
this commit is doing.


ralphbean:
Cool.  :+1:

PR #162, opened 2015-05-26
Use versions API.

voxik:
This should improve performance a bit, since just version and nothing
else is returned.

http://guides.rubygems.org/rubygems-org-api/#get---apiv1versionsgem-namelatestjson

Please note that I have not tested this, so hopefully this does not break anything ;)


pypingou:
Cool, looks good, :+1:

PR #163, opened 2015-05-26
Fix the unit-tests suite

pypingou:


pypingou:
Jenkins seems to agree with these changes: http://jenkins.cloud.fedoraproject.org/job/anitya/188/

tests/test_flask.py
```diff
@@ -240,7 +240,8 @@ def test_projects_search(self):
             '/projects/search/geany*', follow_redirects=True)
         self.assertEqual(output.status_code, 200)
 
-        expected = '<li class="message">Only one result matching with an '\
+        expected = '<li class="list-group-item list-group-item-default">' \
+            'Only one result matching with an '\
```
oddshocks: I don't know if you care about the whitespace here (like between the `\` and the ending `'`). I noticed that it varied.

tests/test_flask_admin.py
```diff
@@ -82,7 +84,7 @@ def test_add_distro(self):
                 sess['nickname'] = 'pingou'
                 sess['email'] = 'pingou@pingoured.fr'
 
-            output = c.get('/distro/add', follow_redirects=True)
```
oddshocks: Again, just curious: why is this being removed? I guess I'm not sure how `follow_redirects` works in this case. What would the redirect be that we're trying to avoid?

tests/test_flask_admin.py
```diff
@@ -543,13 +558,13 @@ def test_flag_project(self):
...
-            data = {}
-
+            # Get a new CSRF Token
+            output = c.get('/distro/add')
             csrf_token = output.data.split(
                 'name="csrf_token" type="hidden" value="')[1].split('">')[0]
 
-            data['csrf_token'] = csrf_token
+            # Grab the CSRF token again so we can toggle the flag again
+            data = {'csrf_token': csrf_token}
```
oddshocks: I like this better ;)

oddshocks:
Just a few comments (mostly curious inquiries). The code looks good, though! :+1:

tests/test_flask_admin.py
pypingou: When we make a GET request against this endpoint, there should not be a redirection, so removing the ``follow_redirects`` ensures that we are not redirected (otherwise it means there is something wrong in the code elsewhere).

tests/test_flask.py
pypingou: I'll adjust thanks

oddshocks:
Good to merge from my point of view. :)

pypingou:
Thanks for the review @oddshocks

PR #164, opened 2015-05-28
Fix all backend tests.

ralphbean:
Now that we have vcrpy, it *should* be easy to keep the entire test suite in working order.  We let it fall into serious disrepair just because of the fluid nature of the upstreams we were trying to test against.  Now, with frozen copies of their responses we can actually make use of this test suite, instead of ignoring it.

There are two tests that still fail, and I'm very confused about them.  In the rubygems module, there is a test that says its for the "custom" backend (even though it is testing the rubygems backend).  And it asserts that a certain exception should be raised, but it doesn't explain why.  That exception is now *not* being raised (causing the test to fail).. but it's unclear to me what should be changed to fix that.

Maybe @pypingou can have a look there?


pypingou:
Ok so for the rubygems backend this is the problem:

a) the docstring is not up to date

b) in https://github.com/fedora-infra/anitya/commit/a099bb7a0dca1e01c68feaa3f94c1c78a87ef495 we changed from http://rubygems.org/api/v1/gems/biofoobar.json to http://rubygems.org/api/v1/versions/biofoobar/latest.json and the first one returns a 404 while the second returns a version ``unknown``. So we need to adjust the backend for this, on it :)

pypingou:
Well actually, both return a 404 but the second one returns a JSON blob while the first one does not.

ralphbean:
Cool!

ralphbean:
:+1: from me!  Look good to you, @pypingou?  Jenkins is happy and running the backend stuff:  http://jenkins.cloud.fedoraproject.org/job/anitya/192/console

pypingou:
Good for me as well, let's merge! :)

PR #167, opened 2015-06-01
add haskell Stackage backend

juhp:
This adds a new backend for Stackage (stable Haskell packages).

Stackage is a stabler subset of Hackage which we want to use for
Fedora Haskell package tracking when available.
(Not all Fedora Haskell packages are in Stackage though so we still need the Hackage backend too.)

This is derived from hackage.py.

I guess some test(s) needed too, if this looks ok?  Anything else?

### DIFF
--- a/anitya/lib/backends/stackage.py
+++ b/anitya/lib/backends/stackage.py
@@ -1,0 +1,58 @@
+# -*- coding: utf-8 -*-
+
+"""
+ (c) 2014-2015 - Copyright Red Hat Inc
+ Authors:
+   Pierre-Yves Chibon <pingou@pingoured.fr>
+   Jens Petersen <petersen@redhat.com>
+"""
+
+
+from anitya.lib.backends import BaseBackend, get_versions_by_regex, REGEX
+
+
+class StackageBackend(BaseBackend):
+    ''' The custom class for Haskell projects hosted on Stackage.org.
+    This backend allows to specify a version_url and a regex that will
+    be used to retrieve the version information.
+    '''
+
+    name = 'Stackage'
+    examples = [
+        'https://www.stackage.org/package/conduit',
+        'https://www.stackage.org/package/cabal-install',
+    ]
+
+    @classmethod
+    def get_version(cls, project):
+        ''' Method called to retrieve the latest version of the projects
+        provided, project that relies on the backend of this plugin.
+        :arg Project project: a :class:`model.Project` object whose backend
+            corresponds to the current plugin.
+        :return: the latest version found upstream
+        :return type: str
+        :raise AnityaPluginException: a
+            :class:`anitya.lib.exceptions.AnityaPluginException` exception
+            when the version cannot be retrieved correctly
+        '''
+        return cls.get_ordered_versions(project)[-1]
+
+    @classmethod
+    def get_versions(cls, project):
+        ''' Method called to retrieve all the versions (that can be found)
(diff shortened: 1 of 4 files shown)


anitya/lib/backends/stackage.py
```diff
@@ -0,0 +1,58 @@
+# -*- coding: utf-8 -*-
+
+"""
+ (c) 2014-2015 - Copyright Red Hat Inc
```
pypingou: I guess the 2014 can be dropped here :)

anitya/lib/backends/stackage.py
```diff
@@ -0,0 +1,58 @@
...
+        :return: a list of all the possible releases found
+        :return type: list
+        :raise AnityaPluginException: a
+            :class:`anitya.lib.exceptions.AnityaPluginException` exception
+            when the versions cannot be retrieved correctly
+        '''
+        url = 'https://www.stackage.org/package/%(name)s' % {
+            'name': project.name}
+
+        regex = '%(name)s <span class="latest-version">([\d.]*) ' % {'name': project.name}
```
pypingou: Should we add the ``</span>`` at the end of the regex?
(do it always have the trailing space before the ``</span>`` ?)

anitya/lib/backends/stackage.py
```diff
@@ -0,0 +1,58 @@
+# -*- coding: utf-8 -*-
+
+"""
+ (c) 2014-2015 - Copyright Red Hat Inc
+ Authors:
+   Pierre-Yves Chibon <pingou@pingoured.fr>
```
pypingou: I guess my name can be removed from here until I actually contribute to this file ;-)

pypingou:
Looks good!

Indeed some tests would be most welcome.

Also, I have been looking to see if there is some kind of API to retrieve the version info and I couldn't find any, maybe it's worth asking the dev if they would consider making one :)

anitya/lib/backends/stackage.py
juhp: Okay up to you - most of the file is just copied from hackage.py.

anitya/lib/backends/stackage.py
juhp: Yes, space seems there but sound reasonable to add </span>.

juhp:
Good point about API - let me ask: there might be some REST but I am not sure.

anitya/lib/backends/stackage.py
juhp: (see below)

juhp:
Okay I asked upstream and there doesn't seem to be an real API for this
(see https://groups.google.com/d/msg/stackage/9h_RA-jNz_g/_cm7109xXDsJ).
They publish yaml files on github instead listing all the current package versions (~1.7MB+ each).
I did ask lightly if they would consider adding an API to query package versions.

So looks like the current approach above is good enough for now anyway.

anitya/lib/backends/stackage.py
juhp: Added in 55a394b

juhp:
How about now? :)

But if you want I can still remove 2014 and your name... (though I intentionally included them:)

pypingou:
I don't mind having them here but it is your work and the file wasn't created in 2014 so that part should still be changed I think :)

juhp:
Okay removed then :)

pypingou:
Looks all good, thanks for this!

PR #171, opened 2015-06-04
Fix the unit-tests for the stackage backend

pypingou:


oddshocks:
:+1:

pypingou:
Apparently not yet, jenkins isn't happy :)

pypingou:
Ok, so if @juhp could have a look at it, there is something odd with jenkins: http://jenkins.cloud.fedoraproject.org/job/anitya/

juhp:
Hmm sorry about all that - I confess I didn't test...

http://jenkins.cloud.fedoraproject.org/job/anitya/lastUnsuccessfulBuild/consoleText

juhp:
How can I reproduce this locally on my machine?
I followed the instructions from README.rst but don't get the above (cpphs) error.

How about merging the fix_unit-tests branch anyway? - nothing to lose I think. :)

pypingou:
Ok so for a reason I do not understand nor could reproduce, jenkins didn't manage to call the stackage server.
Thanks to @ralphbean we're now using the VCR library that allows caching the requested page, so I've added these cache files and now jenkins is happy.

My only wonder is if the files `` tests/tests/request-data/__main__.*`` are needed :)

ralphbean:
> My only wonder is if the files tests/tests/request-data/__main__.* are needed :)

That may depend on how the tests were run?  I.e. if you run the file directly versus using ``nosetests``, I bet the name of the module is different (i.e., it gets ``__main__`` prefixed).

ralphbean:
Easy way to tell:  delete them and re-run the tests with ``runtests.sh``.  If they don't come back, then they're not needed.

pypingou:
So indeed it's not needed :), removed!

ralphbean:
:100:

PR #178, opened 2015-06-26
Make the anitya_cron multithreaded and drop support for the progress-bar

pypingou:
With this version, using 10 threads running the cron drops from 90minutes
to less than 10 minutes.

### DIFF
--- a/files/anitya_cron.py
+++ b/files/anitya_cron.py
@@ -1,19 +1,29 @@
 #!/usr/bin/env python
 #-*- coding: utf-8 -*-
 
+import sys
+import logging
+import itertools
+import multiprocessing
+
 import anitya
 import anitya.app
 import anitya.lib.exceptions
 import anitya.lib.model
 
-import logging
-import sys
 
-PBAR = True
-try:
-    from progressbar import Bar, ETA, Percentage, ProgressBar, RotatingMarker
-except ImportError:
-    PBAR = False
+def update_project(project):
+    """ Check for updates on the specified project. """
+    print(project.name)
+    session = anitya.lib.init(anitya.app.APP.config['DB_URL'])
+    project = anitya.lib.model.Project.by_id(session, project.id)
+    try:
+        anitya.check_release(project, session),
+    except anitya.lib.exceptions.AnityaException as err:
+        print(err)
+    finally:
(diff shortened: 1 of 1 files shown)


files/anitya_cron.py
```diff
@@ -1,19 +1,29 @@
...
-import sys
 
-PBAR = True
-try:
-    from progressbar import Bar, ETA, Percentage, ProgressBar, RotatingMarker
-except ImportError:
-    PBAR = False
+def update_project(project):
+    """ Check for updates on the specified project. """
+    print(project.name)
```
ralphbean: You always run as python3?  or is there a ``from __future__ import print_function`` here that I'm missing.

files/anitya_cron.py
```diff
@@ -41,28 +51,8 @@ def main(debug):
...
-        try:
-            anitya.check_release(project, session)
-        except anitya.lib.exceptions.AnityaException as err:
-            pass
-        cnt += 1
-        if PBAR:
-            pbar.update(cnt)
-    if PBAR:
-        pbar.finish()
+    p = multiprocessing.Pool(anitya.app.APP.config.get('CRON_POOL', 10))
```
ralphbean: Can you check to see if there is a performance difference between ``multiprocessing.Pool`` and ``multiprocessing.pool.ThreadPool``?

ralphbean:
:+1: as-is, with two non-blocking questions above.

files/anitya_cron.py
pypingou: it seems that ``print()`` works fine in python 2.7 without the import of ``__future__`` (tested on pkgdb02 with python 2.7.5).

That being said, it should be a call to the logger anyway :)

files/anitya_cron.py
pypingou: With ``multiprocessing.Poll``:
``
PYTHONPATH=. python files/anitya_cron.py  163.72s user 15.07s system 27% cpu 10:49.97 total 
``
With ``multiprocessing.pool.ThreadPool``:
``
PYTHONPATH=. python files/anitya_cron.py  163.99s user 22.22s system 22% cpu 14:05.40 total
``

PR #180, opened 2015-07-04
Handle flags not pointing to a project

pypingou:
When processing a flag we might end-up deleting the project it is
associated, in which case the template breaks if we do not check
if there is a project before displaying it


puiterwijk:
Looks good to me. :+1:

pypingou:
Thanks for the review

PR #189, opened 2015-08-07
Return the actually error message to the UI

pypingou:
This will help getting a better understand of what is going on and why
retrieving the release failed.


ralphbean:
:+1: yes :)

PR #193, opened 2015-09-01
Fix OpenID login for Fedora/Google/Yahoo

puiterwijk:
The OpenID provider may always decide to use a POST form
instead of a redirect with all the data.

Helps with fedora-infra/the-new-hotness#56

Signed-off-by: Patrick Uiterwijk <puiterwijk@redhat.com>

### DIFF
--- a/anitya/app.py
+++ b/anitya/app.py
@@ -192,8 +192,8 @@ def login():
         next=OID.get_next_url(), error=OID.fetch_error())
 
 
-@APP.route('/login/fedora/')
-@APP.route('/login/fedora')
+@APP.route('/login/fedora/', methods=('GET', 'POST'))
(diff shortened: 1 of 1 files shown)


pypingou:
Should we fix the other openid urls then?

puiterwijk:
In Anitya, or other things?
Because the /login route did have these already, and I think most of our codebases do.

puiterwijk:
https://github.com/fedora-infra/anitya/blob/fix-fedora-openid/anitya/app.py#L169

pypingou:
I was thinking more like fixing yahoo/google/... as well

puiterwijk:
Oh, right. Will also fix those

pypingou:
:+1:

Issue #198, opened 2015-09-22
Querying for projects with an empty homepage shouldn't return all projects


ralphbean:
As mentioned in fedora-infra/the-new-hotness#46, when you query the anitya API for the list of projects with ``homepage=''``, it currently returns all projects (I'm assuming the homepage filter is just ignored since ``bool(homepage) == False``).

Instead, it should return only the projects with *no* homepage **or** it should return some kind of 400 error, indicating a bad request.

pypingou:
Looking at the code, this makes me curious: https://github.com/fedora-infra/anitya/blob/master/anitya/api.py#L174

Akasurde:
We can close this issue as PR is merged.

pypingou:
Indeed :)

PR #201, opened 2015-09-25
Empty homepage returning projects with no homepage

vivekanand1101:
As suggested in issue #198 the check condition is changed.


ralphbean:
:+1: here!

PR #202, opened 2015-09-26
Added column for latest version in search page

### LINKED ISSUE #186: show version in search page
would be cool if project search added extra column to show latest version found:

https://release-monitoring.org/projects/search/?pattern=trac

vivekanand1101:
Tried to Fix #186 
At present, there is no column for latest version of projects in the search page.

### DIFF
--- a/anitya/templates/search.html
+++ b/anitya/templates/search.html
@@ -72,6 +72,11 @@
                     {{ project.homepage }}
                   </a>
                 </td>
+                <td>
+                  {% if project.latest_version %}
+                    {{ project.latest_version }}
+                  {% endif %}
+                </td>
               </tr>
             {% endfor %}
             </table>


ralphbean:
Looks good to me.  What happens if there is no ``latest_version`` for a particular project.  i.e., if it is ``None``?  Does it crash?

pypingou:
jinja won't say anything and display the page fine.

I just wanted to check it locally before merging it :)

vivekanand1101:
I don't have any data in my database so i couldn't verify this. My screen presently shows the default message when there are no projects being monitored. Also, i would like to know if you have some dummy datasets for different projects somewhere. It would be a great help.

vivekanand1101:
anitya_search_page
I used mysql for this.

vivekanand1101:
rsz_10b0f5044-65cd-11e5-83b2-57f767c4d70e
I used mysql for this.
Probably, we can give some default value to `version_url` [here](https://github.com/fedora-infra/anitya/blob/master/anitya/lib/model.py#L307) like 'not yet finalized' or something, just to avoid `None`.

pypingou:
Why do you speak about ``version_url``?

Otherwise, we can always embed the ``{{...}}`` in a ``{% if ... %}...{% endif %}``

vivekanand1101:
That was by mistake, i meant `latest_version`. The url was right though. And i agree on embedding part. It would be better than my suggestion.

vivekanand1101:
rsz_after_display_of_message
This is after i added the if..else part after [here](https://github.com/fedora-infra/anitya/blob/master/anitya/templates/search.html#L74). Following is the piece of code added:
                `<td>
                  {% if project.latest_version is not none %}
                    {{ project.latest_version }}
                  {% else %}
                    This is the message
                  {% endif %}
                </td>`

pypingou:
Looks proper :)

vivekanand1101:
So, should i commit the message part ? If yes, then what should be the message?

pypingou:
Should there be one?

vivekanand1101:
So, i should leave it blank?

pypingou:
What do you think? :)

vivekanand1101:
blank
Blank is just fine :)

vivekanand1101:
rsz_blank
Blank is just fine :)

anitya/templates/search.html
```diff
@@ -72,6 +72,13 @@
                     {{ project.homepage }}
                   </a>
                 </td>
+                <td>
+                  {% if project.latest_version is not none %}
+                    {{ project.latest_version }}
+                  {% else %}
```
pypingou: Why do you keep the ``{% else %}`` if it doesn't do anything? :)

pypingou:
Perfect for me, thanks!

PR #210, opened 2015-10-08
Added default regex in edit/add project page

vivekanand1101:
Besides the regex part, one more issue is resolved. Try selecting 'Source forge' and then select any other back end (edit/add product page). The instruction for source forge will still be visible.

### DIFF
--- a/anitya/lib/backends/custom.py
+++ b/anitya/lib/backends/custom.py
@@ -28,7 +28,7 @@ class CustomBackend(BaseBackend):
         'http://subsurface.hohndel.org/downloads/',
         'http://www.geany.org/Download/Releases',
     ]
-    test_regex_link = '<br /> More information in the '\
+    more_info = '<br /> More information in the '\
         '<a href=\'/about#test-your-regex\'>about#test-your-regex</a>'
     default_regex = REGEX % {'name': '<project name>'}
 
--- a/anitya/lib/backends/custom.py
+++ b/anitya/lib/backends/custom.py
@@ -28,7 +28,7 @@ class CustomBackend(BaseBackend):
         'http://subsurface.hohndel.org/downloads/',
         'http://www.geany.org/Download/Releases',
     ]
-    more_info = '<br /> More information in the '\
+    more_info = 'More information in the '\
         '<a href=\'/about#test-your-regex\'>about#test-your-regex</a>'
     default_regex = REGEX % {'name': '{project name}'}
 
--- a/anitya/lib/backends/custom.py
+++ b/anitya/lib/backends/custom.py
@@ -30,7 +30,7 @@ class CustomBackend(BaseBackend):
     ]
     more_info = '<br /> More information in the '\
         '<a href=\'/about#test-your-regex\'>about#test-your-regex</a>'
-    default_regex = REGEX % {'name': '<project name>'}
+    default_regex = REGEX % {'name': '{project name}'}
 
     @classmethod
     def get_version(cls, project):
--- a/anitya/lib/backends/custom.py
+++ b/anitya/lib/backends/custom.py
@@ -28,8 +28,9 @@ class CustomBackend(BaseBackend):
         'http://subsurface.hohndel.org/downloads/',
         'http://www.geany.org/Download/Releases',
     ]
-    more_info = '<br /> More information in the '\
+    test_regex_link = '<br /> More information in the '\
         '<a href=\'/about#test-your-regex\'>about#test-your-regex</a>'
+    default_regex = REGEX % {'name': '<project name>'}
 
     @classmethod
     def get_version(cls, project):
--- a/anitya/lib/backends/drupal6.py
+++ b/anitya/lib/backends/drupal6.py
@@ -27,7 +27,7 @@ class Drupal6Backend(BaseBackend):
         'https://www.drupal.org/project/pathauto',
         'https://www.drupal.org/project/wysiwyg',
     ]
-    more_info = '<br>If the project exists for Drupal 6 and 7 and is '\
+    more_info = '<br />If the project exists for Drupal 6 and 7 and is '\
         'monitored for both, you can prefix the name with `Drupal6:`, '\
         'for example: `Drupal6: cck`.'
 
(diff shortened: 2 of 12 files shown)


anitya/lib/backends/custom.py
```diff
@@ -28,8 +28,9 @@ class CustomBackend(BaseBackend):
         'http://subsurface.hohndel.org/downloads/',
         'http://www.geany.org/Download/Releases',
     ]
-    more_info = '<br /> More information in the '\
```
pypingou: :-1: for this change this is one of the attribute any backend can use to provide more information

anitya/templates/project_new.html
```diff
@@ -89,22 +93,33 @@
...
-    var more_info = {};
+    var test_regex_link = {};
     {% for plugin in plugins -%}
-    {%- autoescape false -%}
-    examples["{{ plugin.name }}"]="{{ plugin.examples | format_examples }}";
-    more_info["{{ plugin.name }}"]="{{ plugin.more_info }}";
-    {%- endautoescape -%}
+      {%- autoescape false -%}
+        examples["{{ plugin.name }}"]="{{ plugin.examples | format_examples }}";
+        console.log(examples);
```
pypingou: Some debugging left :)

anitya/templates/project_new.html
```diff
@@ -89,22 +93,33 @@
...
     $('#version_url_row').hide();
     $('#insecure_row').hide();
     $('#regex').attr('placeholder', '');
 
     $("label[for='version_url']").text('Version URL');
     var str = $( '#backend option:selected' ).text();
+
+    if (str != 'Sourceforge') {
+      $('#sourceforge_version_url_instr').hide();
+    }
```
pypingou: This should be un-necessary, the text after the version_url should disapear if the backend != SourceForge

anitya/templates/project_new.html
```diff
@@ -119,21 +134,31 @@
     if (str == 'Sourceforge'){
         $("label[for='version_url']").text('Sourceforge name');
         $('#version_url').after(
-        "<p>Use this field if the name used on the rss page of the project "
+        "<p id='sourceforge_version_url_instr'>Use this field if the name used on the rss page of the project "
```
pypingou: Same as above

anitya/templates/project_new.html
```diff
@@ -119,21 +134,31 @@
...
     if (str == 'custom'){
-      $('#regex_row').show();
       $('#version_url_row').show();
       $('#insecure_row').show();
       $('#regex').attr('placeholder', 'Leave empty for default regex');
+
+      if ($('#name').val()){
+        $('#regex_row').show();
+        $('#default_regex_row').show();
+      }
```
pypingou: Why not showing it all the time?

anitya/templates/project_new.html
vivekanand1101: screenshot from 2015-10-08 20-43-06

This is after selecting Source forge and then switching to custom.

pypingou: Hm indeed,

So what about adding at the top of ``show_hide()`` something like:

```javascript
if ($('#version_url').parent().find('p')){
    $('#version_url').parent().find('p').remove()
}
```

anitya/templates/project_new.html
vivekanand1101: It looked better to me if the regex row appears only when the default regex appears. For the appearance of default regex field, it was important that the project name is already entered (this is the method i have used).

anitya/templates/project_new.html
vivekanand1101: Ok, this one does the same thing but probably in a  better way. Will make change.

anitya/templates/project_new.html
vivekanand1101: :)

anitya/templates/project_new.html
pypingou: We could show the default regex all the time and adjust it upon change in the ``#name`` field :)

vivekanand1101: ok

anitya/lib/backends/drupal6.py
```diff
@@ -27,7 +27,7 @@ class Drupal6Backend(BaseBackend):
         'https://www.drupal.org/project/pathauto',
         'https://www.drupal.org/project/wysiwyg',
     ]
-    more_info = 'If the project exists for Drupal 6 and 7 and is '\
+    more_info = '<br>If the project exists for Drupal 6 and 7 and is '\
```
pypingou: This seems odd, why the break-row?

+ I believe it should be ``<br />``

ralphbean:
> This seems odd, why the break-row?

Grepping around in the anitya source:

```bash
anitya/lib/backends/custom.py:    more_info = '<br /> More information in the '\
anitya/lib/backends/drupal6.py:    more_info = 'If the project exists for Drupal 6 and 7 and is '\
anitya/lib/backends/drupal7.py:    more_info = 'If the project exists for Drupal 6 and 7 and is '\
anitya/templates/project_new.html:    var more_info = {};
anitya/templates/project_new.html:    more_info["{{ plugin.name }}"]="{{ plugin.more_info }}";
anitya/templates/project_new.html:    if (more_info[str]){
anitya/templates/project_new.html:      $('#example_txt').append('<br />' + more_info[str]);
```

Note that the custom backend already has a ``<br/>`` there.  Maybe that is where @vivekanand1101 got the idea?

It seems odd to me that the plugins themselves would get to declare some markup like that.  The ``<br/>`` tags should probably be removed and moved to ``anitya/templates/project_new.html`` (and there is already one there!)

ralphbean:
> It seems odd to me that the plugins themselves would get to declare some markup like that.

(My basis for this is [separation of presentation and content](https://en.wikipedia.org/wiki/Separation_of_presentation_and_content).  The content (``more_info``) shouldn't be mixed up with presentation (``"<br/>"``), when possible/reasonable.

vivekanand1101:
@ralphbean  done.

ralphbean:
Looks good to me.  @pypingou?

pypingou:
Will test it, but it looks good :)

pypingou:
Seems to work great, thanks!

PR #213, opened 2015-10-14
Added bit bucket backend

vivekanand1101:
Presently, i have removed 'tip' from the release list as it doesn't appear to be one of the releases.

Fixes https://github.com/fedora-infra/anitya/issues/141

### DIFF
--- a/anitya/lib/backends/bitbucket.py
+++ b/anitya/lib/backends/bitbucket.py
@@ -1,0 +1,75 @@
+# -*- coding: utf-8 -*-
+
+"""
+ (c) 2014 - Copyright Red Hat Inc
+
+ Authors:
+   Pierre-Yves Chibon <pingou@pingoured.fr>
+
+"""
+
+from anitya.lib.backends import BaseBackend, get_versions_by_regex
+from anitya.lib.exceptions import AnityaPluginException
+
+
+REGEX = b'class="name">([^<]*)</td'
+
+
+class BitBucket(BaseBackend):
+    ''' The custom class for projects hosted on bitbucket.org
+
+    This backend allows to specify a version_url and a regex that will
+    be used to retrieve the version information.
+    '''
+
+    name = 'BitBucket'
+    examples = [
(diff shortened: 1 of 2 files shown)


anitya/lib/backends/bitbucket.py
```diff
@@ -0,0 +1,75 @@
+# -*- coding: utf-8 -*-
+
+"""
+ (c) 2014 - Copyright Red Hat Inc
```
pypingou: That's incorrect :)

anitya/lib/backends/bitbucket.py
```diff
@@ -0,0 +1,75 @@
...
+
+ Authors:
+   Pierre-Yves Chibon <pingou@pingoured.fr>
```
pypingou: This as well :)

anitya/lib/backends/bitbucket.py
```diff
@@ -0,0 +1,75 @@
...
+            corresponds to the current plugin.
+        :return: a list of all the possible releases found
+        :return type: list
+        :raise AnityaPluginException: a
+            :class:`anitya.lib.exceptions.AnityaPluginException` exception
+            when the versions cannot be retrieved correctly
+
+        '''
+        if project.version_url:
+            url_template = 'https://bitbucket.org/%(version_url)s/downloads'
```
pypingou: Should this be the version_url or the project name?

pypingou:
So couple of question but the big missing point is the unit-tests :)

ralphbean:
> the big missing point is the unit-tests

Agreed.  Anitya has a good test suite already, it would be good to keep parity between new modules and new tests.

@vivekanand1101, do you know where the tests are and how to run them?

pypingou:
@vivekanand1101 said on IRC that he would not have time to fix this PR before a while, so I went ahead and finished it for him, the review is available at: https://github.com/fedora-infra/anitya/pull/218

Issue #212, opened 2015-10-14
Order backends


pypingou:
It seems the backends aren't ordered alphabetically, we should change this :)

farhaanbukhsh:
I would like to work on this issue :)
@pypingou

farhaanbukhsh:
Hints required :)

pypingou:
@farhaanbukhsh I've seen you work so I'd like to ask you to take a look at the sources and try to find the solution by yourself :)

That being said, if you are not sure, feel free to ping me, here on or IRC and we can do it together, but I'd like it if you could try yourself first :)

farhaanbukhsh:
@pypingou okay i will try it although I don't know what are backends but let me look into the code base :). I will ping you

PR #214, opened 2015-10-15
Fixing #212 ordering backends

farhaanbukhsh:


farhaanbukhsh:
I got to this conclusion  in `ui.py` plugins is getting populated from load_plugin() in `plugins.py` which gets all the plugins form  get_plugins() which in turn calls a load function. I tried to sort it while entering into database but that is already sorted. So either it  has to be sorted in `ui.py` or `form.py`
Your thoughts @pypingou ?

pypingou:
Perfect, thanks!

farhaanbukhsh:
thanks to you :)

PR #217, opened 2015-10-20
Use bootstrap classes everywhere to make the search forms look nicer.

ralphbean:
Use bootstrap classes to align the search button and input.


pypingou:
Looks way nicer, :+1:

PR #219, opened 2015-10-23
Add the possibility to separate odd updates vs failed updates

pypingou:
Odd updates are projects for which we find a new version that we find to
be inferior to the latest update we had.
Failed updates are plain failure to retrieve the newest update.

### DIFF
--- a/anitya/lib/model.py
+++ b/anitya/lib/model.py
@@ -451,7 +451,15 @@ class Project(BASE):
             query = query.filter(
                 Project.logs != None,
                 Project.logs != 'Version retrieved correctly',
+                ~Project.logs.ilike('Something strange occured%'),
             )
+        elif status == 'odd':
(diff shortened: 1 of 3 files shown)


anitya/lib/model.py
```diff
@@ -451,7 +451,15 @@ def updated(
             query = query.filter(
                 Project.logs != None,
                 Project.logs != 'Version retrieved correctly',
+                ~Project.logs.ilike('Something strange occured%'),
```
ralphbean: Is this ``~`` intentional?  What does it do?

pypingou: It's the negation, so in this case I do not want the project whose log start with 'Something strange occured'

ralphbean: Cool!

ralphbean:
:+1:

PR #221, opened 2015-10-27
Record and show the status of the cron job checking new releases

pypingou:
fixes https://github.com/fedora-infra/anitya/issues/190


ralphbean:
No alembic upgrade script to add the runs table?

pypingou:
For new tables I just invoke the createdb script normally

ralphbean:
:+1: then.

PR #254, opened 2016-01-30
Fix bug causing incorrect homepage parsing

cydrobolt:


pypingou:
Thank you very much :)

Issue #252, opened 2016-01-20
ui improvement for not found packages


glensc:
so you search package that does not exist.
would be nice if besides saying "Oups this is embarrassing, it seems that no projects are being monitored currently." it could link directly to adding package page having package name pre-filled.
also the text is kind of misleading like there are NO projects at all monitored, but it intended to say probably NO projects BY THAT NAME.

so,

https://release-monitoring.org/projects/search/?pattern=example-project-not-found

Oups this is embarrassing, it seems that no projects by that name are being monitored currently.
[Try adding it](https://release-monitoring.org/project/new?name=example-project-not-found)

ps: as you can see, the `name` parameter support is already there, should be simple as just changing html

pypingou:
Fixed in #256

PR #256, opened 2016-02-01
Improved UI for not found packages

Prashant-Surya:
Fix for issue #252. A link is added in the not found message, so that a new prjoect can be directly added with the name field filled in the new_project form.


pypingou:
Just perfect :)

PR #261, opened 2016-02-15
Feature/maven central update

msimacek:
Enables using version_url field for inputting artifact coordinates instead of
abusing project name (which will still be possible for backwards compatibility).

### DIFF
--- a/anitya/lib/backends/maven.py
+++ b/anitya/lib/backends/maven.py
@@ -8,8 +8,13 @@
 
 """
 
+import re
 
 from anitya.lib.backends import BaseBackend, get_versions_by_regex
+from anitya.lib.exceptions import AnityaPluginException
+
+
+REGEX = r'\<a[^>]+\>(\d[^</]*)'
 
 
 class MavenBackend(BaseBackend):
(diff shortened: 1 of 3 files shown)


anitya/lib/backends/maven.py
```diff
@@ -53,9 +58,18 @@ def get_versions(cls, project):
...
-        return get_versions_by_regex(url, regex, project)
+        if re.match(r'https?://repo\d+.maven.org/', project.homepage):
+            url = project.homepage
+            coordinates = None
+        else:
+            coordinates = project.version_url or project.name
+        if coordinates:
+            if ':' not in coordinates:
+                raise AnityaPluginException(
+                    "Aritfact coordinates need to be in format "
```
pypingou: The ``coordinates`` here is the first time we use this term publicly. Should we use another one?

anitya/lib/backends/maven.py
```diff
@@ -53,9 +58,18 @@ def get_versions(cls, project):
...
-
-        regex = r'\<a[^>]+\>(\d[^</]*)'
-
-        return get_versions_by_regex(url, regex, project)
+        if re.match(r'https?://repo\d+.maven.org/', project.homepage):
+            url = project.homepage
+            coordinates = None
+        else:
+            coordinates = project.version_url or project.name
+        if coordinates:
```
pypingou: We could drop this ``if`` here and put the entire code block in the ``else`` above, no?

msimacek: Definitely. I had more branches there, but then removed them and forgot to simplify it

anitya/lib/backends/maven.py
msimacek: It's official term: https://maven.apache.org/pom.html#Maven_Coordinates
Do you have an idea for better wording?

pypingou: Not really, but then should we mention it in the UI as well?

msimacek: What about just dropping the word "coordinates"? "Artifact needs to be in format groupId:artifactId"

pypingou: Sounds good to me

pypingou:
Thanks :)

PR #266, opened 2016-02-17
Making homepage field url type to accept only url strictly

sudheesh001:
Fixes https://github.com/fedora-infra/anitya/issues/265

### DIFF
--- a/tests/test_flask.py
+++ b/tests/test_flask.py
@@ -269,6 +269,16 @@ a backend for the project hosting. More information below.</p>"""
             self.assertTrue(
                 '<td><label for="regex">Regex</label></td>' in output.data)
 
+            false_data = {
+                'name': 'repo_manager',
+                'homepage': 'testurlrepo_manager',
+                'backend': 'PyPI',
+            }
+
+            output = c.post(
+                '/project/new', data=false_data, follow_redirects=True)
+            self.assertFalse('<h1>Project: repo_manager</h1>' in output.data)
+
             data = {
                 'name': 'repo_manager',
                 'homepage': 'https://pypi.python.org/pypi/repo_manager',
--- a/anitya/templates/project_new.html
+++ b/anitya/templates/project_new.html
@@ -21,7 +21,7 @@
               {{ render_field_in_row(form.name) }}
             </tr>
             <tr>
-              {{ render_field_in_row(form.homepage) }}
+              {{ render_field_in_row(form.homepage, type="url") }}
             </tr>
             <tr>
               {{ render_field_in_row(form.backend) }}


pypingou:
Should we add some tests for this?

sudheesh001:
@pypingou Could you guide me around that, I've seen the `test_project.py` file and most of it seems like the test for the data which is selected than the information which is to be inserted via the form. Since this is a front end change, does it really need the test ?

pypingou:
You probably want to look more at: https://github.com/fedora-infra/anitya/blob/master/tests/test_flask.py#L249-L315

sudheesh001:
@pypingou Is this what is required ?
```py
            falseData = {
                'name': 'repo_manager',
                'homepage': 'testurlrepo_manager',
                'backend': 'PyPI',
            }

            output = c.post(
                '/project/new', data=falseData, follow_redirects=True)
            self.assertTrue('Please enter a URL.' in output.data)
```
I am unable to run the tests because of the following error it throws up
```
>>> from tests import Modeltests, create_distro, create_project
  File "<stdin>", line 1, in <module>
  File "tests/__init__.py", line 33, in <module>
    import vcr
ImportError: No module named vcr
```

pypingou:
> ImportError: No module named vcr

Is listed in: https://github.com/fedora-infra/anitya/blob/master/test_requirements.txt

As for the tests, it looks good I think yes.

sudheesh001:
@pypingou Thanks, i've made the changes and updated the PR after rebasing

tests/test_flask.py
```diff
@@ -30,6 +30,8 @@
 import unittest
 import sys
 import os
+import string
+import urlparse
```
pypingou: Why these two imports?

tests/test_flask.py
```diff
@@ -275,6 +277,16 @@ def test_new_project(self):
...
 
+            falseData = {
+                'name': 'repo_manager',
+                'homepage': 'testurlrepo_manager',
+                'backend': 'PyPI',
+            }
+
+            output = c.post(
+                '/project/new', data=falseData, follow_redirects=True)
+            self.assertTrue('Please enter a URL.' in output.data)
```
pypingou: Maybe this should be moved up, we define data above ``falseData`` (which in python should be ``false_data``) but only use ``data`` below.

pypingou:
Could you run the test suite?

sudheesh001:
Yup i just ran the test suite after making the updates in the test and it throws the error as follows
```
======================================================================
FAIL: test_new_project (__main__.FlaskTest)
Test the new_project function.
----------------------------------------------------------------------
  File "tests/test_flask.py", line 286, in test_new_project
    self.assertTrue('Please enter a URL.' in output.data)
AssertionError: False is not true

----------------------------------------------------------------------
```

tests/test_flask.py
```diff
@@ -276,6 +282,10 @@ def test_new_project(self):
             }
 
             output = c.post(
+                '/project/new', data=false_data, follow_redirects=True)
+            self.assertTrue('Please enter a URL.' in output.data)
+
```
pypingou: When speaking about moving up, I was thinking about the whole block not just ``false_data``.

pypingou:
So it looks like your test has failed :)

sudheesh001:
@pypingou I've made the changes, on second thought, It was easier to check for the strings on the success page and assertFalse() for this data as it shouldn't move to a success page because of the bad homepage URL. It passes the tests now.

pypingou:
Anything I can help you with? :)

ralphbean:
ping @sudheesh001 - can we help with resolving this?

pypingou:
I opened https://github.com/fedora-infra/anitya/pull/307 based on the work from @sudheesh001 and w/ a little additional tweak :)

PR #270, opened 2016-03-04
Appended UTC as timezone

### LINKED ISSUE #268: timezoneless dates
`Last check ended at 2016-02-22 12:26:13`

it's not clear what the time is if you do not include timezone in it. the specific example was taken from page footer, but i guess there are more places where datetime is presented (logs handler for example)

jinankjain:
Fixes #268


pypingou:
Cool, thanks!

Issue #283, opened 2016-04-15
stackage backend broken


pypingou:
@juhp it seems that the stackage backend is broken: https://release-monitoring.org/projects/updates/failed?name=&log=stackage want to look at it?

pypingou:
@juhp I think I got it :) (cf the PR above, #284)

juhp:
Thank you for noticing and fixing it!

Sorry I suspect it was probably me who broke it unknowingly on the Stackage side.
Do you want me to test it?

pypingou:
@juhp if you could test the PR that would be awesome, but at least the unit-tests seem happy :)

juhp:
I tested it locally and it seems to work fine: I am not sure that multiple regex's are really needed though. Maybe enough just to update the regex to the new html?

pypingou:
I was wondering if all the pages changed or only some, since I didn't know I made it this way

juhp:
Okay sure thanks - you're very quick. :)

They all changed: see fpco/stackage-server@07e6121.

pypingou:
this got fixed :)

pypingou:
hm, there is something odd, the stackage page is being inconsistent, one time the version number for LTS is linked, one time it isn't.

It's as if there are 2 servers running 2 different versions of the application

pypingou:
Reported upstream at https://github.com/fpco/stackage-server/issues/182

juhp:
It is possible that one of the server is out of sync yes - let me ask upstream about that...

We could also track Stackage via the Hackage pages though it would be less flexible since it only shows the latest LTS version but maybe it is good enough: the Stackage page still seems to be seeing some changes regularly... sorry for the trouble.

pypingou:
It got apparently fixed upstream :)

PR #285, opened 2016-04-19
update Stackage version html regexp (#283)

juhp:
I think this should be sufficient to cover the html changes on stackage.org.


pypingou:
Isn't this a duplicate of #284 ?

juhp:
Perhaps I am doing this wrong - I thought simpler to start with a new PR: should I PR to the fix_stackage branch instead?

juhp:
Ah, I need to update the tests too I guess. :-(

pypingou:
Well, it's more than if #284 works, then let's go ahead with it, otherwise indeed we can merge your PR :)

juhp:
Okay it is up to you I think: like I say there is only one global html format on stackage.org, so no need really to support the old html.

pypingou:
Ok, then let's go with your PR indeed, mind adjusting the unit-tests then?

Thanks :)

juhp:
I think they should pass now.

pypingou:
Thanks :)

Issue #286, opened 2016-04-19
Github backend doesn't find plexus-archiver release


mizdebsk:
A couple of latest plexus-archiver releases can't be found by Github backend.

https://github.com/codehaus-plexus/plexus-archiver/tags
https://www.release-monitoring.org/project/3655

mizdebsk:
And that's what I'm getting when trying to force version update:
```
    $ curl -d id=3655 https://www.release-monitoring.org/api/version/get
    {
      "backend": "GitHub",
      "created_on": 1412175072.0,
      "homepage": "https://github.com/codehaus-plexus/plexus-archiver",
      "id": 3655,
      "name": "plexus-archiver",
      "packages": [
        {
          "distro": "Fedora",
          "package_name": "plexus-archiver"
        }
      ],
      "regex": null,
      "updated_on": 1461071695.0,
      "version": "plexus-archiver-3.0",
      "version_url": "https://github.com/codehaus-plexus/plexus-archiver",
      "versions": [
        "plexus-archiver-3.0.2",
        "plexus-archiver-3.0",
        "plexus-archiver-2.9.1",
        "plexus-archiver-2.9",
        "plexus-archiver-2.8.4",
        "plexus-archiver-2.8.3",
        "plexus-archiver-2.8.2",
        "plexus-archiver-2.8.1",
        "plexus-archiver-2.8",
        "plexus-archiver-2.10"
      ]
    }
```

PR #287, opened 2016-04-19
Add test case for #286

mizdebsk:
Reproducer and regression test for issue #286


pypingou:
And the test passes?

mizdebsk:
Fails (as expected) due to issue it is reproducing (#286)

PR #289, opened 2016-04-29
Prevent blacklisted users from logging in

pypingou:
Anitya is made for more than just Fedora, if people keep trashing its
content despite having been told not to and having been explained that
pkgdb2 is the place where notifications need to be turn off, I see
black-listing them as the only solution to keep anitya clean.


anitya/app.py
```diff
@@ -112,13 +112,17 @@ def after_openid_login(resp):
     login was successful on the OpenID server.
     '''
     default = flask.url_for('index')
+    blacklist = APP.config['BLACKLISTED_USERS']
     if resp.identity_url:
+        next_url = flask.request.args.get('next', default)
         openid_url = resp.identity_url
+        if openid_url in blacklist or resp.email in blacklist:
```
puiterwijk: I would suggest to at least add a flask.flash to tell them what went on, or they'll come asking for support.

anitya/app.py
```diff
@@ -112,13 +112,17 @@ def after_openid_login(resp):
     login was successful on the OpenID server.
     '''
     default = flask.url_for('index')
+    blacklist = APP.config['BLACKLISTED_USERS']
```
puiterwijk: First up, you probably want to use APP.config.get('BLACKLISTED_USERS', []), to make sure that it is backwards compatible.

puiterwijk: Also, in the case of Fedora, you will most likely want to maintain this blacklist in the private repo.
Otherwise we are ending up doing public naming & shaming, which is going to cause a lot of hostility.
Also, will this really stop people? Since it's quite easy to create a new OpenID identity, which will once more allow them to mess it all up.

anitya/app.py
pypingou: I was wondering about this and didn't find a good message. Would you have an idea?

anitya/app.py
pypingou: > APP.config.get('BLACKLISTED_USERS', []), to make sure that it is backwards compatible.

No need since it's defined in the default_config file

> maintain this blacklist in the private repo.

That's a good idea

> Also, will this really stop people?

It is easy to use another openid identity but since the case we've had so far is not really a real spammer, I figure the work needed to go around this blacklist is greater than the annoyance (at least I hope so).

anitya/app.py
puiterwijk: "We are very sorry, but due to repeated vandalism, your account has been blocked from logging in to this service."?

puiterwijk:
+1, albeit reluctantly since I'd hoped we didn't need something like this..

pypingou:
I hope we will not use it, but I realize an open service might require it and I'd rather have it present in the code instead of missing them in time of needs.

pypingou:
Thanks for the review :)

PR #291, opened 2016-04-29
Port from using user's email to use user's email address

pypingou:
Fixes https://github.com/fedora-infra/anitya/issues/205

(We will need a change to fedmsg_meta to convert back from email Fedora users using the Fedora openid identifiers)


lmacken:
LGTM :+1:

PR #295, opened 2016-05-08
Fix #276: Specified (UTC) timezone in the footer.

### LINKED ISSUE #276: Specify the (UTC) timezone in the footer
In the footer of anitya we inform the user about the time when the last run of the cron job checking for new versions ran. This date/time is provided in UTC but doesn't explicitly say so.

We should fix this.

Reported originally by @glensc in #268

souravbadami:
Appended (UTC) timezone in the footer.


pypingou:
Cool, thanks!

PR #296, opened 2016-05-09
Fix #293: Displays backend in search results.

### LINKED ISSUE #293: Display backend in search results
When searching for projects, it would be nice to directly see which backend they're using. The use case is that lots of projects were initially created with custom backend and I'd like to see this to be able to switch them to better backend. Possibly even display more columns, there's plenty of space in that view.

souravbadami:
Added backend column in search results. #293

### DIFF
--- a/anitya/templates/search.html
+++ b/anitya/templates/search.html
@@ -73,6 +73,9 @@
                   </a>
                 </td>
                 <td>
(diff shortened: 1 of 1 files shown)


pypingou:
Simple enough, thanks :)

PR #298, opened 2016-05-11
Fix timezone-sensitive tests.

ralphbean:
These fail for me during certain times of the day when the date in my locale
and the date in UTC do not match.  This switches the tests to be consistently
in UTC, removing that sensitivity.


pypingou:
Good idea, :+1:

PR #282, opened 2016-04-12
Change over to updates.drupal.org backend for dupal versions

daveol:
This is a pull request to fix #281

PR #300, opened 2016-05-11
Add a new check-feed mode for the backend.

ralphbean:
This is a rewrite of #239.

- It adds a new check_feed method to all of the backends that I could figure out.
- It adds tests for those.
- It adds a new --check-feed option to the cronjob, which uses those methods instead of the regular approach.

Without this PR, we currently run the cronjob twice a day (once every 12 hours)
which means that some new upstream releases we don't find out about for a "very
long time" (~12 hours).  We do that because our cronjob checks for a new
upstream release for *every* project we know about.  That takes a very long
time to run.

This new mode for the cronjob checks only for projects that are listed in the
RSS feeds and APIs of public indexes like pypi, rubygems, and cpan, etc..
It runs very quickly in comparison.  I bet we could run it as a cronjob once
every 5 minutes without much worry (we should wrap it in a lock script, though,
just to make sure they don't pile up).  I'd like to get as close to realtime as
possible.

In addition to scanning for new upstream releases, if the cronjob encounters a
*new* package in the RSS feed or API, then it also adds that to anitya's
database of projects (so that will grow over time, automatically now).

### DIFF
--- a/anitya/lib/backends/hackage.py
+++ b/anitya/lib/backends/hackage.py
@@ -13,7 +13,7 @@ from anitya.lib.backends import BaseBackend, get_versions_by_regex, REGEX
 
 
 class HackageBackend(BaseBackend):
-    ''' The custom class for projects hosted on Google code.
+    ''' The custom class for projects hosted on hackage.
 
     This backend allows to specify a version_url and a regex that will
     be used to retrieve the version information.
@@ -62,3 +62,12 @@ class HackageBackend(BaseBackend):
         regex = REGEX % {'name': project.name}
 
         return get_versions_by_regex(url, regex, project)
+
+    @classmethod
+    def check_feed(cls):
+        ## See http://hackage.haskell.org/api#recentPackages
+        ## It should be possible to query this, but I can't figure out how to
+        ## get it to give me non-html.
+        # url = 'http://hackage.haskell.org/packages/recent/revisions'
+        raise NotImplementedError()
+
--- a/anitya/lib/backends/npmjs.py
+++ b/anitya/lib/backends/npmjs.py
@@ -1,13 +1,13 @@
 # -*- coding: utf-8 -*-
 
 """
- (c) 2014 - Copyright Red Hat Inc
+ (c) 2014-2016 - Copyright Red Hat Inc
 
  Authors:
    Pierre-Yves Chibon <pingou@pingoured.fr>
+   Ralph Bean <rbean@redhat.com>
 
 """
-
 
 from anitya.lib.backends import BaseBackend
 from anitya.lib.exceptions import AnityaPluginException
@@ -92,3 +92,28 @@ class NpmjsBackend(BaseBackend):
             raise AnityaPluginException('No versions found at %s' % url)
 
         return data['versions'].keys()
+
+    @classmethod
+    def check_feed(cls):
+        ''' Return a generator over the latest 40 uploads to npmjs.org
+
+        by querying an weird JSON endpoint.
+        '''
+
+        url = 'https://registry.npmjs.org/-/all/static/today.json'
+
+        try:
+            response = cls.call_url(url)
+        except Exception:  # pragma: no cover
+            raise AnityaPluginException('Could not contact %s' % url)
+
+        try:
+            data = response.json()
+        except Exception:  # pragma: no cover
+            raise AnityaPluginException('No JSON returned by %s' % url)
+
+        for item in data[:40]:
+            for version in item.get('versions', []):
+                name = item['name']
+                homepage = 'http://npmjs.org/package/%s' % name
+                yield name, homepage, cls.name, version
--- a/anitya/lib/backends/pypi.py
+++ b/anitya/lib/backends/pypi.py
@@ -1,13 +1,15 @@
 # -*- coding: utf-8 -*-
 
 """
- (c) 2014 - Copyright Red Hat Inc
+ (c) 2014-2016 - Copyright Red Hat Inc
 
  Authors:
    Pierre-Yves Chibon <pingou@pingoured.fr>
+   Ralph Bean <rbean@redhat.com>
 
 """
 
+import xml2dict
 
 from anitya.lib.backends import BaseBackend
 from anitya.lib.exceptions import AnityaPluginException
@@ -76,3 +78,30 @@ class PypiBackend(BaseBackend):
             raise AnityaPluginException('No JSON returned by %s' % url)
 
         return data['releases'].keys()
+
+    @classmethod
+    def check_feed(cls):
+        ''' Return a generator over the latest 40 uploads to PyPI
+
+        by querying an RSS feed.
+        '''
+
+        url = 'https://pypi.python.org/pypi?%3Aaction=rss'
+
+        try:
+            response = cls.call_url(url)
+        except Exception:  # pragma: no cover
+            raise AnityaPluginException('Could not contact %s' % url)
+
+        try:
+            parser = xml2dict.XML2Dict()
+            data = parser.fromstring(response.text)
+        except Exception:  # pragma: no cover
+            raise AnityaPluginException('No XML returned by %s' % url)
+
+        items = data['rss']['channel']['item']
+        for entry in items:
+            title = entry['title']['value']
+            name, version = title.rsplit(None, 1)
+            homepage = 'https://pypi.python.org/pypi/%s' % name
+            yield name, homepage, cls.name, version
--- a/anitya/lib/backends/rubygems.py
+++ b/anitya/lib/backends/rubygems.py
@@ -1,10 +1,11 @@
 # -*- coding: utf-8 -*-
 
 """
- (c) 2014 - Copyright Red Hat Inc
+ (c) 2014-2016 - Copyright Red Hat Inc
 
  Authors:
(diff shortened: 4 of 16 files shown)


anitya/lib/backends/npmjs.py
```diff
@@ -92,3 +92,28 @@ def get_versions(cls, project):
...
+            response = cls.call_url(url)
+        except Exception:  # pragma: no cover
+            raise AnityaPluginException('Could not contact %s' % url)
+
+        try:
+            data = response.json()
+        except Exception:  # pragma: no cover
+            raise AnityaPluginException('No JSON returned by %s' % url)
+
+        for item in data[:40]:
```
ncoghlan: Does the limit here risk missing updates if not polled frequently enough or if a large batch of updates land at the same time?

ralphbean: Yes.  In fact, all of these backends share that same risk.

The saving grace is that, in production we should schedule this ``--check-feed`` mode to run every 5 minutes but we should *continue* to schedule the full scan mode to run every 12 hours, which can catch anything that falls through the cracks (as well as also catch updates from backends which do not support this ``check_feed`` stuff).

files/anitya_cron.py
```diff
@@ -13,11 +12,40 @@
...
+
+def projects_by_feed(session):
+    """ Return the list of projects out of sync, found by feed listings.
+
+    If a new entry is noticed and we don't have a project for it, add it.
+    """
+    for name, homepage, backend, version in indexed_listings():
+        project = anitya.lib.model.Project.get_or_create(
+            session, name, homepage, backend)
+        if project.latest_version == version:
```
ncoghlan: This may check may be false for an already recorded release to a project with multiple stable branches (e.g. a Django 1.8.x security update that's already been detected may be reported again later, since Django 1.9.x is the latest version in absolute terms)

ralphbean: Hm.  If that's the case, then the subsequent code will schedule an unnecessary but harmless check of the project.

ncoghlan:
This looks like a very nice improvement to me :+1:

anitya/lib/backends/hackage.py
```diff
@@ -62,3 +62,12 @@ def get_versions(cls, project):
         regex = REGEX % {'name': project.name}
 
         return get_versions_by_regex(url, regex, project)
+
+    @classmethod
+    def check_feed(cls):
+        ## See http://hackage.haskell.org/api#recentPackages
+        ## It should be possible to query this, but I can't figure out how to
+        ## get it to give me non-html.
+        # url = 'http://hackage.haskell.org/packages/recent/revisions'
```
pypingou: maybe @juhp would know (but that's something to look at later)

anitya/lib/backends/pypi.py
```diff
@@ -76,3 +78,30 @@ def get_versions(cls, project):
...
+
+        url = 'https://pypi.python.org/pypi?%3Aaction=rss'
+
+        try:
+            response = cls.call_url(url)
+        except Exception:  # pragma: no cover
+            raise AnityaPluginException('Could not contact %s' % url)
+
+        try:
+            parser = xml2dict.XML2Dict()
```
pypingou: I quite like this approach (rss to dict) and I wonder if it wouldn't make sense to port the SourceForge backend to use it

pypingou:
:+1: for me as well., nice change! :)

anitya/lib/backends/rubygems.py
```diff
@@ -74,3 +74,27 @@ def get_versions(cls, project):
                 'Project or version unknown at %s' % url)
 
         return [data['version']]
+
+    @classmethod
+    def check_feed(cls):
+        ''' Return a generator over the latest 50 uploads to rubygems.org
+
```
voxik: Will this be checked often enough to be sure that we don't miss anything? Should I open upstream RFE asking to implement some dynamic filter, e.g. "give all updates since gem foo v1.2.3 was released" or alternatively "give me all project since some timestamp"?

pypingou: @ralphbean started to answer it at: https://github.com/fedora-infra/anitya/pull/300#discussion_r62793372

But RFE upstream would be cool nonetheless as it would reduce the amount of data we ask them.

anitya/lib/backends/npmjs.py
voxik: "full scan"? There is nothing like full scan as far as I understand. The "full scan" looks just for known packages, i.e. if you miss newly introduced package, the record for that package is never created in Anitya and hence it won't be discovered until its new release is caught by this code again.

pypingou: full scan is what we are currently doing

anitya/lib/backends/rubygems.py
voxik: Opened RFE: https://github.com/rubygems/rubygems.org/issues/1266

anitya/lib/backends/npmjs.py
voxik: Right, but then you miss some newly introduced packages ...

anitya/lib/backends/rubygems.py
pypingou: Thanks! :)

anitya/lib/backends/npmjs.py
pypingou: I'm not sure I fully follow you, how would this PR change anything in that regards compared to the current approach?

anitya/lib/backends/rubygems.py
ralphbean: The *best* case we could get would be to have all of the upstream forges enabled with *webhooks*, so they can ping us every time someone uploads a new release of anything.

I looked at all of them last night, and I think rubygems.org is the only one [that can do this](http://guides.rubygems.org/rubygems-org-api/#webhook-methods).  If more of them supported it, it would be worth adding support to anitya to do it, I think.  Then we can have real-real-time.

anitya/lib/backends/npmjs.py
ralphbean: > the record for that package is never created in Anitya and hence it won't be discovered until its new release is caught by this code again.

This is correct.

- @pypingou, our current "full scan" mode scans for all of the projects that anitya already knows about, and it does so very well.
- This new mode not only updates projects that anitya already know about, but it also creates entries for *new* projects that it discovers.

If we worry only about the first part here, then it is ok to miss an announcement due to the aliasing effect at play here, since our full scan will pick up the release that we missed.  However, like @voxik says, we run the risk of failing to discover new projects if they get pushed off the list between our 5-minute scan windows.

pypingou: Ok, I understand the issue, that could be a problem if people assume anitya knows every projects

anitya/lib/backends/rubygems.py
pypingou: Maybe something to ping @dstuff about for next-gen pypi :)

dstufft: See pypa/warehouse#360

ralphbean:
Thanks all.  Merging this.

anitya/lib/backends/hackage.py
lubomir: Just adding `.rss` seems to do it: http://hackage.haskell.org/packages/recent.rss

anitya/lib/backends/pypi.py
daveol: Maybe the drupal ones could be ported too, since they also return xml (since #282 )

anitya/lib/backends/hackage.py
ralphbean: Doh!  I tried passing all kinds of headers... but not that.  Thanks @lubomir!  Filed #301.

anitya/lib/backends/rubygems.py
voxik: @ralphbean yeah, that would be really awesome.

juhp:
Very nice!

anitya/lib/backends/npmjs.py
```diff
@@ -92,3 +92,28 @@ def get_versions(cls, project):
...
         return data['versions'].keys()
+
+    @classmethod
+    def check_feed(cls):
+        ''' Return a generator over the latest 40 uploads to npmjs.org
+
+        by querying an weird JSON endpoint.
+        '''
+
+        url = 'https://registry.npmjs.org/-/all/static/today.json'
```
TomasTomecek: you can even directly connect to their database and get realtime feed: `curl -vL "https://skimdb.npmjs.com/registry/_changes?descending=true&include_docs=true&feed=continuous"`

TomasTomecek:
Ralph, this is an awesome piece of work!

My only concern is this: some of the calls return specific number of recent releases -- it's pretty easy to miss something e.g. when someone releases tenths (or hundreds) of packages in a short period of time.

anitya/lib/backends/npmjs.py
ralphbean: Say whaaaaat?  That's awesome.

TomasTomecek: It's just that it's not super-stable. It can timeout, return invalid responses, hang...

Issue #238, opened 2015-12-04
Version prefix is actually stripped everywhere


tomhughes:
The "version prefix" actually seems to be stripped everywhere it appears, not just at the start. See, for example https://release-monitoring.org/project/7680/ where `v2.0.0-develop.1` has been turned into `2.0.0-deelop.1`.

pypingou:
The fix is to be done in: https://github.com/fedora-infra/anitya/blob/master/anitya/__init__.py#L58

multipig:
Can you please assign it to me? I'll try to fix this.

pypingou:
Cool, welcome :)

multipig:
For the [project in the example](https://github.com/philsquared/Catch), the [Anitya page] (https://release-monitoring.org/project/7680/) shows versions as `2.0.0-deelop.1`, `1.5.4`, `1.5.2`, etc. but the corresponding [project's github tags page](https://github.com/philsquared/Catch/tags) shows versions serially as `v1.5.4`,`v1.5.3`, `v1.5.2`, etc

tomhughes:
The github page shows them in date order, with the v2.0.0 ones on page 2 currently (https://github.com/philsquared/Catch/tags?after=v1.3.3).

Also note that I have changed the anitya entry to use the custom backend now, with a regexp that excludes the -develop etc versions.

multipig:
Oh! I missed out on that. Also, the latest version seems to be [v2.0.0-develop-2] (https://github.com/philsquared/Catch/releases/tag/v2.0.0-develop-2). The regex should be  `class="tag-name">v(\d+(?:\.-\d+)*)</span`, right?

pypingou:
Closing this as fixed by @chaiku :)

PR #302, opened 2016-05-15
Issue #238 - Changed the __init__.py file to get version prefix correctly 

multipig:
As mentioned in the issue, for example the `v2.0.0-develop.1` was turned into `2.0.0-deelop.1`. The fix removes just the first appearance of the `version_prefix`.


pypingou:
Cool, thanks :)

PR #304, opened 2016-05-27
Fix #275: Improved search results by adding substring search.

### LINKED ISSUE #275: improve searching projects
to avoid such mistakes again #274,
please improve searching, so that if no exact matches are found, fall back to substring match

1. search for [`sodium`](https://release-monitoring.org/projects/search/?pattern=sodium)
2. it gives no exact matches, so retry search with [`sodium*`](https://release-monitoring.org/projects/search/?pattern=sodium*)
3. and if still no matches, retry with [`*sodium*`](https://release-monitoring.org/projects/search/?pattern=*sodium*)

of course could just change default pattern to be `*sodium*`

souravbadami:


pypingou:
hm, I know some people were un-happy with the behavior of the current search but on the other side this change makes sense to me

I'll let the PR open a little bit to see if more people are interested in commenting on this change and if not, we can merge :)

souravbadami:
@pypingou -- Sure, Thanks!

pypingou:
This should be long enough :)

PR #307, opened 2016-06-16
Enforce the homepage is an URL

pypingou:
Fixes https://github.com/fedora-infra/anitya/issues/265
Fixes https://github.com/fedora-infra/anitya/pull/266

Based on the work from @sudheesh001


puiterwijk:
:+1:

PR #321, opened 2016-07-29
Fixes/Cleans up the Docker file

PrahlM93:
This PR does the following:
* Fixes the Docker file so that it can be run on a Docker host and accessible externally
* Adds a host option to runserver.py (required for the port forwarding to work in Docker)
* Turns off app debug by default as this should not be run this way with Docker
* Copies the source code from the working directory instead of `git clone` in order to allow the developer to test local changes in Docker before committing code

Eventually, this should run behind Apache + mod_wsgi, but this at least gets Docker in a functioning state.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,13 +1,14 @@
 FROM fedora
 # cffi>=1.1.0->cryptography->fedmsg->-r
-RUN dnf install -y gcc python-devel libffi-devel openssl-devel git gcc-c++ redhat-rpm-config
+RUN dnf install -y gcc python-devel libffi-devel openssl-devel gcc-c++ redhat-rpm-config && \
+    dnf autoremove -y && \
+    dnf clean all -y
 
-RUN git clone https://github.com/fedora-infra/anitya.git /src
+COPY ./ /src
 WORKDIR /src
 
-# in Fedora we have ancient cffi atm
-# RUN dnf install -y python-cffi
 RUN pip install --user -r requirements.txt
 RUN python createdb.py
 EXPOSE 5000
-CMD ./runserver.py
+ENTRYPOINT python runserver.py --host '0.0.0.0'
+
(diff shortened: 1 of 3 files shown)


Dockerfile
```diff
@@ -1,13 +1,14 @@
 FROM fedora
 # cffi>=1.1.0->cryptography->fedmsg->-r
-RUN dnf install -y gcc python-devel libffi-devel openssl-devel git gcc-c++ redhat-rpm-config
+RUN dnf install -y gcc python-devel libffi-devel openssl-devel gcc-c++ redhat-rpm-config && \
+    dnf autoremove -y && \
+    dnf clean all -y
```
pypingou: Why these two lines? I'm not sure I see the added value

runserver.py
```diff
@@ -46,5 +50,5 @@
 
 
 from anitya.app import APP
-APP.debug = True
```
pypingou: I'm actually :-1: to this, runserver is meant for dev not for actually running the app, so debugging here is pretty useful :)

pypingou:
Thanks for working on this, I do want to point out though that so far the docker image was provided as a way for people to quickly and easily get anitya running to work on it.
It was not and is not meant to be use in prod, that's why it uses runserver which is definitely not meant for prod and why debugging is enabled.

If you want to run the application in prod, with or without docker, you should really look into wsgi or something like gunicorn.

Dockerfile
PrahlM93: This in theory is to make the container smaller by removing cached files.

pypingou:
Okido, thanks for these changes :)

PrahlM93:
@pypingou, thanks for the feedback. I'll probably work on getting mod_wsgi up in the container some time in the future.

pypingou:
@PrahlM93 would there be a way to have two docker files then? One for prod and the other (I guess the actual one) for dev?

PrahlM93:
@pypingou, I think so, but I'm thinking Vagrant would be more appropriate to set up a dev environment. Perhaps Docker becomes production, and Vagrant becomes the way to dev?

PR #322, opened 2016-08-01
WIP: Initial pass at Py3k compatibility

ncoghlan:
- Initial automated update with python-modernize
- Additional updates to get tests passing on 3.5
- Also checked with pylint --py3k
- Uses environment markers to install right variant of python[3]-openid

I also tried to add a tox configuration to test 2.6, 2.7 & 3.5, but anitya's current packaging isn't tox friendly (since it imports itself from `setup.py`), but the current source tree layout (without a `src` directory) starts to pose problems if you try to fix it by putting the VERSION metadata in a dedicated text file.


ncoghlan:
I mostly did this as a way of gaining some initial familiarity with the code base, but I also think it's a change worth working towards, and this gives some idea of the magnitude of updates needed.

Overall, they're relatively minor, but the main one that gives me pause is the use of regular expressions to scrape page content - the current PR keeps that as a text domain operation, which should be fine in most cases, but the decode step needs to be checked to see how robust it is likely to be against text encoding errors when retrieving pages for scraping.

Near term, though, my objective is to enable local testing and debugging under Python 3 without breaking deployment under Python 2, so more in-depth testing of the Unicode compatibility handling can wait for now.

anitya/lib/backends/pagure.py
```diff
@@ -59,7 +61,7 @@ def get_versions(cls, project):
         try:
             req = cls.call_url(url)
         except Exception as err:  # pragma: no cover
-            print err
+            print(err)
```
pypingou: I wonder if this is worth keeping since the other backend aren't doing it

requirements.txt
```diff
@@ -7,7 +7,9 @@ flask
 flask-openid
 flask-wtf
 markupsafe
-python-openid
+# TODO: Make next 2 lines conditional on python version
+# python-openid
+python3-openid
```
pypingou: So do we default to 2 or 3 ?

runtests.sh
```diff
@@ -1,4 +1,4 @@
 #!/bin/bash
 
-PYTHONPATH=anitya ./nosetests \
```
pypingou: The PYTHONPATH isn't needed?

tests/test_flask.py
```diff
@@ -61,7 +62,7 @@ def test_index(self):
         output = self.app.get('/')
         self.assertEqual(output.status_code, 200)
 
-        expected = """
+        expected = b"""
```
pypingou: An alternative to ``b'''`` is to replace ``output.data`` by ``output.get_data(as_text=True)``, don't know which you prefer though :)

tests/test_flask_api.py
```diff
@@ -39,6 +41,13 @@
 from anitya.lib.backends import REGEX
 from tests import Modeltests, create_distro, create_project, create_package
 
+# Py3 compatibility: UTF-8 decoding and JSON decoding become separate steps
+if sys.version_info[0] >= 3:
```
pypingou: Since we use six, we could use ``six.PY3``?

tests/test_flask_api.py
```diff
@@ -58,7 +67,7 @@ def test_api_projects(self):
 
         output = self.app.get('/api/projects')
         self.assertEqual(output.status_code, 200)
-        data = json.loads(output.data)
```
pypingou: cf comment above about ``.get_data()``

anitya/__init__.py
```diff
@@ -1,10 +1,19 @@
 # -*- coding: utf-8 -*-
 
+from __future__ import absolute_import
```
pypingou: While I guess it does not harm, is there really any interest in this one?

ncoghlan: It's not actually needed, I just haven't gone through to clean up things added by python-modernize that aren't strictly necessary - I'll add that to the TODO list.

anitya/lib/backends/pagure.py
ncoghlan: As in the print()? Deleting it would certainly work in making it compatible :)

requirements.txt
ncoghlan: Neither, I need to replace these lines with:

    python-openid; python_version < '3.0' 
    python3-openid; python_version >= '3.0' 

That way it will get the correct version regardless.

runtests.sh
ncoghlan: No, as the current directory gets added to sys.path implicitly, and that's all that's needed to make "anitya" importable. Actually adding the package directory to the path is potentially problematic, as it sets up the ability to import the same component twice under different names: http://python-notes.curiousefficiency.org/en/latest/python_concepts/import_traps.html#the-double-import-trap

tests/test_flask.py
ncoghlan: Staying entirely in the binary domain rather than decoding makes sense to me for these cases.

tests/test_flask_api.py
ncoghlan: I went with using the sys module mainly because this module already imports that. However, the suggestion to use `output.get_data(as_text=True")` as an inherently version independent alternative sounds more attractive, so I'll switch to that.

ncoghlan:
OK, I think the only thing still WIP about this now is the fact I need to squash it into a single commit with a nice commit message.

Since I didn't get tox working, I also need to check it in a Py2 venv

anitya/lib/backends/pagure.py
pypingou: Yes, I think we could just drop the print, other backends aren't printing either

ncoghlan:
Yay, I figured out how to get tox running (and passing the tests under both Python 2.7 and 3.5 - I didn't check Python 2.6, since I'm testing on Fedora). There does seem to be a problem with tox's handling of environment markers, but I was able to work around that: https://bitbucket.org/hpk42/tox/issues/236/tox-must-create-the-source-distribution#comment-29617355

I also update the README to cover running the test suite, both directly and via tox.

pypingou:
Alright, this looks all good to me, thanks for working on it, it wasn't a small change!

PR #324, opened 2016-08-02
Modifies the Dockerfile for production and adds Vagrant support

PrahlM93:
* Modifies the Dockerfile for production by using Apache and mod_wsgi
* Adds Vagrant support
* Moves the default SQLite database to the repo root due to potential permission issues

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -21,7 +21,7 @@ Notifications service).
 
 :Github page: https://github.com/fedora-infra/anitya
 
-Hacking
+Development
 -------
 
 Anitya deployment is currently only supported on Python 2.x, but local
@@ -29,11 +29,23 @@ development can use either Python 2.x or 3.x. The following steps should
 all work regardless of which runtime you have your virtual environment
 configured to use.
 
+Vagrant
+```````
+
+To run Anitya in a Vagrant guest, simply run::
+
+    $ sudo vagrant up
+
+You may then access Anitya on your host at::
+
+    http://127.0.0.1:8080
+
+
 virtualenv
 ``````````
 
 Here are some preliminary instructions about how to stand up your own instance
-of anitya. We'll use a virtualenv and a sqlite database and we'll install
+of Anitya. We'll use a virtualenv and a sqlite database and we'll install
 our dependencies from the Python Package Index (PyPI).  None of these are best
 practices for a production instance, but they will do for development.
 
@@ -90,12 +102,29 @@ Open your browser and visit http://localhost:5000 to check it out.
 
 Docker
 ``````
-
 To build the Docker image::
 
     $ cd anitya/
-    $ docker build --tag=anitya .
+    $ docker build -t anitya .
 
-To run the container::
+To run the container with a disposable SQLite database::
 
-    $ docker run -d -p 5000:5000 anitya
+    $ docker run -e DB_URL='sqlite:////opt/anitya/anitya.db' -d -p 80:80 anitya
+
+
+Deployment
+-------
+
+Docker
+``````
+To build the Docker image::
+
+    $ cd anitya/
+    $ docker build -t anitya .
+
+To run the container, execute the command below. Be sure to replace the value of DB_URL with the URL to connect to
+your production database. Also ensure to replace SECRET_KEY with a random string (preferably hex values) that is the
+same on every deployment of Anitya, as this is used for session management::
+
+    $ docker run -e DB_URL='db_type://user:password@server.domain.local:3306/database_name' \
+                 -e SECRET_KEY='123456789abcdef123456789' -d -p 80:80 anitya
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -1,0 +1,29 @@
+# -*- mode: ruby -*-
+# vi: set ft=ruby :
+
+$script = <<SCRIPT
+    dnf install -y \
+        python-flask \
+        python-flask-wtf \
+        python-flask-openid \
+        python-wtforms \
+        python-openid \
+        python-docutils \
+        python-dateutil \
+        python-markupsafe \
+        python-bunch \
+        python-straight-plugin \
+        python-setuptools \
+        python-sqlalchemy \
+        fedmsg
+    cd /opt/anitya/src/ && python createdb.py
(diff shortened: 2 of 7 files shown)


pypingou:
One thing that bugs me a little is the use of virtual env. I see the benefit but that does mean that spinning a new docker will give you a new environment, so spinning two images at say 24h interval may results in two different images.
Is that expected in the docker world? (I don't use docker much so I do not know what the best practices are)

anitya/default_config.py
```diff
@@ -18,7 +18,8 @@
 SECRET_KEY = 'changeme please'
 
 # URL to the database
-DB_URL = 'sqlite:////var/tmp/anitya-dev.sqlite'
+repo_root = path.abspath(path.dirname(path.dirname(__file__)))
+DB_URL = 'sqlite:///{0}/anitya.db'.format(repo_root)
```
pypingou: I'm also curious why the change in location here? Especially by default as this can be overridden as much as people like :)

PrahlM93: The main reason was that if this wasn't overridden, it could cause potential permission issues such as Apache not being able to access the database. I just thought it was a friendlier default value.

PrahlM93:
@pypingou, good point. Whenever I've Dockerized Flask apps, they've always had specific versions in requirements.txt. In this case, however, the versions are not specified.

We can either put specific versions that Anitya supports in requirements.txt (some on the Fedora team dislike this idea), or I can try to use RPMs to install the Python dependencies. This would still provide potential inconsistencies, but it'd be at a less frequent rate than getting it straight from Pypi.

PrahlM93:
@pypingou, I think this latest commit addresses your concerns. Please take a look when you can.

anitya/default_config.py
pypingou: apache should always have permission to write on ``/var/tmp``.

Since you figured out the way to override the DB_URL, I think I would prefer to keep the default as it is now.

PrahlM93: @pypingou, I had trouble with Apache reading the database in that location. I was able to get it to work when the database was under /opt or /var/www though.

Anyways, I returned it to the previous value in my latest commit.

pypingou:
Ok, this is looking good to me, I believe @puiterwijk wanted to have a look at this as well (related to the SECRET_KEY iirc).

I'll wait for his input and then we can merge :)

Vagrantfile
```diff
@@ -0,0 +1,15 @@
+# -*- mode: ruby -*-
+# vi: set ft=ruby :
+
+$script = <<SCRIPT
+    dnf install -y python-flask python-flask-wtf python-flask-openid python-wtforms python-openid python-docutils python-dateutil python-markupsafe python-bunch python-straight-plugin python-setuptools python-sqlalchemy fedmsg
+    cd /opt/anitya/src/ && python createdb.py
+SCRIPT
+
+Vagrant.configure("2") do |config|
+  config.vm.box = "box-cutter/fedora23"
```
puiterwijk: Why 23 here, if you're using fedora:24 for docker?

puiterwijk:
Just one question, ack for me otherwise.

puiterwijk:
So, one thing I'm worried about is the fact that the secret key gets generated inside the container and doesn't get synced out to the other containers. This means that if you have two of them behind a load balancer, their sessions won't be trusted by the other, meaning people will be logged out again every request.

Vagrantfile
PrahlM93: @puiterwijk, there isn't a fedora24 Vagrant image yet from a reputable source. I'm waiting on this PR to go through:
https://github.com/boxcutter/fedora/pull/20

PrahlM93:
@puiterwijk, good point. I had changed it to be generated within the container in an attempt to protect the SECRET_KEY. I will move it back to the way it was.

PrahlM93:
@puiterwijk, this latest commit should address your concern.

pypingou:
But isn't the SECRET_KEY still created inside the container here?

PrahlM93:
@pypingou, it's created in the Docker image, and not in the container, therefore, for every container you spin up, they would have the same SECRET_KEY. The only time the SECRET_KEY is regenerated is when the Docker image is rebuilt. The downside to this is that during upgrades, the login sessions would be invalidated, and also if the user uploaded the built image to a Docker registry, the SECRET_KEY would no longer be public.

Perhaps the middle ground is if the user supplies a SECRET_KEY environment var on the first time running the container, that value is used for the SECRET_KEY instead. If that environment var is not provided, a SECRET_KEY is generated. I'll go implement that now.

pypingou:
> SECRET_KEY would no longer be public.

no longer public or no longer private?

Sorry for the stupid questions, my docker skills are low to say the least :)

PrahlM93:
@pypingou, private, just fixed it :)

PrahlM93:
@pypingou @puiterwijk, I believe I have addressed your latest concerns in the latest commit. Once you approve it, I'll squash the commits to clean up the commit history for this PR. Then you can go ahead and merge it.

README.rst
```diff
@@ -122,6 +122,6 @@ To build the Docker image::
     $ cd anitya/
     $ docker build -t anitya .
 
-To run the container (replace the value of DB_URL with the URL to connect to your production database)::
+To run the container, execute the command below. Be sure to replace the value of DB_URL with the URL to connect to your production database. Also ensure to replace SECRET_KEY with a 24 character hex value that is the same on every deployment of Anitya (used for session management)::
```
pypingou: I don't think there is a limit on the size of SECRET_KEY.

Also, nitpicking but could you cut the lines at ~80 chars? thanks :)

pypingou:
Looks fine to me

@puiterwijk do you want to have a final look at it as well?

PrahlM93:
@pypingou: I just fixed the latest issues you pointed out. I squashed the commits, and now it's ready to merge once you are both ok with it.

pypingou:
Let's see if @puiterwijk has more comments, otherwise I'll be merging this :)

pypingou:
Alright left's merge, we can always re-adjust later if needed :)

PR #327, opened 2016-08-03
Query by ecosystem

ncoghlan:
This is a draft PR that adds Ecosystems to the data model, as a subset of the defined backends.

The full PR will also add the ability to query by ecosystem and the package name within that ecosystem.

### DIFF
--- a/anitya/api.py
+++ b/anitya/api.py
@@ -550,3 +550,66 @@ def api_get_project_distro(distro, package_name):
     jsonout = flask.jsonify(output)
     jsonout.status_code = httpcode
     return jsonout
+
+@APP.route('/api/by_ecosystem/<ecosystem>/<project_name>/', methods=['GET'])
+@APP.route('/api/by_ecosystem/<ecosystem>/<project_name>', methods=['GET'])
+def api_get_project_ecosystem(ecosystem, project_name):
+    '''
+    Retrieve a project from a given ecosystem
+    -------------------------------
+    Retrieves a project in an ecosystem via the name of the ecosystem
+    and the name of the project as registered with Anitya.
+
+    ::
+
+        /api/by_ecosystem/<ecosystem>/<project_name>
+
+    Accepts GET queries only.
+
+    :arg ecosystem: the name of the ecosystem (case insensitive).
+    :arg project_name: the name of the project in Anitya.
+
+    Sample response:
+
+    ::
+
+      {
+        "backend": "pypi",
+        "created_on": 1409917222.0,
+        "homepage": "https://pypi.python.org/pypi/six",
+        "id": 2,
+        "name": "six",
+        "packages": [
+          {
+            "distro": "Fedora",
+            "package_name": "python-six"
+          }
+        ],
+        "regex": null,
+        "updated_on": 1414400794.0,
+        "version": "1.10.0",
+        "version_url": null,
+        "versions": [
+          "1.10.0"
+        ]
+      }
+
+    '''
+
+    project = anitya.lib.model.Project.by_name_and_ecosystem(
+        SESSION, project_name, ecosystem)
+
+    if not project:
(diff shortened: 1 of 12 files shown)


ncoghlan:
Note: this PR currently includes the changes from #322 as well. Once that is squashed & merged, I'll rebase this one so it shows just the changes specific to this PR.

ncoghlan:
OK, this should be feature complete now, but remains WIP pending the merging of the 3.x compatibility changes and the subsequent rebasing of this PR.

One point I think is worth calling out is the use of ".one()" for the model query: for backends that define an ecosystem, there should be *at most one* project with a given name. ".one()" enforces that, while using ".first()" the way the distro query API does wouldn't.

pypingou:
Hm looks like it'll need to be rebased so to clean the py3 commits now that #322 has been merged

pypingou:
> One point I think is worth calling out is the use of ".one()" for the model query: for backends that define an ecosystem, there should be at most one project with a given name. ".one()" enforces that, while using ".first()" the way the distro query API does wouldn't.

I agree one the semantic but technically the ``at most`` is covered at the database level and ``.first()`` returns ``None`` if it doesn't find anything while ``.one()`` raises an exception. Therefore I find the use of the former easier to deal with than the later.

ncoghlan:
The particular use of `.one()` in this patch recreates the `NoResultFound` -> `return None` conversion for the `Project.by_name_and_ecosystem()` convenience API, while letting the duplicate record exception escape as an internal server error (since it indicates either a bug in the query definition, or in the database schema definition). The tests then implicitly rely on that characteristic by setting up cases where the project name queried is duplicated in multiple ecosystems and checking the server gives the right answer rather than a 500 error.

ncoghlan:
OK, squashed and rebased with the WIP flag removed.

However, I just realised there's a bug in the way ecosystems are loaded - at the moment, if a previously entered plugin has an ecosystem name added, the plugin loader will miss that. The "Add new ecosystems" step needs to be separate from the "Add new backends" step, and calculated from the whole list of plugins, not just the new ones.

ncoghlan:
PR should now correctly handling existing plugins gaining an Ecosystem definition, but I couldn't think of any straightforward way to automatically test that, as it involves making a change to the plugin code and seeing if the loader picks it up.

The best I thought of was to:

1. Load the plugins
2. Set `anitya.lib.backends.custom.CustomBackend.ecosystem_name = "Test Ecosystem"`
3. Load the plugins again to make sure the new dummy ecosystem shows up

It's not particularly clean, but it should exercise the code path of interest, and the custom backend should never gain a legitimately defined ecosystem.

ncoghlan:
Ah, I remembered the problem with the implied uniqueness constraint on "name per ecosystem", and why this patch wasn't able to fully enforce that at the DB level: to minimise the ripple effect of the patch, ecosystems are implied by the Project backend rather than being a new project property, and the uniqueness constraint only applies for projects that use backends that define an ecosystem.

if you use the custom backend (for example), as many projects as you like can share a name, so long as their home page is different.

We could *add* the constraint to the model by defining a mapping from Ecosystems directly back to Project names (rather than project IDs), and making that unique, but I'm not sure of the best way to do that.

ncoghlan:
OK, I updated the PR to actually test the "new ecosystem defined" case. However, talking to @pypingou on IRC, he pointed out that much of the complexity in this patch comes from the original notion of having the choice of backend plugin imply a particular ecosystem.

We can likely make both this PR and the follow-on implementation for #329 a lot simpler by instead having a *second set of plugins* (e.g. "anitya.lib.ecosystems"), and have those plugins define both a default backend and a default homepage to be used for projects in that ecosystem.

ncoghlan:
This was replaced by https://github.com/fedora-infra/anitya/pull/330

PR #330, opened 2016-08-09
Add upstream ecosystems to model

ncoghlan:
- model structured ecosystems (like PyPI or Maven Central)
  separately from version querying backends
- module level plugin APIs still just work with backend plugins
  by default
- new BACKEND_PLUGINS and ECOSYSTEM_PLUGINS plugin manager
  instances allow access to the particular kinds of plugin
- module level "load_all_plugins" API (implicitly called by the
  old "load_plugins" API) reports both kinds of plugin to caller
- ecosystem plugins can define a bidirectional "default"
  relationship with a named backend plugin
- new API endpoint allows projects to be looked up by ecosystem
  in addition to by numeric ID and by distro
- project names must be unique within a given ecosystem

### DIFF
--- a/anitya/lib/model.py
+++ b/anitya/lib/model.py
@@ -175,7 +175,10 @@ class Ecosystem(BASE):
 
     @classmethod
     def by_name(cls, session, name):
-        return session.query(cls).filter_by(name=name).first()
+        try:
+            return session.query(cls).filter_by(name=name).one()
+        except NoResultFound:
+            return None
 
     get = by_name
 
@@ -434,7 +437,6 @@ class Project(BASE):
                     .filter(Ecosystem.name==ecosystem).one())
         except NoResultFound:
             return None
-
 
     @classmethod
     def all(cls, session, page=None, count=False):
--- a/anitya/lib/model.py
+++ b/anitya/lib/model.py
@@ -19,6 +19,8 @@ import time
 
 import sqlalchemy as sa
 from sqlalchemy.ext.declarative import declarative_base
+from sqlalchemy.orm import relationship
+from sqlalchemy.orm.exc import NoResultFound
 
 import anitya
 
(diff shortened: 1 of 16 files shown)


anitya/lib/model.py
```diff
@@ -136,6 +138,35 @@ class Backend(BASE):
...
+                                     back_populates="default_backend")
+
+    @classmethod
+    def all(cls, session):
+        query = session.query(cls).order_by(cls.name)
+        return query.all()
+
+    @classmethod
+    def by_name(cls, session, name):
+        return session.query(cls).filter_by(name=name).first()
```
pypingou: Here we use `.first()` while below to used `.one()`, should we be consistent ?

anitya/lib/model.py
```diff
@@ -385,6 +426,17 @@ def by_name_and_homepage(cls, session, name, homepage):
...
+    def by_name_and_ecosystem(cls, session, name, ecosystem):
+        try:
+            return (session.query(cls)
+                    .filter_by(name=name)
+                    .join(Project.ecosystem)
+                    .filter(Ecosystem.name==ecosystem).one())
+        except NoResultFound:
+            return None
+
+
```
pypingou: pep8 will complain about those 2 lines here

pypingou:
2 small comments, but nothing blocking the merge imho, shall I just merge or do you want to address them?

ncoghlan:
I'll quickly push the style/consistency cleanups

pypingou:
Thanks!

Oh, that will need an alembic migration script no?

ncoghlan:
Ah, you're right - I thought of that yesterday, but then I forgot again.

ncoghlan:
Alembic mapping added, as well as instructions for autogenerating mappings based on differences between the current model in the database and that in the source code.

pypingou:
Looks good to me, thanks! :)

PR #331, opened 2016-08-09
Open XML as utf-8 file

bkabrda:
I was trying to run the cron job with --check-feed, which fetched the PyPI RSS feed. At that moment, one of the packages in the feed was [1], summary of which contains non-ascii chars, thus the XML parsing failed. This PR fixes that error.

[1] https://pypi.python.org/pypi/simcu.synapse/0.1.1


pypingou:
Cool, thanks!

PR #335, opened 2016-08-11
Make anitya_cron use thread pool instead of process pool

bkabrda:
So this is probably the last error that I've hit with the cron job using `--check-feed`. The explanation is kind of long-ish (see the comment in code), but I still think it's worth putting the long comment into code, so that everyone knows why `multiprocessing.dummy` is used right away.
FTR, I think using `multiprocessing.dummy.Pool` (which is actually a thread pool) is actually ok here, since threads will lift GIL while fetching URLs, so the performance drop shouldn't be significant (although there probably _is_ some performance drop).

### DIFF
--- a/files/anitya_cron.py
+++ b/files/anitya_cron.py
@@ -3,7 +3,14 @@
 
 import sys
 import logging
-import multiprocessing
+# We need to use multiprocessing.dummy, since we use the Pool to run
+# update_project. This in turn uses anitya.lib.backends.BaseBackend.call_url,
+# which utilizes a global requests session. Requests session is not usable
+# with multiprocessing, since it would need to share the SSL connection between
+# processes (see https://stackoverflow.com/q/3724900#3724938).
+# multiprocessing.dummy.Pool is in fact a Thread pool, which works ok
+# with a global shared requests session.
+import multiprocessing.dummy as multiprocessing
 
 import anitya
 import anitya.app


pypingou:
This change surprises me a little. Is it related to recent changes to the cron job? As currently the cron is running fine (except for getting stuck every now and then)

bkabrda:
No, I don't think is related. Are you by any chance running cron job on Python that doesn't have SSL certificate validation on by default, e.g. Python < 2.7.9 [1]? This might be why you're not getting any issues. I'm using Python 2.7.12 and I'm getting these errors when using `multiprocessing.Pool`:

```
2016-08-12 05:46:13,378 - anitya - ERROR - AnityaError catched:
  File "/src/anitya/__init__.py", line 75, in check_release
    up_version = backend.get_version(project)
  File "/src/anitya/lib/backends/pypi.py", line 45, in get_version
    raise AnityaPluginException('Could not contact %s' % url)
AnityaPluginException: Could not contact https://pypi.python.org/pypi/pymuv/json
```

When I print the actual exception that caused the above one to be thrown, I get:

```
exception type: requests.exceptions.SSLError
exception: [SSL: DECRYPTION_FAILED_OR_BAD_RECORD_MAC] decryption failed or bad record mac (_ssl.c:1754)
```

This is exactly what [2] is about. Can you try running this on e.g. Fedora's Python to see if you can reproduce the issue?

[1] https://www.python.org/dev/peps/pep-0493/#abstract
[2] http://stackoverflow.com/questions/3724900/python-ssl-problem-with-multiprocessing

pypingou:
So in prod we are indeed using: python-2.7.5-34.el7.x86_64 I'll try on Fedora but the number of projects failing to update make the output hard to read

jeremycline:
> No, I don't think is related. Are you by any chance running cron job on Python that doesn't have SSL certificate validation on by default, e.g. Python < 2.7.9 [1]? This might be why you're not getting any issues.

I don't think that should affect this one way or another, and I'm really surprised this is working at all (which makes me think maybe I'm missing something).

``requests`` sessions are sort of thread safe (there a few exceptions IIRC), but I am fairly certain they're _not_ multiprocess safe. SSL/TLS or not, sharing sockets among processes is not going to end well. Maybe prod running with a pool size of 1? I can see that working.

As for what to do about it, it might be better to adjust how the requests session is constructed so each process gets its own session rather than dropping support for multiprocessing. What do you think, @pypingou?

jeremycline:
Okay, I looked into this and figured out what's happening. When you provide the ``--check-feed`` flag, this ultimately results in the backend's ``check_feed`` method being called, which in some (or maybe all) cases calls ``call_url`` which causes requests to open one or more sockets and leave them open in a connection pool.

When multiprocessing is used, each process gets a copy of the process memory, including the open sockets. ``urllib3``, the library underneath requests, hands the same socket back to all the various processes because they all have copies of the same socket. Then 10 or more processes all try to send/receive a new HTTP request over the same socket in parallel. Disaster follows, obviously.

I've given this a bit of thought and I'm happy to switch to a thread pool for now, since it's mostly going to be waiting on IO anyway. I'm going to file an issue to refactor this a bit in the future, but I don't think that's something we need to bite off right now.

PR #357, opened 2016-10-26
Simplify extended pattern logic

akien-mga:
As mentioned in #275.


akien-mga:
As far as I can tell even the case where `pattern` is `''` should be handled just fine, as it will satisfy the first condition and not the second one (so it won't become `'**'`, just `'*'`).

pypingou:
Thanks :)

PR #368, opened 2016-10-29
Update anitya's mail_logging to be compatible with old and new psutil

pypingou:
The API of python-psutil has changed between 1.0 and 2.0/3.0 and with
this change we are compatible with all of them.


sayanchowdhury:
:+1: for me

pypingou:
Thanks!

PR #369, opened 2016-10-30
Allow user to see the logs related to their openid or their email

pypingou:
We recently moved from using the user's email to use the user's openid
identifier as well identifier :)
This broke the logs page since non-admin can only see their logs, which
now means: see logs related to their openid. So all their old logs are
no longer accessible.
To fix this, we can simply ask to see all the logs related to either the
user's openid or the user's email.
That's what this commit does.

### DIFF
--- a/anitya/admin.py
+++ b/anitya/admin.py
@@ -299,7 +299,6 @@ def delete_project_version(project_id, version):
 @login_required
 def browse_logs():
 
-    email = None
     if is_admin():
         user = flask.request.args.get('user', None)
     else:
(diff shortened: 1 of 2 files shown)


anitya/admin.py
```diff
@@ -299,10 +299,11 @@ def delete_project_version(project_id, version):
 @login_required
 def browse_logs():
 
+    email = None
```
akien-mga: I'm not really familiar with the code base yet, but isn't this variable unused?

pypingou: yes it is, left over from a former design, let me remove it

pypingou:
Thanks for the review :)

PR #370, opened 2016-10-30
Strip the version prefix only after it's no longer a tuple

pypingou:


anitya/lib/backends/__init__.py
```diff
@@ -329,15 +329,16 @@ def get_versions_by_regex_for_text(text, url, regex, project):
...
-        version = version.lstrip(project.version_prefix)
         upstream_versions[index] = version
 
         # If the version retrieved is a tuple, re-constitute it
         if type(version) == tuple:
             version = ".".join([v for v in version if not v == ""])
             upstream_versions[index] = version
 
+        # Strip the version_prefix early
+        version = version.lstrip(project.version_prefix)
```
akien-mga: Doesn't this mean that `upstream_versions[index]` won't have the stripped version? In case it's not the intended behaviour, then this would be better:
```
for index, version in enumerate(upstream_versions):
    # If the version retrieved is a tuple, re-constitute it
    if type(version) == tuple:
        version = ".".join([v for v in version if not v == ""])

    # Strip the version_prefix if defined
    version = version.lstrip(project.version_prefix)

    upstream_versions[index] = version

    if " " in version:
[...]
```

pypingou: hm, indeed, not sure why I didn't move that line when I moved the first two.

Thanks for catching this!

pypingou:
Thanks for the review :)

PR #376, opened 2016-11-12
Report the reason fetching a url failed in the folder backend

jeremycline:
It was noted in https://github.com/fedora-infra/anitya/issues/338 that
FTP urls fail. In my tests, FTP connections to the host mentioned in the
issue fail occasionally when the connection is reset by the host. This
does not make the FTP call more robust, but it does report the actual
error that occurred.


pypingou:
Cool :+1: for me

PR #377, opened 2016-11-15
Add a timeout to HTTP requests

jeremycline:
By default, requests waits forever for a response from the server. If
the server is very slow, or something else weird happens, the request
will hang indefinitely. This commit sets the timeout[0] to be 30
seconds, which matches the FTP timeout.

[0] http://docs.python-requests.org/en/master/user/quickstart/#timeouts

### DIFF
--- a/anitya/lib/backends/__init__.py
+++ b/anitya/lib/backends/__init__.py
@@ -293,7 +293,8 @@ class BaseBackend(object):
                 'From': from_email,
             }
 
-            return http_session.get(url, headers=headers, verify=not insecure)
+            return http_session.get(url, headers=headers, timeout=60,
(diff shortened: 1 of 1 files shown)


pypingou:
Looks good to me, I'd propose to make it little higher, maybe 45 or 60?
But that's neat picking, I'm :+1: as is if you prefer.

pypingou:
I guess you liked the idea of 60 seconds :)

:+1: for me

jeremycline:
I bumped it up to 60. If we see lots of timeouts, it's possible to configure requests to automatically retry ``n`` times, or set timeouts individually for the ``connect`` and ``read`` calls.

PR #378, opened 2016-11-16
Configure the Vagrant environment with an Ansible role

jeremycline:
This adds an Ansible role to configure an Anitya development
environment. The Vagrant-provisioned VM is configured using the role.


pypingou:
:+1: for me (untested though)

PR #379, opened 2016-11-19
Add a style test to enforce PEP8 compliance.

jeremycline:
This adds a test to enforce PEP8 compliance and fixes all the violations
in the anitya package. This does not fix the unit tests and style is not
currently enforced there. They can be incrementally fixed and added to
the enforced paths list.

### DIFF
--- a/anitya/lib/__init__.py
+++ b/anitya/lib/__init__.py
@@ -9,12 +9,11 @@
 anitya internal library.
 """
 
-
-__requires__ = ['SQLAlchemy >= 0.7']
-import pkg_resources
-
 import logging
 import sys
+
+__requires__ = ['SQLAlchemy >= 0.7']  # NOQA
+import pkg_resources  # NOQA
 
 import sqlalchemy as sa
 from sqlalchemy import create_engine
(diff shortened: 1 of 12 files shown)


anitya/lib/__init__.py
```diff
@@ -9,9 +9,7 @@
 anitya internal library.
 """
 
-
-__requires__ = ['SQLAlchemy >= 0.7']
-import pkg_resources
```
pypingou: I believe the import is needed for the ``__requires__`` to work, no?

anitya/ui.py
```diff
@@ -252,7 +252,7 @@ def projects_search(pattern=None):
             SESSION, pattern=get_extended_pattern(pattern), count=True)
     else:
         projects_count = anitya.lib.model.Project.search(
-            SESSION, pattern=pattern, distro=distroname, count=True)
+            SESSION, pattern=pattern, count=True)
```
pypingou: This is an odd left over

pypingou:
2 questions otherwise looks good

jeremycline:
Good catch on both of those, I've never used pkg_resources before and didn't realize importing it had a side-effect.

pypingou:
:+1: for me

PR #381, opened 2016-11-21
replace lstrip() with proper chopping

ignatenkobrain:
Doing lstrip() is wrong in such situation, e.g. "libgphoto2-2.1.15.tar.gz".lstrip("libgphoto2-") -> ".1.15.tar.gz"

### DIFF
--- a/anitya/lib/backends/__init__.py
+++ b/anitya/lib/backends/__init__.py
@@ -336,7 +336,8 @@ def get_versions_by_regex_for_text(text, url, regex, project):
             version = ".".join([v for v in version if not v == ""])
 
         # Strip the version_prefix early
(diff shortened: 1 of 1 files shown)


jeremycline:
I always forget ``lstrip`` has that behavior. Can I convince you to write a unit test for this :smile:? If you don't have time that's fine, I'm happy to add one.

jeremycline:
I added some unit tests in https://github.com/fedora-infra/anitya/commit/e25c95cecb3efae8c9eef703c8454f903192e02c and made one modification since the prefix can be ``None``. Thanks so much for the PR!

pypingou:
I didn't know that behavior, thanks! :)

PR #391, opened 2016-11-29
Add a Sphinx documentation project

jeremycline:
Start a documentation project for Anitya.


docs/conf.py
```diff
@@ -0,0 +1,348 @@
...
+project = 'Anitya'
+copyright = '2016, Red Hat, Inc. and others'
+author = 'Red Hat, Inc. and others'
+
+# The version info for the project you're documenting, acts as replacement for
+# |version| and |release|, also used in various other places throughout the
+# built documents.
+#
+# The short X.Y version.
+version = app.__version__
```
bowlofeggs: I've always interpreted this to mean that it should only include the X.Y, and not the Z. So version 3.1.2 should just be 3.1 here. I think it's because docs are typically maintained for X.Y branches, but not typically for each .Z.

docs/contributing.rst
```diff
@@ -0,0 +1,67 @@
...
+you during the code review process.
+
+You can run the unit tests by running ``python setup.py test`` in the root of the
+repository.
+
+
+Development environment
+-----------------------
+
+The development environment is set up using `Ansible <https://www.ansible.com/>`_. You can use
```
bowlofeggs: It's pronounced "Ansible".

jeremycline: You make me sad

bowlofeggs: hahahah

PR #393, opened 2016-12-01
Return 4XX codes in error cases for /projects/new

### LINKED ISSUE #246: http status codes for failed form submissions?
currently if you submit a form with some bogus data, it will return you to the same form with some helpful error messages about what you messed up.

However, the status code on the http response is '200' which is misleading for scripts.  I think we can make that code '400' and still return the form to the user.

jeremycline:
Previously, HTTP 200 was returned regardless of the result of a POST
request. This causes requests that fail form validation to return HTTP
400. In addition, attempts to create duplicate projects now returns HTTP
409 (Conflict).

fixes #246

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -1,6 +1,14 @@
 Changes
 =======
 
+dev (master)
+------------
+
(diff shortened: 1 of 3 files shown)


pypingou:
Looks good to me

PR #394, opened 2016-12-08
Projects with the folders backend can use insecure HTTPS

### LINKED ISSUE #386: type=folder: ERROR: Could not call : "https://www.soft-switch.org/downloads/spandsp/" of "spandsp"
https://release-monitoring.org/project/12600/
```
Unable to retrieve the latest version from upstream!
ERROR: Could not call : "https://www.soft-switch.org/downloads/spandsp/" of "spandsp"
```

jeremycline:
Projects have a flag to allow for insecure HTTPS connections where the
signature on the certificate presented by the server is not verified.
This enables users to set that flag when creating or editing projects
using the folder backend. The reason for this is that the backend URL
can be any page on the Internet and there's no guarantee that user is
serving a certificate signed by a trusted CA (or they might have messed
up their configuration).

fixes #386

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -6,6 +6,12 @@ dev (master)
 
 * Return 4XX codes in error cases for /projects/new rather than 200 (Issue #246)
 
+* Allow projecst using the "folder" backend to make insecure HTTPS requests
+  (Issue #386)
+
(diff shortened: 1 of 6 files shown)


pypingou:
Cool :+1:

pypingou:
Still :+1:

PR #395, opened 2016-12-21
Insert the repo path into the Python path for the docs

jeremycline:
This is motivated by readthedocs, which is having trouble installings
the package since it doesn't install with pip and fails to understand
the ``python_version < '3.0'`` directive.

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -31,6 +31,11 @@
 # extensions coming with Sphinx (named 'sphinx.ext.*') or your custom
 # ones.
 
+import os
+import sys
+
+sys.path.insert(0, os.path.abspath('../'))  # NOQA
+
 from anitya import app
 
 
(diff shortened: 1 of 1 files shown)


jeremycline:
https://jcline-anitya.readthedocs.io/en/insert-path-docs/ is a build of this branch :shipit:

docs/conf.py
```diff
@@ -31,14 +31,18 @@
...
+sys.path.insert(0, os.path.abspath('../'))  # NOQA
+
 from anitya import app
 
 
 extensions = [
     'sphinx.ext.autodoc',
     'sphinx.ext.doctest',
     'sphinx.ext.viewcode',
-    'sphinx.ext.githubpages',
```
bowlofeggs: I recommend making this change in a separate commit since it's a separate concern.

PR #397, opened 2017-01-03
Stop passing around a single logging instance

jeremycline:
This configures logging and stops passing around a single logger
instance.

### DIFF
--- a/anitya/app.py
+++ b/anitya/app.py
@@ -10,12 +10,11 @@ User-facing Flask routes should be placed in the ``anitya.ui`` module and API
 routes should be placed in ``anitya.api``.
 """
 
-
 import functools
 import logging
+import logging.config
 import logging.handlers
 import os
-import sys
 
 import flask
 from bunch import Bunch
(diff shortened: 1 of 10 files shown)


anitya/app.py
```diff
@@ -38,22 +37,46 @@
 APP.oid = OpenID(APP)
 
 # Set up the logging
-if not APP.debug:
-    APP.logger.addHandler(anitya.mail_logging.get_mail_handler(
-        smtp_server=APP.config.get('SMTP_SERVER', '127.0.0.1'),
-        mail_admin=APP.config.get('MAIL_ADMIN', 'admin@fedoraproject.org')
+logging_config = {
```
pypingou: What about placing this in the configuration file?

pypingou:
👍 with one question

anitya/app.py
jeremycline: While that does allow for maximum configuration, when I wrote this I thought that maybe it's more user-friendly to have one log level setting. I don't mind either way so if you think it's better to expose all this in the config I can switch it.

pypingou: I don't mind either way, it's just that I've often seen this dict provided as configuration.

Anyway, we can always start like this and reconsider later if needed.

pypingou:
Still 👍

PR #399, opened 2017-01-11
Correctly disable fedmsg signatures for local dev

ncoghlan:
The setting to turn off signature validation during development
is `validate_signatures` rather than `validate_messages`.

### DIFF
--- a/fedmsg.d/fedmsg-config.py
+++ b/fedmsg.d/fedmsg-config.py
@@ -25,5 +25,5 @@ config = dict(
     zmq_strict=False,
 
     sign_messages=False,
-    validate_messages=False,
+    validate_signatures=False,
 )


ncoghlan:
An alternative option would be to explicitly recommend the use of `--no-validate` option when running `fedmsg-tail` against the local instance (at the moment there aren't any docs in the README for local use of fedmsg, so recommending this would necessarily be part of adding a new section on that).

ncoghlan:
https://github.com/fedora-infra/anitya/pull/406 implements the second option (recommending the use of --no-validate when running fedmsg-tail locally)

jeremycline:
Even though our new documentation explicitly disables signature checking, I'm inclined to merge this since that fedmsg configuration is marked explicitly for development use. On the other hand, someone who reads that documentation might expect signatures to be checked if they omit that CLI flag, so maybe it's better to set it to ``True`` by default?

PR #402, opened 2017-01-11
Data migration for by_ecosystem query

ncoghlan:
This migration allows pre-existing projects to be looked
up via the by_ecosystem API endpoint.

### DIFF
--- a/alembic/versions/9c29da0af3af_populate_ecosystem_data.py
+++ b/alembic/versions/9c29da0af3af_populate_ecosystem_data.py
@@ -1,0 +1,36 @@
+"""Populate missing ecosystem data
+
+Revision ID: 9c29da0af3af
+Revises: 921c612ba0da
+Create Date: 2017-01-11 22:23:58.497998
+
+"""
+
+# revision identifiers, used by Alembic.
+revision = '9c29da0af3af'
+down_revision = '921c612ba0da'
+
+from alembic import op
+import sqlalchemy as sa
+
+
+def upgrade():
+    # We use a subquery instead of an UPDATE FROM with a table join
+    # due to the fact that SQLite doesn't allow joins in update statements
+    op.execute("""
+        UPDATE projects
+        SET ecosystem_name=(
+            SELECT ecosystems.name
+            FROM projects AS subquery_projects
+                INNER JOIN ecosystems ON
+                    subquery_projects.backend = ecosystems.default_backend_name
+            WHERE projects.id = subquery_projects.id
+        )
+        WHERE ecosystem_name is null
+    """)
+
+
+def downgrade():
+    # Do nothing on downgrade - column removal will be handled by previous
+    # revision
+    pass


jeremycline:
Most unfortunately, I think this won't work as it stands. 

I played with this a bit on the latest [production database dump](https://infrastructure.fedoraproject.org/infra/db-dumps/) for release-monitoring.org, and discovered the following problem: there's a reasonably large number of duplicate (or potentially duplicate) entries which have the same (name, backend) but a different homepage. When this migration runs the uniqueness constraint on (name, ecosystem_name) kicks in and these duplicates are noticed. For example, [cpptest](https://release-monitoring.org/projects/search/?pattern=cpptest) has two entries, neither of which have ``ecosystem_name`` set.

It also looks like new projects created in the web UI don't have their ecosystem_name set, either. I think we could (and should) be stricter about this in the UI. It seems like a good opportunity to make the user experience better since we can immediately point them to the existing project if they enter the matching (name, backend-that-maps-to-an-ecosystem). People regularly accidentally make duplicates.

I think the next step here is to make sure more bad data can't get created by making sure that if the backend is the default backend for an ecosystem, the ecosystem_name gets set. We could also possibly expose that field in the web form. After that, we can clean up the data (manually, I guess? Maybe it wouldn't be too bad to make a tool to merge them automatically and present its changes for human inspection). At that point this migration should work.

ncoghlan:
Huh, I thought `create_project` *did* set the `ecosystem_name` automatically: https://github.com/fedora-infra/anitya/blob/master/anitya/lib/__init__.py#L85

I wonder if there's a problem with the Ecosystem plugin loading: https://github.com/fedora-infra/anitya/blob/master/anitya/lib/plugins.py#L72

jeremycline:
Ah, you're right. I think I did something stupid yesterday and ended up experimenting with a sourceforge package which isn't an ecosystem. We're all sorted there, then.

I wrote an awful script that found all the duplicates (I think). There were a few tricky cases where there were distribution mappings that conflicted so I felt is was easiest to clean up by hand since there were only around 30. This migration _should_ work now on release-monitoring.org.

jeremycline:
The migration applies cleanly to the release-monitoring.org database now that I've cleaned up the duplicates and the result looks sane 😄. I'll try to make a release in the next week.

Thanks!

ncoghlan:
I'm currently still getting 404's from the live site for components I know it tracks:

```
$ curl https://release-monitoring.org/api/by_ecosystem/pypi/requests/
{
  "error": "No project \"requests\" found in ecosystem \"pypi\"",
  "output": "notok"
}
```

Is that just a matter of the data migration not being deployed yet?

jeremycline:
@ncoghlan Yup, I just need to tag a release and deploy it. I was hesitant to do so this week since most people are at DevConf and I didn't want to rock the boat unnecessarily. I'm still pretty new to all this and since this upgrade includes a data migration I want someone around to hold my hand while I do it.

jeremycline:
@ncoghlan I just upgraded production to 0.11.0 and the migration has been applied.

Sorry for the very long time it took to get deployed. Next time should be much faster!

PR #404, opened 2017-01-11
Update the Debian backend to strip .orig suffix

### LINKED ISSUE #398: Debian project backend adds ".orig" to the version number
The Debian project backend should strip the ".orig" from the version numbers.

reprepro is an example of a package that is using the Debian backend: https://release-monitoring.org/project/12798/ . It currently shows the latest version number as `5.1.1.orig` rather than `5.1.1`

jeremycline:
Debian tarballs have a '.orig' suffix in their tarballs that we
previously didn't account for.

fixes #398


pypingou:
👍 for me

PR #408, opened 2017-01-16
Fix the failed migration task in Ansible

jeremycline:
I must have had a dirty alembic.ini in my home directory already in the VM that caused this to work when I added the production database dump stuff. This is necessary to migrate the database on a new VM.

### DIFF
--- a/ansible/roles/anitya-dev/tasks/db.yml
+++ b/ansible/roles/anitya-dev/tasks/db.yml
@@ -43,8 +43,14 @@
       creates: /home/vagrant/.db-imported
 
 - name: Create /home/vagrant/alembic.ini
+  become_user: "{{ ansible_env.SUDO_USER }}"
   copy: src=/home/vagrant/devel/alembic.ini dest=/home/vagrant/alembic.ini remote_src=True
 
+- name: Create /home/vagrant/alembic.ini
+  replace:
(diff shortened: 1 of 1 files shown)


ansible/roles/anitya-dev/tasks/db.yml
```diff
@@ -43,8 +43,14 @@
...
 
 - name: Create /home/vagrant/alembic.ini
+  become_user: "{{ ansible_env.SUDO_USER }}"
   copy: src=/home/vagrant/devel/alembic.ini dest=/home/vagrant/alembic.ini remote_src=True
 
+- name: Create /home/vagrant/alembic.ini
+  replace:
+    dest: /home/vagrant/alembic.ini
+    regexp: "^script_location = alembic*$"
+    replace: "script_location = devel/alembic"
```
pypingou: The copy above is done via sudo, so will the regular user have enough permission to edit the file?

jeremycline: Ah, it becomes the user who's using sudo, so in the Vagrant VM's case that's the ``vagrant`` user.

pypingou: thanks :)

pypingou:
👍 for me, thanks for the clarification

PR #414, opened 2017-01-24
implement crates.io backend

ignatenkobrain:
Closes: https://github.com/fedora-infra/anitya/issues/413
Signed-off-by: Igor Gnatenko <ignatenkobrain@fedoraproject.org>

### DIFF
--- a/anitya/lib/backends/crates.py
+++ b/anitya/lib/backends/crates.py
@@ -17,67 +17,69 @@
 # along with this program; if not, write to the Free Software
 # Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301  USA
 """
-A backend for the Rust community’s crate host, `crates.io <https://crates.io/`_.
+A backend for the Rust community’s crate host, `crates.io <https://crates.io/>`_.
 
 Crates.io provides an API to determine a package's version, so this backend
 makes use of that API rather than using regular expressions like some other
 backends. The API is very simple. An HTTP GET request to
 ``https://crates.io/api/v1/crates/{name}/versions`` returns a JSON response
 using the schema::
-    {
-      "$schema": "http://json-schema.org/draft-04/schema#",
-      "title": "crates.io versions",
-      "description": "https://crates.io/api/v1/crates/$name/versions",
-      "type": "array",
-      "definitions": {
-        "version": {
-          "type": "object",
-          "properties": {
-            "crate":      { "type": "string" },
-            "created_at": { "type": "string", "format": "date-time" },
-            "dl_path":    { "type": "string" },
-            "downloads":  { "type": "integer" },
-            "features": {
-              "type": "object",
-              # FIXME: 0+ elements
-              # str: list(str)
-            },
-            "id":         { "type": "integer" },
-            "links": {
-              "type": "object",
-              "properties": {
-                "authors":           { "type": "string" },
-                "dependencies":      { "type": "string" },
-                "version_downloads": { "type": "string" }
-              }
-              "required": ["authors", "dependencies", "version_downloads"]
-            },
-            "num":        { "type": "string" },
-            "updated_at": { "type": "string", "format": "date-time" },
-            "yanked":     { "type": "boolean" },
+
+  {
+    "$schema": "http://json-schema.org/draft-04/schema#",
+    "title": "crates.io versions",
+    "description": "https://crates.io/api/v1/crates/$name/versions",
+    "type": "array",
+    "definitions": {
+      "version": {
+        "type": "object",
+        "properties": {
+          "crate":      { "type": "string" },
+          "created_at": { "type": "string", "format": "date-time" },
+          "dl_path":    { "type": "string" },
+          "downloads":  { "type": "integer" },
+          "features": {
+            "type": "object",
+            # FIXME: 0+ elements
+            # str: list(str)
           },
-          "required": [
-            "crate",
-            "created_at",
-            "dl_path",
-            "downloads",
-            "features",
-            "id",
-            "links",
-            "num",
-            "updated_at",
-            "yanked"
+          "id":         { "type": "integer" },
+          "links": {
+            "type": "object",
+            "properties": {
+              "authors":           { "type": "string" },
+              "dependencies":      { "type": "string" },
+              "version_downloads": { "type": "string" }
+            }
+            "required": ["authors", "dependencies", "version_downloads"]
+          },
+          "num":        { "type": "string" },
+          "updated_at": { "type": "string", "format": "date-time" },
+          "yanked":     { "type": "boolean" },
+        },
+        "required": [
+          "crate",
+          "created_at",
+          "dl_path",
+          "downloads",
+          "features",
+          "id",
+          "links",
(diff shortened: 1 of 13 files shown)


anitya/lib/backends/crates.py
```diff
@@ -0,0 +1,47 @@
...
+        'https://crates.io/crates/clap',
+        'https://crates.io/crates/serde',
+    ]
+
+    @classmethod
+    def _get_versions(cls, project):
+        url = 'https://crates.io/api/v1/crates/%s/versions' % project.name
+        try:
+            req = cls.call_url(url)
+        except Exception:  # pragma: no cover
```
jeremycline: One day soon it'd probably be good to add all the error handling to the ``call_url`` function so new backends don't have to worry about this, but today is not that day, sadly.

This should catch a less broad exception. Since the URL will never be an FTP URL it'll always go through requests and you can catch an ``requests.RequestException`` for the most general requests error. It would be good to at least include the str() of the exception in the AnityaPluginException message. Something like ``'Could not contact {url}: {reason}'.format(url=url, reason=str(e))``

anitya/lib/backends/crates.py
```diff
@@ -0,0 +1,47 @@
...
+            raise AnityaPluginException('Could not contact %r' % url)
+
+        try:
+            data = req.json()
+        except Exception:  # pragma: no cover
```
jeremycline: Calling ``json()`` on a response that can't be decoded will raise a ``ValueError`` so we should explicitly catch that here rather than ``Exception``.

anitya/lib/backends/crates.py
```diff
@@ -0,0 +1,47 @@
...
+            raise AnityaPluginException('No JSON returned by %r' % url)
+
+        return data['versions']
+
+    @classmethod
+    def get_version(cls, project):
+        return self._get_versions(project)[0]['num']
```
jeremycline: Is it possible for a project to be on crates.io without (yet) having a version?

anitya/lib/backends/crates.py
```diff
@@ -0,0 +1,47 @@
...
+
+    @classmethod
+    def get_versions(cls, project):
+        return [v['num'] for v in self._get_versions(project)]
+
+    @classmethod
+    def get_ordered_versions(cls, project):
+        # crates API returns already ordered versions
+        return self.get_versions()
```
jeremycline: This call is incorrect, it needs to be passed the ``project`` object.

anitya/lib/backends/crates.py
ignatenkobrain: looks like not...

anitya/lib/backends/crates.py
```diff
@@ -0,0 +1,47 @@
...
+    ''' The crates class for projects hosted on crates.io. '''
+
+    name = 'crates.io'
+    examples = [
+        'https://crates.io/crates/clap',
+        'https://crates.io/crates/serde',
+    ]
+
+    @classmethod
+    def _get_versions(cls, project):
```
ignatenkobrain: documented (schema is a bit incomplete, but example is also there.

anitya/lib/backends/crates.py
ignatenkobrain: fixed

anitya/lib/backends/crates.py
ignatenkobrain: fixed

anitya/lib/backends/crates.py
ignatenkobrain: fixed

ignatenkobrain:
we probably want also to do something with `yanked` versions (not sure what exactly)...

```
(12:04:58 AM) ignatenkobrain: jistone: btw, what means yanked on crates?
(12:05:25 AM) jistone: it's like un-publishing a version, without making it inaccessible and breaking builds
(12:05:37 AM) jistone: so cargo won't consider that version for resolving deps
(12:06:33 AM) jistone: people usually yank because it's found broken
```

Not sure how we want to handle this from anitya's side.

jeremycline:
Hmm. Right now I don't think Anitya really has a concept of revoked releases, but it seems like that would be a nice-to-have. Some ecosystems don't (currently) provide a way to yank an update, but you can always delete git tags or remove a tarball.

I don't think we want to try and handle it here initially, but I'll file an RFE.

anitya/lib/backends/crates.py
```diff
@@ -0,0 +1,216 @@
...
+
+        Raises:
+            AnityaPluginException: If the URL was unreachable or the response
+                was in an unexpected format.
+        """
+        url = 'https://crates.io/api/v1/crates/{}/versions'.format(project.name)
+        try:
+            req = cls.call_url(url)
+            req.raise_for_status()
+            data = req.json()
```
pypingou: If for some reason the data returned is not JSON it will raise an exception that will not be caught here

anitya/lib/backends/crates.py
```diff
@@ -0,0 +1,216 @@
...
+            project (anitya.lib.model.Project): The Rust project to retrieve
+                versions for.
+
+        Returns:
+            list: A list of version strings.
+
+        Raises:
+            AnityaPluginException: If the URL was unreachable or the response
+                was in an unexpected format.
+        """
```
pypingou: I would point out that the docstring format used here is different from the docstring format used elsewhere (for example in pypi.py) and I would be in favor of being consistent on that formatting, but I'll leave this up to @jeremycline

docs/reference/anitya.lib.backends.rst
```diff
@@ -0,0 +1,12 @@
+Anitya Backends
+===============
+
+These are the backends supported by Anitya
+
+Crates.io
+---------
+
+.. automodule:: anitya.lib.backends.crates
```
pypingou: I like that idea but I think I would favor having it in a different PR so that all the modules can be added not just this one

anitya/lib/backends/crates.py
jeremycline: requests hands the data to the ``json`` module and it raises the exception. In Python 2 this will result in a ValueError (handled below) and in Python 3 it'll be a [JSONDecodeError](https://docs.python.org/3.6/library/json.html#json.JSONDecodeError) which is a subclass of ValueError so it'll also get handled below.

anitya/lib/backends/crates.py
jeremycline: Fair point. I added a section to the contribution guide about documentation and mention the style used (https://github.com/fedora-infra/anitya/pull/421). I know all the existing documentation uses the plain RST style, but I find this style much easier to read inline and I plan to convert any old strings as I go along.

They'll both render fine in HTML so it's just a matter of consistency in the code.

docs/reference/anitya.lib.backends.rst
jeremycline: The docs project is definitely a work-in-progress (there's a bunch of build warnings/errors from some docblocks). My plan is to chip away at the docs and add the modules as they're completed. The ones with docblocks don't have module-level documentation, so it makes the rendered results look a little odd.

PR #417, opened 2017-02-01
Do not install system versions of Python dependencies

### LINKED ISSUE #415: Anitya doesnt start after 'vagrant up'
After doing 'vagrant up' and 'vagrant ssh', the command to start anitya: 
`systemctl --user start anitya.service`
doesn't work.
On irc at #fedora-admin, jcline told me that It can be fixed (and yes, it does) doing :
$: workon  anitya
$: pip install -U flask
I leave as a reference the conversation i had with jcline:

> <jcline> x-ip, one thing I did find yesterday when I re-upped my vagrant environment is that flask isn't automatically upgraded to the latest version so anitya doesn't start. You'll need to run `workon anitya && pip install -U flask` once you `vagrant up` until that gets fixed in the ansible setup.

jeremycline:
Currently, we have to set up the virtualenv with the system site
packages since we have a hard dependency on the rpm module which isn't
on PyPI (as far as I can tell).

Although it is nice to have the system version of packages available
outside the virtual environment for testing, they cause problems when
added to the virtual environment. Specifically, it caused the flask
binary to not be inside the virtualenv and that is used to run the web
application. This commit removes all the non-essential system packages
to ensure the virtualenv is clean.

fixes #415

### DIFF
--- a/ansible/roles/anitya-dev/tasks/main.yml
+++ b/ansible/roles/anitya-dev/tasks/main.yml
@@ -17,19 +17,6 @@
     - openssl-devel
     - python2-devel
     - python3-devel
-    - python-bunch
-    - python-dateutil
-    - python-docutils
-    - python-flask
-    - python-flask-openid
-    - python-flask-wtf
-    - python-markupsafe
-    - python-openid
-    - python-psycopg2
-    - python-setuptools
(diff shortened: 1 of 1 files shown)


pypingou:
Wfm 👍

PR #423, opened 2017-02-07
Remove the system installation of Sphinx in Vagrant

jeremycline:
The sphinx-build binary is provided by the system installation of
Sphinx, but because we currently have to use the virtualenv with
--site-packages it causes the system installation of sphinx-build to get
called in the docs builder which fails because all the libs are in the
virtualenv.

I also noticed that the postgres headers were not installed and we're no
longer using the system postgres driver.


pypingou:
Looks fine to me

PR #424, opened 2017-02-07
Enhance the tox.ini

jeremycline:
I reviewed a nice tox PR by @abompard and so I'm putting some of the new
tricks I learned into practice. Thanks @abompard!

This adds a few new tests to the tox.ini and removes the various other
test runners. It also updates the documentation to make it clear the
tests should be run with tox.


jeremycline:
@ncoghlan I'd be very grateful if you could review this since you initially added tox support and I'm a new(ish) tox user.

ncoghlan:
Looks good to me, but the one comment I'd make is that the direct shell script remains useful for testing that Anitya runs correctly in an *existing* virtual environment.

jeremycline:
Good point. I added the shell script back.

PR #425, opened 2017-02-07
Prep for the 0.11.0 release

jeremycline:
Unfortunately to build this on EPEL 7 I had to add a patch. I plan to package Anitya for Fedora soon so we can pull all this specfile stuff out, but I want to release this first.

### DIFF
--- a/files/anitya.spec
+++ b/files/anitya.spec
@@ -2,13 +2,14 @@
 %distutils.sysconfig import get_python_lib; print (get_python_lib())")}
 
 Name:           anitya
-Version:        0.10.1
+Version:        0.11.0
 Release:        1%{?dist}
 Summary:        Monitor upstream releases and announce them on fedmsg
 
 License:        GPLv2+
 URL:            https://github.com/fedora-infra/anitya/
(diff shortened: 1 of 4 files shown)


files/anitya.spec
```diff
@@ -2,13 +2,14 @@
...
 Name:           anitya
-Version:        0.10.1
+Version:        0.11.0
 Release:        1%{?dist}
 Summary:        Monitor upstream releases and announce them on fedmsg
 
 License:        GPLv2+
 URL:            https://github.com/fedora-infra/anitya/
 Source0:        %{url}/archive/%{version}/anitya-%{version}.tar.gz
+Patch0:         0001-Remove-conditional-requires.patch
```
pypingou: It's been added but not applied.

Note that for such changes you could also just use sed in the ``%prep`` section

jeremycline: I have a terrible problem with adding patches and neglecting to apply them. Luckily, ``autosetup`` automatically applies all patches. Unluckily, I neglected to pass ``-p1`` to autosetup. I've fixed that and made sure it builds.

pypingou:
All good for me

PR #427, opened 2017-02-13
Fix all warnings generated when building the Sphinx docs

jeremycline:
The only two issues were that titles makes the HTTP API docs generator unhappy
and the indentation on one module's docblock was off.

This also enables warnings in the Sphinx project being treated as errors
and will cause the build to fail.

### DIFF
--- a/anitya/lib/exceptions.py
+++ b/anitya/lib/exceptions.py
@@ -1,12 +1,26 @@
 # -*- coding: utf-8 -*-
-
+#
+# This file is part of the Anitya project.
+# Copyright (C) 2014-2016  Red Hat, Inc.
+#
+# This program is free software; you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation; either version 2 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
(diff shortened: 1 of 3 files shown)


anitya/lib/exceptions.py
```diff
@@ -1,12 +1,26 @@
...
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program; if not, write to the Free Software
+# Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301  USA
 """
- (c) 2014-2016 - Copyright Red Hat Inc
```
pypingou: Dropping the copyright?

akien-mga: It's a few lines up in the reformatted header.

pypingou: yeah, just realized this when viewing another PR :)

pypingou:
👍 for me

anitya/lib/exceptions.py
jeremycline: Yeah, that's the way the GPL recommends applying the copyright notice

PR #401, opened 2017-01-11
Add preliminary OpenID Connect support to API

ncoghlan:
- adds OpenID Connect support for API authentication
- implements an API endpoint to request project monitoring
- requesting tokens for a live service (rather than just for testing) will be implemented in a follow-up patch

Starts implementation of #329

PR #428, opened 2017-02-13
Move tests inside the package and ignore them when packaging

jeremycline:
I had a lot of time on a plane recently so I did some reorganizing and tidying.

This places the tests inside the ``anitya`` package and ignores them
when packaging. This allows us to avoid fiddling with the system path in
the tests.

All VCR cassettes have been renamed to match the new test paths.


anitya/tests/base.py
```diff
@@ -1,6 +1,6 @@
 # -*- coding: utf-8 -*-
 #
-# Copyright © 2014  Red Hat, Inc.
+# Copyright © 2017  Red Hat, Inc.
```
pypingou: Shouldn't it be 2014-2017?

anitya/tests/lib/backends/test_gnome.py
```diff
@@ -27,13 +27,10 @@
 import sys
 import os
 
-sys.path.insert(0, os.path.join(os.path.dirname(
-    os.path.abspath(__file__)), '..'))
-
```
pypingou: sys and os are still imported here while you dropped them from other tests, intended?

anitya/tests/lib/backends/test_gnu.py
```diff
@@ -31,13 +31,10 @@
 import sys
 import os
```
pypingou: Same as above :)

anitya/tests/lib/backends/test_launchpad.py
```diff
@@ -31,13 +31,10 @@
 import sys
 import os
```
pypingou: Here as well :)

pypingou:
Couple of questions regarding some import left, otherwise looks fine

anitya/tests/base.py
jeremycline: I think this is the result of an odd diff where I made the new base.py file, then started moving stuff out of __init__.py. In the end I moved pretty much everything so git thinks (fairly, I suppose) I renamed the file and made a new __init__.py. I will adjust it.

jeremycline:
@pypingou I removed all the imports I missed.

@ncoghlan this will affect #401 (although it should be a simply git mv of those two new test modules).

pypingou:
Looks fine to me :)

PR #429, opened 2017-02-13
Stop using warnings.warn when fedmsgs fail

jeremycline:
This switches the exception block that handles fedmsg failures to use
the logger rather than the warnings module. It logs at the error level
as it's a general, unexpected exception.


pypingou:
wfm 👍

PR #430, opened 2017-02-13
Break a large unit test up into two tests

jeremycline:
This particular test was responsible for generating warnings and I
noticed it should be split up.

### DIFF
--- a/tests/test_anityalib.py
+++ b/tests/test_anityalib.py
@@ -22,7 +22,6 @@
 '''
 anitya tests for the anitya.lib module.
 '''
-
 __requires__ = ['SQLAlchemy >= 0.7']
 import pkg_resources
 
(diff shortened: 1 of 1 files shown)


tests/test_anityalib.py
```diff
@@ -102,9 +101,21 @@ def test_edit_project(self):
             user_id='noreply@fedoraproject.org')
 
         project_objs = anitya.lib.model.Project.all(self.session)
-        self.assertEqual(len(project_objs), 3)
```
pypingou: No longer checking the length returned?

jeremycline: I'm not entirely sure what prompted me to remove that, but I added it back.

pypingou:
👍 for me

PR #431, opened 2017-02-13
Fix deprecation warnings

jeremycline:
This fixes several deprecation warnings. The first two were from
importing flask extensions via ``flask.ext`` (``flask_wtf`` and
``flask_openid``). The last one was a class rename in ``flask_wtf``.


pypingou:
👍 for me

PR #433, opened 2017-02-17
Add a fedmsg consumer that handles libraries.io messages

jeremycline:
This adds a fedmsg consumer that handles messages from libraries.io by
way of the sse2fedmsg bridge. This is not complete yet, but it mostly works and
I wanted some initial feedback to see what people thing. Things left to do are:

- Tests

- A way to handle when libraries.io announces a version with a prefix (like v) and Anitya doesn't. 

- What to do when anitya finds a different version than libraries.io

- What do do when anitya doesn't find any version at all

This also sets up the Vagrant environment to use this.

### DIFF
--- a/anitya/lib/ecosystems/crates.py
+++ b/anitya/lib/ecosystems/crates.py
@@ -28,3 +28,4 @@ class CratesEcosystem(BaseEcosystem):
 
     name = 'crates.io'
     default_backend = 'crates.io'
+    aliases = ['Cargo']
--- a/anitya/tests/test_config.py
+++ b/anitya/tests/test_config.py
@@ -50,6 +50,8 @@ social_auth_login_url = "/login/"
 social_auth_login_redirect_url = "/"
 social_auth_login_error_url = "/login-error/"
 
+create_librariesio_projects = true
+
 [anitya_log_config]
     version = 1
     disable_existing_loggers = true
@@ -163,6 +165,7 @@ class LoadTests(unittest.TestCase):
             'SOCIAL_AUTH_LOGIN_URL': '/login/',
             'SOCIAL_AUTH_LOGIN_REDIRECT_URL': '/',
             'SOCIAL_AUTH_LOGIN_ERROR_URL': '/login-error/',
+            'CREATE_LIBRARIESIO_PROJECTS': True,
         }
         config = anitya_config.load()
         self.assertEqual(sorted(expected_config.keys()), sorted(config.keys()))
--- a/anitya/tests/test_librariesio_consumer.py
+++ b/anitya/tests/test_librariesio_consumer.py
@@ -1,0 +1,239 @@
+# -*- coding: utf-8 -*-
+#
+# This file is part of the Anitya project.
+# Copyright (C) 2017 Red Hat, Inc.
+#
+# This program is free software; you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation; either version 2 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program; if not, write to the Free Software
+# Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA  02110-1301  USA
+"""
+Tests for the libraries.io fedmsg consumer.
+"""
+
+from __future__ import absolute_import, unicode_literals
+
+import unittest
+
+import mock
+
+from anitya import config
+from anitya.lib.exceptions import AnityaPluginException, AnityaException
+from anitya.librariesio_consumer import LibrariesioConsumer
+from anitya.lib.model import Project
+from anitya.tests.base import DatabaseTestCase
+
+
+@mock.patch('anitya.librariesio_consumer.model.initialize', mock.Mock())
+class LibrariesioConsumerTests(DatabaseTestCase):
+
+    def setUp(self):
+        super(LibrariesioConsumerTests, self).setUp()
+
+        self.supported_fedmsg = {
+            'body': {
+                'username': 'vagrant',
+                'i': 236,
+                'timestamp': 1488561045,
+                'msg_id': '2017-630afdfd-4780-48ce-ba99-3a306d8de2d2',
+                'topic': 'org.fedoraproject.dev.sse2fedmsg.librariesio',
+                'msg': {
+                    'retry': 500,
+                    'data': {
+                        'name': 'ImageMetaTag',
+                        'project': {
+                            'status': None,
+                            'repository_url': 'https://github.com/SciTools-incubator/',
+                            'latest_release_published_at': '2017-03-03 16:44:46 UTC',
+                            'description': 'Tags images created by matplotlib with ...',
+                            'language': 'Python',
+                            'platform': 'Pypi',
+                            'package_manager_url': 'https://pypi.python.org/pypi/ImageMetaTag/',
+                            'latest_release_number': '0.1',
+                            'rank': 4,
+                            'stars': 1,
+                            'keywords': [],
+                            'normalized_licenses': ['BSD-3-Clause'],
+                            'forks': 1,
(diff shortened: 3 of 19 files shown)


anitya/lib/librariesio_consumer.py
```diff
@@ -0,0 +1,205 @@
...
+        name = librariesio_msg['name']
+        platform = librariesio_msg['platform']
+        try:
+            ecosystem = list(filter(
+                lambda eco: eco.librariesio_name == platform, ECOSYSTEM_PLUGINS.get_plugins()))[0]
+        except IndexError:
+            # TODO probably better to raise an exception here
+            return
+
+        session = Session()
```
pypingou: Should it be closed somewhere?

jeremycline: Based on my (not yet complete) understanding of SQLAlchemy, the whole point of using its ``scoped_session`` is to create a session registry where each thread gets its own session so you don't have to pass session objects around. The session "created" here is the same session "created" [here](https://github.com/release-monitoring/anitya/pull/433/files#diff-5379d281c0b6a0060c136e2c384cd573R133). When ``session.remove()`` is called the session is removed from the registry and its resources released. Next time ``Session()`` is called a new session is placed in the thread-local registry.

Not really on-topic, but we're not using a lot of nice SQLAlchemy features. I think in the near future we need to alter all the database APIs to use custom query classes (where appropriate) in combination with the [query_property](http://docs.sqlalchemy.org/en/latest/orm/contextual.html#sqlalchemy.orm.scoping.scoped_session.query_property) rather than all those static methods on the models that accept sessions. The backend should create a scoped session rather than relying on the Flask application do to so (otherwise we should be using flask-sqlalchemy since we're tying the backend to the frontend anyway).

anitya/lib/librariesio_consumer.py
```diff
@@ -0,0 +1,205 @@
...
+            Project: the Anitya project that maps to the librariesio project;
+                this may be a new project.
+            None: If there is no ecosystem that maps to the librariesio platform
+        """
+        librariesio_msg = message['body']['msg']['data']
+        name = librariesio_msg['name']
+        platform = librariesio_msg['platform']
+        try:
+            ecosystem = list(filter(
+                lambda eco: eco.librariesio_name == platform, ECOSYSTEM_PLUGINS.get_plugins()))[0]
```
ncoghlan: This would likely be clearer as a search loop (which would also save creating the list every time):

    for ecosystem in ECOSYSTEM_PLUGINS.get_plugins():
        if ecosystem.librariesio_name == platform:
            break
    else:
        # Log a debug message about an unknown ecosystem
        # (that will be pretty common - libraries.io currently tracks more ecosystems than Anitya does)
        return

anitya/lib/librariesio_consumer.py
```diff
@@ -0,0 +1,205 @@
...
+        # TODO handle things like 'v' prefix. Should be a general version normalizer
+        # used everywhere.
+        # TODO handle when we find a different version than librariesio. Create the project anyway?
+        equivalent_versions = (librariesio_version == anitya_version)
+        if not equivalent_versions:
+            _log.info('libraries.io found %r had a latest version of %s, but Anitya found %s',
+                      project, librariesio_version, anitya_version)
+        return equivalent_versions
+
+    def _fedmsg_to_project(self, message):
```
ncoghlan: I'd suggest calling this `_fedmsg_to_candidate_project_update`, as it does more than just work out which project the message relates to:

- for an unknown project, it creates a candidate project instance
- for an existing project, it speculatively updates the latest version info
- in both cases, it *avoids* committing the change, until after the legitimacy of the info can be checked

anitya/lib/librariesio_consumer.py
```diff
@@ -0,0 +1,205 @@
...
+        This serves two purposes. Firstly, we connect to libraries.io over HTTP
+        so we don't have any authentication. Secondly, we might map the
+        libraries.io project to the wrong Anitya project, so this is a good
+        opportunity to catch and log those problems.
+        """
+        librariesio_version = message['body']['msg']['data']['version']
+        try:
+            anitya_version = get_version(project)
+        except exceptions.AnityaPluginException as e:
+            # TODO what to do here? Bail? Trust libraries.io?
```
ncoghlan: Bailing out here with a logged error is correct, as the query may fail because the candidate project was complete nonsense, and instead injected via an MitM attack against the unauthenticated feed from libraries.io.

anitya/lib/ecosystems/crates.py
```diff
@@ -28,3 +28,4 @@ class CratesEcosystem(BaseEcosystem):
 
     name = 'crates.io'
     default_backend = 'crates.io'
+    librariesio_name = 'Cargo'
```
ncoghlan: Having to make all the ecosystem plugins libraries.io aware irks me a bit, but I guess the alternative is making the libraries.io handling code aware of all the ecosystems, and that runs somewhat counter to the design of the plugin model.

However, normalising with `str.lower()` would be sufficient to map all of the libraries.io names except `cargo`, and for that we could potentially change the ecosystem name on the Anitya plugin side.

Alternatively (and to better support deciding we want to change canonical ecosystem names in the future), this could be handled by way of a general purpose "aliases" field on the plugins, rather than a libraries.io specific field.

jeremycline: Yeah I'm not pleased with this either, it was just the first thing that popped into my head. Having an "aliases" field definitely sounds better to me.

anitya/lib/librariesio_consumer.py
```diff
@@ -0,0 +1,217 @@
...
+        else:
+            _log.debug('Dropped librariesio update to %s for %s (%s) since it is on the '
+                       'unsupported %s platform', version, name, homepage, platform)
+            return
+
+        session = Session()
+        project = Project.by_name_and_ecosystem(session, name, ecosystem.name)
+        if project is None:
+            # We've found a new project!
+            project = Project(
```
jeremycline: A problem with this approach that I've just found is that we're not going to be emitting the correct fedmsgs since those are sent all over the place via ``anitya.log``.

I could re-work this PR to use that approach, but I'm inclined to open a new PR that adds [ORM event listeners](http://docs.sqlalchemy.org/en/latest/orm/events.html) that publish the messages instead.

jeremycline:
Okay, I think this is ready.

ralphbean:
https://github.com/fedora-infra/sse2fedmsg

^^ cool.

anitya/tests/test_config.py
```diff
@@ -163,6 +165,7 @@ def test_full_config_file(self, mock_exists, mock_log):
             'SOCIAL_AUTH_LOGIN_URL': '/login/',
             'SOCIAL_AUTH_LOGIN_REDIRECT_URL': '/',
             'SOCIAL_AUTH_LOGIN_ERROR_URL': '/login-error/',
+            'CREATE_LIBRARIESIO_PROJECTS': True,
```
ncoghlan: Should this be `'CREATE_LIBRARIESIO_PROJECTS': create_librariesio_projects` to make sure the two test settings always stay in sync?

anitya/tests/test_librariesio_consumer.py
```diff
@@ -0,0 +1,239 @@
...
+                    },
+                    'event': 'event',
+                    'id': None
+                }
+            }
+        }
+        self.mock_hub = mock.Mock(config={'environment': 'dev', 'topic_prefix': 'anitya'})
+
+    def test_dev_environment_topics(self):
+        """Assert when the environment is set to prod, only the prod topic is used."""
```
ncoghlan: This docstring is incorrect for the dev test case.

anitya/tests/test_config.py
jeremycline: I'm not entirely sure what you mean here, could you rephrase this or show me what you mean? Thanks!

ncoghlan: Further up in this file, you have the line `create_librariesio_projects = True`. Presumably that global and this setting are supposed to be kept consistent, so the code in my comment was meant to be a literal replacement for the current dict entry.

Alternatively, if the global isn't actually used anywhere (I didn't check), it could be deleted entirely.

jeremycline: Ahhh I see. I think the diff is not very clear. That's inside a string used as the test configuration file's content.

ncoghlan: Ah, you're right - I only looked at the diff, not the full original file. In that case, cool, it makes sense as is :)

Issue #332, opened 2016-08-10
RFE: Backend based version comparison


bkabrda:
So I've been investigating why my local instance of Anitya claims that the latest version of Django is `1.10b1` when `1.10` has already been released. I found out that the underlying issue is that versions of all backends are compared using RPM version comparison. For example:
```
$ rpmdev-vercmp 1.10b1, 1.10
1.10b1, > 1.10
```
So this explains why Django `1.10b1` is rated as newer release than `1.10`. It would be cool if Anitya implemented version comparison algorithm per-backend, since every backend has some different corner cases that compare versions differently than RPM version comparison (prereleases manifest this most often).

This is a tough nut to crack, since version comparisons would need to be implemented for all backends properly. Perhaps we could start by implementing Python-specific version comparison and let all other backends fall back to the RPM based one? If this sounds good, I'll try to give it a shot and send a PR.

Also I've been using Anitya for several days right now and everything is pretty smooth, so let me thank you for writing a nice piece of software :)

pypingou:
I would much rather use one version comparison for all. The human brain is the most creative part of the universe when it comes to version  and I don't think a per backend approach would bring much except for complexity in the code base.

What you read here in `1.10b1` as the beta 1 of 1.10, some other will read it as 1.10 post release b1.
For example: https://pypi.python.org/pypi/straight.plugin/ claims the latest version is 1.4.1 while it actually us 1.4.1-post-1 (cf https://github.com/ironfroggy/straight.plugin/issues/17).

My take is that versions are a mess and I doubt we will ever be able to be flexible enough to cover all of them. One thought which I have is to try porting from rpm to the python module implementing https://www.python.org/dev/peps/pep-0440/

bkabrda:
I totally understand and agree with what you're saying in regards to using one version comparison for all (and the additional complexity).

I think the issue with the straight.plugin version is just a bug in PyPI code, which isn't all that important since pip will do the right thing and people are busy working on Warehouse rather than fixing old PyPI code.

I also agree that versions are a mess... Nevertheless, I still think this would be worth the effort, assuming we could get a library that would do this for us and we'd just call sth. like `compare_versions('pypi', '1.10', '1.10b')`. So my proposal is to keep it open, because I may actually do that (or someone from my current team, we're actually pretty interested in this :))

pypingou:
I'm fine with leaving this open for further discussion but I am not convinced currently (not saying that I couldn't be) :)

[...]

PR #448, opened 2017-03-14
Provide select-able version schemes for projects

jeremycline:
I'm _really_ sorry about the size of this patch. I didn't mean it to get this big, but I'm not sure anything can be broken out of the commit.

This patch provides select-able versions for projects by using
SQLAlchemy's single table inheritance feature. Each version type is
implemented as a sub-class of the ``ProjectVersion`` model and
SQLAlchemy creates the correct Python class via the new ``type`` column
in the ``ProjectVersion`` table. I have implemented two version
comparison schemes:

* PEP-440 via the [packaging](https://github.com/pypa/packaging) library.
  This version is very flexible since it has a LegacyVersion class that
  parses everything. It should probably be the default version for
  projects.

* Semantic version scheme via
  [semanticversion](http://pythonhosted.org/semantic_version/).

It would be good to also have a calendar version scheme, but this PR is
already much larger than I wanted.

Many of the tests had to be adjusted since the way versions are sorted
has been altered.

### DIFF
--- a/alembic/versions/8075a90bbac0_add_a_version_scheme_column_to_projects.py
+++ b/alembic/versions/8075a90bbac0_add_a_version_scheme_column_to_projects.py
@@ -1,0 +1,29 @@
+"""
+Add a version scheme column to Projects
+
+Revision ID: 8075a90bbac0
+Revises: 9c29da0af3af
+Create Date: 2017-03-09 21:32:35.672952
+"""
+from alembic import op
+import sqlalchemy as sa
+
+# revision identifiers, used by Alembic.
+revision = '8075a90bbac0'
+down_revision = '9c29da0af3af'
+
+
+def upgrade():
+    op.add_column('projects_versions', sa.Column('type', sa.String(length=50),
+                  nullable=False, server_default='PEP-440'))
+    # Once we initialize all the rows with a default, we can drop the server
+    # default. We want it to be set automatically by a pre-flush hook in Anitya
+    # or explicitly by outside users.
+    op.alter_column('projects_versions', 'type', server_default=None)
+    op.add_column('projects', sa.Column('version_scheme', sa.String(length=50),
+                  nullable=False, server_default='PEP-440'))
+
+
+def downgrade():
+    op.drop_column('projects_versions', 'type')
+    op.drop_column('projects', 'version_scheme')
--- a/anitya/__init__.py
+++ b/anitya/__init__.py
@@ -4,14 +4,11 @@ import logging
 
 import anitya.lib.plugins
 import anitya.lib.exceptions
-# Feature check: version comparisons were historically done with cmp
-# For Python 3, we need to switch to using a key function instead
-try:
-    cmp
-except NameError:
-    from functools import cmp_to_key as _convert_cmp_to_key
-else:
-    _convert_cmp_to_key = None
+# Import the events to ensure the event handlers are registered with SQLAlchemy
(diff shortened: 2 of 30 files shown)


kentfredric:
Downsides of "per-project" based schemes is that for things that are published on CPAN, for example, you create a lot of work because literally /everything/ on CPAN uses the Perl version scheme. 

Which is neither semantic, nor PEP* anything.

I'd hope at least that:

1. Every backend has a *default* version scheme when a project is added.
2. That when this change is rolled out, a one-time assignment of all affected packages on a given backend be set to the default scheme.

But the upside of per-project is you can use schemes without needing them be on a given backend.

However, you're still going to need to implement a perl version scheme, and its a [substantive amount of work to get right](https://github.com/release-monitoring/anitya/issues/332#issuecomment-245027173), and there's no python library I'm aware of that handles it.  But as long as that arrives in a close followup PR , then :+1: 

I even have a basic implementation of something I threw together for something else that you can possibly steal some logic from, but my python is gross so it will probably make your eyes bleed. 

https://gist.github.com/kentfredric/cfd5a593d3d90d87ae11

ncoghlan:
Huh, the Perl version comparisons are even more complicated than I thought - setuptools was heavily inspired by CPAN, so I thought `packaging.LegacyVersion` was a reasonably close approximation to `version.pm`. I guess not :)

For the design itself, I think that having just gotten the Backend and Ecosystem definitions out of the database model, it would be a step backwards to have to inject the version schemes directly into it. 

Instead, I'd propose the following implementation model:

1. Sorting by version always takes place on the client, never in the database (already part of the patch)
2. Only the raw upstream version string is ever stored in the database (already part of the patch)
3. Like backends and ecosystems, version schemes are managed as Python plugins
4. Projects get a null-by-default version scheme field in the database
5. We do a lazy cached lookup of the version scheme field based on:
    * project scheme if defined
    * default ecosystem scheme if defined
    * default backend scheme if defined
    * otherwise the PEP 440 scheme (which isn't entirely universal, but is reasonably comprehensive and has well-defined behaviour for everything it covers)
6. The version scheme plugins would provide the conversion from the raw strings to ordered runtime objects that is provided by `ProjectVersion` in the current patch

I believe that approach would also be more amenable to being broken up into at least two steps - the first step could be to move the existing RPM-style comparison out to a plugin, and hardcode the scheme lookup to always return that plugin (i.e. a pure refactoring with no functional changes), and then adding more plugins and making things configurable would happen in a subsequent patch.

kentfredric:
Yeah. That approach seems sensible.

Upstream's advice for downstream has long been "make sure you normalise perl version schemes into a consistent scheme that makes sense for you before you do anything with it, and so that approach basically does that.

Also means if somebody accidentally implements a bug in a normalizer, you can just patch it and then nuke the cache and things will JustWork™

jeremycline:
> For the design itself, I think that having just gotten the Backend and Ecosystem definitions out of the database model, it would be a step backwards to have to inject the version schemes directly into it. (Note: I think the subclassing trick is cool, I just don't think it's a good idea to tightly couple the known version schemes to the core data model)

Yeah, that's fair. One use case I had in mind was that sometimes projects change their version scheme so it'd be good to know what type a version is so we can display and sort old versions, _but_ that's a serious edge case and there's no way to edit the type a version is in the UI anyway. I think the implementation you proposed is good (and has the advantage of being easily expanded to the database later if we learn we need that).

PR #450, opened 2017-03-15
Decouple the Anitya config from Flask

jeremycline:
This commit changes how the configuration for Anitya is loaded. The
motivation for this is to get rid of the troubling circular imports
caused by the backend using a scoped session that's created by the
front-end. This is done because the configuration is loaded by Flask.

Now the configuration is loaded in the ``anitya.config`` module. This
module has no dependencies on other pieces of Anitya. This will let the
backend set up the scoped session. The Flask frontend, which (obviously)
already depends on the backend can then make use of that scoped session.

Note: I changed the configuration format to YAML (because I like YAML),
but if people are opposed to this it can be easily switched back to a
Flask-compatible Python file. I _also_ removed the ANITYA_LOG_LEVEL
setting in favor of using a dict config.


ncoghlan:
What would your feelings be on using TOML and pytoml over YAML?

I'm not *opposed* to YAML, it's just way more complicated than is typically needed for a static config file, while TOML is about as expressive as JSON, while being far friendlier to human editing.

ncoghlan:
Big +1 for the general idea of the change though (both decoupling the backend from the front end, and moving more of the config out of Python and into a static config file)

The main thing I looked for when reviewing is that the YAML is being loaded with `safe_load` (which it is), but that's also why I asked about potentially using TOML instead :)

pypingou:
The item for which I'd have the most concern is ``PERMANENT_SESSION_LIFETIME`` since I doubt we can specify a timedelta object in a static yaml (or toml) file, can we? Though maybe it would accept a basic integer, I didn't check

jeremycline:
> What would your feelings be on using TOML and pytoml over YAML?

I don't have strong feelings one way or the other. The format does look nice. The only thing that makes me hesitate is it's not officially stable yet, but since it hasn't changed in over a year maybe that's stable enough. If you and @pypingou are in favor of TOML I'm happy to switch.

> The item for which I'd have the most concern is PERMANENT_SESSION_LIFETIME since I doubt we can specify a timedelta object in a static yaml (or toml) file, can we?

I think you theoretically _could_ in YAML, but I opted for making it a value in seconds which is turned into a timedelta by the configuration loader.

ncoghlan:
The version of TOML that gets used in practice (including by Rust's cargo config files and Python's new `pyproject.toml` file) is specifically TOML 0.4, rather than the yet-to-be-defined 1.0. There's a bit more about how TOML 0.4 came to be chosen as the preferred format in the `pyproject.toml` PEP:

* https://www.python.org/dev/peps/pep-0518/#specification
* https://www.python.org/dev/peps/pep-0518/#other-file-formats

jeremycline:
I'm convinced, I've switched it to TOML.

pypingou:
Note about toml, we'll need to ask for an epel7 branch :)

jeremycline:
Okay, pytoml is in being pushed to epel7-testing as we speak so we should be set on that front. It seems like everyone is satisfied with this change. Any final comments before I merge it?

Issue #445, opened 2017-03-07
check_feed for maven backend


pkajaba:
Hello, 

I would like to implement `check_feed` method for maven backend. From my investigation maven does central does not have any endpoint where you can query for new packages.

However there is maven index https://repo.maven.apache.org/maven2/.index/ and tool to manipulate maven index: https://github.com/apache/maven-indexer 

It should be doable this way, however anitya would have to call java program during check_feed for maven. What I want ask is - is something like upstreamable?

jeremycline:
Personally, I think that's upstream-able. I'm sure @ncoghlan and @pypingou have opinions as well, but if they don't object then I'd say go for it.

ncoghlan:
I don't see any problems with adding java dependencies to the environment where anitya-cron runs, so there just needs to be a mechanism for making sure the right dependencies are installed.

pypingou:
Any idea on how the two program would interact? How the java piece would get started and report its results?

pkajaba:
I have almost completed java application can read maven index. 

@pypingou my idea about cooperation is that `check_feed` of maven backend will call java application as sub-command are java application will outputs JSON of recently changed jars inside maven index.
Cron would invoke `check_feed` for maven backed every week (changes of maven index in maven central are published every week) and anitya would store changes inside registry. 
I hope it makes sense.

However there is one thing with which I am struggling to understand. For `pypi` [1] or `npmjs` [2] backend anitya picks just first 40 recent changes. why is that so? 

[1] https://github.com/release-monitoring/anitya/blob/master/anitya/lib/backends/pypi.py#L84
[2] https://github.com/release-monitoring/anitya/blob/master/anitya/lib/backends/npmjs.py#L115

Zlopez:
I removed the check_feed method from cron job #754, because it was not used for anything. So I'm closing this issue.

PR #449, opened 2017-03-14
Update tox.ini and setup Travis CI

jeremycline:
Note: This depends on #448 getting merged

This updates the tox.ini file and creates a ``.travis.yml`` file for CI builds. [Here's a passing build from a branch based on #448](https://travis-ci.org/jeremycline/anitya/builds/211108058).

One thing I'd like to propose as part of this is dropping Python 2.6 support. I wouldn't be opposed to dropping 2.7 support in the near future, but release-monitoring.org is still using 2.7 so I'm not quite ready for that, yet.

PR #452, opened 2017-03-16
Addition of support for check_feed method for Maven backend

pkajaba:
Hello, 

I am trying to add support for `check_feed` for maven backend. Initial discussion is here #445.

This support is dependent on [this small application](https://github.com/pkajaba/maven-index-checker) which downloads, unpacks and reads mavne index.

In terms of functionality this should work. Only problem what I have is how to make that java application as dependency properly.


pkajaba:
I see that CI is failing, but I don't think it's my fault (other than too long line). Am I right?

jeremycline:
Correct. #449, which introduced the Travis configuration, depends on #448 to pass. Once #448 gets merged you can rebase and the tests should pass.

pkajaba:
OK, thanks. And what do you think about overall approach? Would you kindly do me a review?

anitya/lib/model.py
```diff
@@ -439,6 +439,19 @@ def by_name_and_homepage(cls, session, name, homepage):
...
 
     @classmethod
+    def by_version_url(cls, session, version_url):
+        try:
+            query = session.query(
+                cls
+            ).filter(
+                cls.version_url == version_url
+            )
+            return query.one()
```
pypingou: There will be many cases where this will return more than one result, in which case, there is in theory no guarantee about which project you are retrieving

anitya/lib/backends/maven.py
```diff
@@ -70,3 +75,34 @@ def get_versions(cls, project):
...
+        try:
+            data = subprocess.check_output(["java", "-jar", jar_path], universal_newlines=True)
+        except Exception:
+            raise AnityaPluginException('Could not start %s' % maven_url)
+
+        data = json.loads(data)
+        for item in data[:40]:
+            item = json.loads(item)
+            version_url = '{group_id}:{artifact_id}'.format(group_id=item['groupId'], artifact_id=item['artifactId'])
+            retrieved_project = Project.by_version_url(session, version_url)
```
pypingou: I'd encourage we find something a little more specific that this, version_url isn't and isn't meant to be unique

anitya/lib/model.py
pkajaba: I agree that there will be more then one result in this. However I created this function because I need to query database whether there is `name` defined for `version_url`. When I am going to add new maven project, I have no information about `name`. All I have is `version_url` which consists of `groupId:artifactId`. So I will try whether there is package with such `version_url` and if I use that `name`.

I know that this query returns more results, but they should vary just in version. I will update it to return all matching projects and I will get just one from which I obtain name.

anitya/lib/backends/maven.py
pkajaba: from previous maven backend code `version_url` is name for maven coordinates. https://github.com/release-monitoring/anitya/blob/master/anitya/lib/backends/maven.py#L64

Do you think that `maven_coordinates` would be better name?

pkajaba:
@pypingou I was thinking how to make this like optional dependency and I can make rpm and build in copr. What do you think?

anitya/lib/backends/maven.py
pypingou: Yeah maybe we can change the variable name and put a comment that for maven backend we store the coordinate in the version_url field

pkajaba:
@jeremycline ping

anitya/lib/backends/maven.py
```diff
@@ -70,3 +75,37 @@ def get_versions(cls, project):
...
+    @classmethod
+    def check_feed(cls):
+        ''' Return a generator over the latest 40 packages to Maven central index
+
+        by calling maven-index-checker application
+        '''
+        maven_url = 'http://repo2.maven.org/maven2'
+        jar_path = '/src/maven-release-checker-1.0-SNAPSHOT-jar-with-dependencies.jar'
+        # session for checking whether there is already created package
+        session = init(APP.config['DB_URL'], create=True, debug=False)
```
jeremycline: ``create=True`` will cause SQLAlchemy to emit create statements which I don't think we really want to do here.

FYI when https://github.com/release-monitoring/anitya/pull/466/commits/d8c789b3f512c01560518f6a89798885a86fb202 is merged you can just
```
from anitya.lib.model import Session
session = Session()
session.query.....
Session.remove()
```

In fact, if all you're doing is querying the projects, you'll be fine with
```
from anitya.lib.model import Project
try:
    projects = Project.query.filter(Project.version_url == version_url).all()
except NoResultFound:
    projects = None
```

pkajaba:
Thanks for review! I have edited everything as you wanted. Only problem is with tests. Since there is java application I can just prerecord http as it is done with other backends. 

And If I don't have prerecorded results I can't create any meaningful tests around that.

jeremycline:
> I can edit java application and it will serve json through some localhost endpoint.

I would be fine with just a unit test that mocked out the subprocess call and returned the JSON you expect from the application. Depending on the size you could just dump it into the test. If you're not sure how to do that I'm happy to write the tests, it'll just be a few days (probably not this week or early next week).

> how to bring that java thing as dependency?

Maybe the jar path could be a configuration value (empty by default) that this backend checks for. If it's defined it tried to use it, otherwise emits a warning that the functionality requires that jar to be configured.

Then packagers can optionally require that and, if they do, provide an altered base configuration to point to the jar.

pkajaba:
Hello, I have added tests. Unfortunately there are no tables in database during tests.

anitya/lib/backends/maven.py
```diff
@@ -70,3 +74,50 @@ def get_versions(cls, project):
...
+            raise AnityaPluginException('no maven-release-checker jar file specified')
+
+        try:
+            data = check_output([APP.config['JAVA_PATH'], "-jar",
+                                 APP.config['JAR_NAME']], universal_newlines=True)
+        except CalledProcessError:
+            raise AnityaPluginException('maven-release-checker exited with non zero value')
+
+        data = json.loads(data)
+        session = init(APP.config['DB_URL'])
```
jeremycline: Sorry I missed this earlier - this is why the database tables aren't present in your tests. This function creates a new SQLAlchemy engine and session factory using the app's configured database, but the tests use an in-memory SQLite database it sets up for you. 

I've [got a PR open](https://github.com/release-monitoring/anitya/pull/466) to make this whole pattern much simpler. When that's merged, all you need to do is:
```
Project.query.filter(Project.version_url == maven_coordinates).all()
```
The query property takes care of creating a query object from the scoped session automatically and you don't have to worry about it.

I'll get that PR passing CI and badger someone to review it ASAP, then you can rebase, modify your query here, and hopefully the tests should start passing!

pkajaba: No problem, So I have to just wait for now? Or do you request any other changes? 

I will change the database connection as soon as you merge PR you mentioned

jeremycline: Nothing else jumped out at me, so waiting for now is good. I _will_ get to it this week, I promise :smile:. Stuff came up last week that I still haven't quite finished dealing with. Sorry about that!

pkajaba:
I have realized after some testing that URL are not always created same way. 

for example this: `http://repo2.maven.org/maven2/com/liferay/com.liferay.amazon.rankings.web/` URL has dots inside of it's name.

another example is `http://repo2.maven.org/maven2/io/get-coursier/coursier_sjs0.6_2.11/` where normal replacing dots for '/' leads to broken link.

I have tried to address this issue. I will add tests too for new methods

anitya/lib/backends/maven.py
```diff
@@ -70,3 +75,96 @@ def get_versions(cls, project):
...
 
         return get_versions_by_regex(url, REGEX, project)
+
+    @classmethod
+    def verify_url(cls, group_id, artifact_id):
+        match_decimal = '(?=\d)'
+        match_non_decimal = r'(?=\D)'
+
+        # match name like like artifact_id23.23_1
+        temp_artifact = do_replace(artifact_id, match_decimal, match_decimal)
```
jeremycline: Can you explain exactly what you want this regex to do? I _think_ you can rewrite this entire block with:
```
temp_artifact = artifact_id
temp_group = group_id
for pattern1, pattern2 in itertools.product(['(?=\d)', r'(?=\D)'], repeat=2):
    temp_artifact = re.sub(r'{}\.{}'.format(pattern1, pattern2), r'/', temp_artifact)
    temp_group = re.sub(r'{}\.{}'.format(pattern1, pattern2), r'/', temp_group)
```
But at first glance this seems to (eventually) match any ``.`` character since it'll match ``([0-9] | [a-zA-Z])\.([0-9] | [a-zA-Z])`` right? I'm generally a big fan of prefacing any regex with a comment about what I think it should be doing, since it's so easy to write confusing regex.

anitya/lib/backends/maven.py
```diff
@@ -70,3 +75,96 @@ def get_versions(cls, project):
...
+        temp_artifact = do_replace(temp_artifact, match_non_decimal, match_decimal)
+        # just replace dots for slashes
+        temp_artifact = do_replace(temp_artifact, match_non_decimal, match_non_decimal)
+
+        # do same thing with group id
+        temp_group = do_replace(group_id, match_decimal, match_non_decimal)
+        temp_group = do_replace(temp_group, match_non_decimal, match_decimal)
+        temp_group = do_replace(temp_group, match_non_decimal, match_non_decimal)
+
+        homepage = do_request(temp_group, temp_artifact)
```
jeremycline: This whole set of if blocks could be re-written as
```
for group, artifact in [(temp_group, temp_artifact), (temp_group, artifact_id),
                                (group_id, temp_artifact), (group_id, artifact_id)]:
            homepage = '{maven_url}/{group}/{artifact}/'.format(
                maven_url=MavenBackend.maven_url, group=group, artifact=artifact)
            try:
                request = requests.get(homepage)
                if request.status_code == 200:
                    return homepage
            except requests.RequestException:
                pass
```

anitya/lib/backends/maven.py
```diff
@@ -70,3 +75,96 @@ def get_versions(cls, project):
...
+            yield name, homepage, cls.name, version
+
+
+def do_replace(input_string, first_pattern, second_pattern):
+    return re.sub(r'{first_pattern}\.{second_pattern}'.
+                  format(first_pattern=first_pattern,
+                         second_pattern=second_pattern), r'/', input_string)
+
+
+def do_request(group_id, artifact_id):
```
jeremycline: I think you can rewrite the code that uses this function to only need to call this once, so this function can go away as well.

anitya/lib/backends/maven.py
```diff
@@ -70,3 +75,96 @@ def get_versions(cls, project):
                           .replace(':', '/'))
 
         return get_versions_by_regex(url, REGEX, project)
+
+    @classmethod
+    def verify_url(cls, group_id, artifact_id):
```
jeremycline: This needs a documentation block in the ["google"-style](http://www.sphinx-doc.org/en/1.5.5/ext/example_google.html).

anitya/lib/backends/maven.py
pkajaba: `r'(?=\d)' and `r'(?=\D)'` are both non capturing regexps. They should match decimal/non decimal characters with capturing them. What I am trying to achieve is replace `dots` for `slashes` in `artifactId` and `groupId`. In theory `replace('.', '/')` should be enough. However it's harder, because URL can contain numbers with `dots` [1],
Another thing is that some URL's don't have slashes instead of dots [2].
So I tried to create approach which tests all combinations.

[1] http://repo2.maven.org/maven2/com/zenecture/neuroflow-application_2.12/
[2] http://repo2.maven.org/maven2/com/liferay/com.liferay.amazon.rankings.web/

anitya/lib/backends/maven.py
pkajaba: Thanks way more elegant!

anitya/lib/backends/maven.py
```diff
@@ -70,3 +75,96 @@ def get_versions(cls, project):
...
+            else:
+                name = maven_coordinates
+                homepage = cls.verify_url(item['groupId'], item['artifactId'])
+                if homepage is None:
+                    continue
+            version = item['version']
+            yield name, homepage, cls.name, version
+
+
+def do_replace(input_string, first_pattern, second_pattern):
```
pkajaba: changed

anitya/lib/backends/maven.py
pkajaba: changed

anitya/lib/backends/maven.py
```diff
@@ -9,10 +9,14 @@
...
 import re
+from subprocess import check_output, CalledProcessError
+import json
+import requests
 
 from anitya.lib.backends import BaseBackend, get_versions_by_regex
 from anitya.lib.exceptions import AnityaPluginException
-
+from anitya.lib.model import Project, NoResultFound
+from anitya.app import APP
```
pkajaba: no problem, changed it

jeremycline:
Thanks for updating everything! I'll carve out some time in the next few days to give this another thorough look-over.

If I want to play with this locally, what packages would I need to install? It would probably be good to add a little guide to the documentation about this feature.

pkajaba:
I will update README. All you need is java, maven and build this package https://github.com/pkajaba/maven-index-checker

pkajaba:
I have added instructions. I have a suggestion for a future that maybe we can move `maven-index-checker` from my page to this organization.

jeremycline:
I played around with this a little bit, but something seems odd. It's loading 10,000 items from the JSON data, but the documentation indicates it'll yield the latest 40 packages.

pkajaba:
Yes, I have added new functionality meanwhile. I have to update this to ouput just 40 packages. Or is it ok if it outputs more (given I update documentation)

jeremycline:
I think 10,000 is too many, especially given it results in ~40,000 HTTP requests. It looks like there's a scheme to the dots vs. slashes, see #487. Can we do that here, as well?

PR #458, opened 2017-03-22
When connection is created read enviroment variable for database url

pkajaba:
This adds option to specify `db_url` from environment variable.

### DIFF
--- a/anitya/default_config.py
+++ b/anitya/default_config.py
@@ -10,6 +10,7 @@ anitya default configuration.
 """
 
 from datetime import timedelta
+import os
 
 # Set the time after which the session expires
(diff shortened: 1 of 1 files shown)


pypingou:
I think this should be covered by using the environment variable to override the configuration for some specific keys rather than using this mechanism to override a specific key (leading to multiple variables if you want to override multiple keys).

Create a file (say ``config``), put the DB_URL you want to use in it and point to it with the variable ``ANITYA_WEB_CONFIG``. It should allow you to modify any of the existing configuration keys without having to change default_config.py or using this mechanism.

pkajaba:
Thanks, I believe I can close this after knowing this.

PR #459, opened 2017-03-22
Remove the ecosystems and backends tables

jeremycline:
These tables are in the database for historical reasons, but they don't
need to remain there. This removes the tables with a migration, removes
the models, and adjusts the code that used the models previously.

This came up in #332 since I found myself writing a migration involving ecosystems and wondering what the point was.

### DIFF
--- a/anitya/lib/model.py
+++ b/anitya/lib/model.py
@@ -17,8 +17,8 @@ __requires__ = ['SQLAlchemy >= 0.7']  # NOQA
 import pkg_resources  # NOQA
 
 import sqlalchemy as sa
+from sqlalchemy.orm import validates
 from sqlalchemy.ext.declarative import declarative_base
-from sqlalchemy.orm import relationship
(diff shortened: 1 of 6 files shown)


anitya/lib/model.py
```diff
@@ -369,6 +302,26 @@ class Project(BASE):
                             name="UNIQ_PROJECT_NAME_PER_ECOSYSTEM"),
     )
 
+    @validates('backend')
+    def validate_backend(self, key, value):
```
ncoghlan: We may have to revisit this logic later if we ever remove any plugins (such that a backend or ecosystem is referenced in the DB without a runtime plugin available), but it seems fine as long as we're only adding new plugins, rather than removing or renaming existing ones.

ncoghlan:
The Travis failures look to be solely related to the missing RPM bindings, so I went ahead and merged this.

Issue #460, opened 2017-03-23
Make the tests pass without system RPM bindings


ncoghlan:
The tests are currently consistently failing in Travis due to `import rpm` failing in `backends.rpm_cmp`. The same problem can be seen when testing with `tox` - even if you have other Python runtimes available (as on Fedora), you can still only test in the system Python installations, as those are the only ones where `import rpm` will work.

I see three possible ways forward for this:

1. Just skip the affected tests when `import rpm` fails
2. Replicate the `rpm.labelCompare` logic in pure Python (as described in http://stackoverflow.com/a/3206477/597742 ) and fall back to that emulation when `rpm.labelCompare` isn't available.
3. Work with the RPM project to make the RPM bindings pip-installable (although this probably still won't work on non-RPM distros)

While option 3 would fix the tox-on-Fedora problems, it wouldn't help with Travis, so I think we can rule that out.

Option 1 has the significant downside that it would end up skipping quite a few tests in the pre-merge testing.

That leaves option 2, and if it's just a fallback for testing purposes, we don't even need to handle arbitrary label comparisons, just the ones we actually use in the test suite.

ncoghlan:
#461 implements option 2, and that does indeed let the tests pass in Travis.

The fallback attempts to emulate the full algorithm, and does that at least well enough to let Anitya's own tests pass.

PR #461, opened 2017-03-23
 Issue #460: Handling rpm library being missing

ncoghlan:
RPM bindings are not available when running in an
isolated virtual environment, in a non-system
Python installation, or on a non-RPM distro.

This adds a fallback implementation of
`rpm.labelCompare` for use when the
system library is unavailable.


ncoghlan:
Huh, I hadn't noticed that the way RPM defines label comparisons means it doesn't actually matter whether or not you split the epoch and release out from the version first - you'll get the same answer regardless :)

PR #462, opened 2017-03-23
Stop using pkg_resources

jeremycline:
This drops all the various pkg_resources statements. Firstly, we no
longer support Python 2.6 and the statements were to support installing
Anitya on EL6 which no longer works since we don't support 2.6.

Secondly, this is a downstream packaging concern that I would prefer to
carry as a patch. It causes all sorts of problems. For example, having
it in the setup.py means you must have SQLAlchemy installed before you
can install Anitya which is annoying.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,9 +2,6 @@ sudo: false
 language: python
 install:
   - pip install -U setuptools pip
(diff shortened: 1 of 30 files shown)


pypingou:
👍

Issue #441, opened 2017-02-28
v2 REST API emits 500 errors if project registration fails


ncoghlan:
Follow-up to PR #401:

The initial version of the project monitoring API emits 500 errors if the supplied arguments pass validation, but project registration fails anyway (e.g. when attempting to register an already registered project)

ncoghlan:
Related problem: the API currently returns 200 instead of 201 when registration *succeeds*. The docs merged in #465 describe the correct status codes for the endpoint.

PR #465, opened 2017-03-24
Add the v2 HTTP API to the documentation.

jeremycline:

### DIFF
--- a/anitya/api_v2.py
+++ b/anitya/api_v2.py
@@ -34,7 +34,65 @@ class ProjectsResource(Resource):
 
     @anitya.authentication.require_api_token("upstream")
     def post(self):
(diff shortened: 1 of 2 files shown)


jeremycline:
Just a note, the second commit adds documentation for what the API _should_ be, not what it actually is at the moment. For example, the current API returns 200 rather than 201, and crashes rather than returning most 4xx codes.

ncoghlan:
Ah, the power of documentation to find all the error handling code you forgot to write :)

ncoghlan:
Turns out I had already filed https://github.com/release-monitoring/anitya/issues/441 to cover the TODO about the missing 4xx errors, so I just added a note there to fix the successful creation response as well.

PR #466, opened 2017-03-24
Add pagination to the GET api/v2/projects/ endpoint

jeremycline:
This contains 3 commits, which should all be atomic and self-contained, so you can review them separately (oldest to newest, obviously).

It moves the scoped session creation to the model module, then adds a query property to models with a custom Query class that provides pagination for all models. Finally, it uses this in the new REST API.

### DIFF
--- a/anitya/api_v2.py
+++ b/anitya/api_v2.py
@@ -20,6 +20,50 @@ _BASE_ARG_PARSER = reqparse.RequestParser(trim=True, bundle_errors=True)
 _BASE_ARG_PARSER.add_argument('access_token', type=str)
 
 
+def _page_validator(arg):
+    """
+    Validator for a pagination page number.
+
+    Args:
+        arg (object): The object to validate as an integer greater than or
+            equal to 1.
+
+    Returns:
+        int: The validated argument.
+
+    Raises:
+        ValueError: If the integer is smaller than 1 or it can't be cast to an
+            int.
+    """
+    arg = int(arg)
+    if arg < 1:
+        raise ValueError(_('Value must be greater than or equal to 1.'))
+    return arg
+
+
+def _items_per_page_validator(arg):
+    """
+    Validator for a pagination items_per_page number.
+
+    Args:
+        arg (object): The object to validate as an integer greater than or
+            equal to 1 and less or equal to 250.
+
+    Returns:
+        int: The validated argument.
+
+    Raises:
+        ValueError: If the integer is smaller than 1 or it can't be cast to an
+            int.
+    """
+    arg = int(arg)
+    if arg < 1:
+        raise ValueError(_('Value must be greater than or equal to 1.'))
+    if arg > 250:
+        raise ValueError(_('Value must be less than or equal to 250.'))
+    return arg
+
+
 class ProjectsResource(Resource):
     """
     The ``api/v2/projects/`` API endpoint.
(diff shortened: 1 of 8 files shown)


anitya/api_v2.py
```diff
@@ -18,19 +18,100 @@
 _BASE_ARG_PARSER.add_argument('access_token', type=str)
 
 
+def _page_validator(arg):
+    """Validator for a pagination page number."""
+    arg = int(arg)
```
pypingou: What is int() raises an exception, do we want to handle that here?

jeremycline: No, the flask-restful middleware will catch the exception and turn it into a pretty HTTP 400 with details about what went wrong. I did neglect to add tests for that though (which is why codecov was displeased with me). You can check out the tests to see what happens when the cast to int fails :)

anitya/api_v2.py
```diff
@@ -18,19 +18,100 @@
 _BASE_ARG_PARSER.add_argument('access_token', type=str)
 
 
+def _page_validator(arg):
+    """Validator for a pagination page number."""
```
bowlofeggs: It would be good to docblock the ```arg``` arg.

anitya/api_v2.py
```diff
@@ -18,19 +18,100 @@
...
+    arg = int(arg)
+    if arg < 1:
+        raise ValueError(_('Value must be greater than 1'))
```
bowlofeggs: Technically, greater than or equal to judging by the code.

anitya/api_v2.py
```diff
@@ -18,19 +18,100 @@
...
+    return arg
+
+
+def _items_per_page_validator(arg):
+    """Validator for a pagination items_per_page number."""
```
bowlofeggs: It would be good to docblock the ```arg``` arg.

anitya/api_v2.py
```diff
@@ -18,19 +18,100 @@
...
+    arg = int(arg)
+    if arg < 1:
+        raise ValueError(_('Value must be greater than 1'))
```
bowlofeggs: greater than or equal to

anitya/api_v2.py
```diff
@@ -18,19 +18,100 @@
...
+    if arg > 250:
+        raise ValueError(_('Value must be less than 250'))
```
bowlofeggs: Less than or equal to.

anitya/api_v2.py
```diff
@@ -18,19 +18,100 @@
...
+            }
+
+
+        :query int page: The project page number to retrieve (defaults to 1).
+        :query int items_per_page: The number of items per page (defaults to
+                                   25, maximum of 250).
+        :statuscode 200: If all arguments are valid. Note that even if there
+                         are no projects, this will return 200.
+        :statuscode 400: If one or more of the query arguments is invalid.
+        """
```
bowlofeggs: Very nice docblock!

anitya/config.py
```diff
@@ -21,6 +21,7 @@
 
 from datetime import timedelta
 import logging
+import logging.config
```
bowlofeggs: This change seems unrelated to the commit. Is it needed? I recommend making a separate commit with it if it is.

anitya/lib/model.py
```diff
@@ -70,7 +71,88 @@ def initialize(config):
...
+    A sub-class of namedtuple that represents a page.
+
+    Attributes:
+        items (object): The database objects from the query.
+        page (int): The page number used for the query.
+        items_per_page (int): The number of items per page.
+        total_items (int): The total number of items in the database.
+    """
+
+    def as_dict(self):
```
bowlofeggs: It would be good to docblock this, even though it's fairly obvious.

PR #467, opened 2017-03-26
Update the projects v2 API to match the documented behaviour

### LINKED ISSUE #441: v2 REST API emits 500 errors if project registration fails

jeremycline:
fixes #441

### DIFF
--- a/anitya/lib/model.py
+++ b/anitya/lib/model.py
@@ -350,8 +350,8 @@ class Project(BASE):
             version_url=self.version_url,
             version=self.latest_version,
             versions=self.versions,
-            created_on=time.mktime(self.created_on.timetuple()),
-            updated_on=time.mktime(self.updated_on.timetuple()),
+            created_on=time.mktime(self.created_on.timetuple()) if self.created_on else None,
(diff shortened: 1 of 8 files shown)


anitya/lib/model.py
```diff
@@ -353,8 +353,8 @@ def __json__(self, detailed=False):
             version_url=self.version_url,
             version=self.latest_version,
             versions=self.versions,
-            created_on=time.mktime(self.created_on.timetuple()),
-            updated_on=time.mktime(self.updated_on.timetuple()),
+            created_on=time.mktime(self.created_on.timetuple()) if self.created_on else None,
```
pypingou: When would you have a JSON representation of a project without a creation date?

jeremycline: The ``created_on`` and ``updated_on`` fields get set when the object is being committed to the database, so if the object never got committed the fields won't be there. This occurred when the ProjectExists exception tries to turn the object to a dictionary.

PR #468, opened 2017-03-27
Add code coverage support with pytest-cov and codecov

jeremycline:
This generates code coverage reports with pytest-cov and enables
integration with codecov to report coverage on PRs.


pypingou:
Changes look fine to me

PR #470, opened 2017-04-11
Implement version plugins

jeremycline:
This is a v2 of https://github.com/release-monitoring/anitya/pull/448, but it should not change any user-facing behaviour and is strictly a refactor. It should make it easy to add additional version schemes later.

This is still a pretty large PR, but can reviewed commit-by-commit and each commit is reasonably sized and self-contained.

### DIFF
--- a/anitya/__init__.py
+++ b/anitya/__init__.py
@@ -4,14 +4,6 @@ import logging
 
 import anitya.lib.plugins
 import anitya.lib.exceptions
-# Feature check: version comparisons were historically done with cmp
-# For Python 3, we need to switch to using a key function instead
-try:
-    cmp
-except NameError:
-    from functools import cmp_to_key as _convert_cmp_to_key
-else:
-    _convert_cmp_to_key = None
 
 
 __api_version__ = '1.0'
@@ -35,24 +27,6 @@ def fedmsg_publish(*args, **kwargs):  # pragma: no cover
         fedmsg.publish(*args, **kwargs)
     except Exception as err:
         _log.error(str(err))
-
-
-# Ordering is handled differently based on Python version
-if _convert_cmp_to_key is None:
-    def order_versions(vlist):
-        ''' For a provided list of versions, return the list ordered from the
-        oldest to the newest version.
-        '''
-        import anitya.lib.backends  # Avoid circular import
-        return sorted(vlist, cmp=anitya.lib.backends.upstream_cmp)
-else:
-    def order_versions(vlist):
-        ''' For a provided list of versions, return the list ordered from the
-        oldest to the newest version.
-        '''
-        import anitya.lib.backends  # Avoid circular import
-        sort_key = _convert_cmp_to_key(anitya.lib.backends.upstream_cmp)
-        return sorted(vlist, key=sort_key)
 
 
 def check_release(project, session, test=False):
@@ -98,7 +72,8 @@ def check_release(project, session, test=False):
 
     odd_change = False
     if up_version and up_version != p_version:
-        max_version = order_versions([up_version, p_version])[-1]
+        version_class = project.get_version_class()
+        max_version = max(up_version, p_version, key=version_class)
         if project.latest_version and max_version != up_version:
             odd_change = True
             project.logs = 'Something strange occured, we found that this '\
--- a/anitya/lib/backends/__init__.py
+++ b/anitya/lib/backends/__init__.py
@@ -30,81 +30,8 @@ import requests
 import anitya
 import anitya.app
 from anitya.lib.exceptions import AnityaPluginException
+from anitya.lib.versions import RpmVersion
 import six
-
-try:
-    from rpm import labelCompare as _compare_rpm_labels
-except ImportError:
-    # Emulate RPM field comparisons as described in
-    # http://stackoverflow.com/questions/3206319/how-do-i-compare-rpm-versions-in-python/3206477#3206477
-    #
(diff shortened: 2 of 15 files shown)


anitya/lib/model.py
```diff
@@ -334,8 +364,42 @@ def validate_ecosystem_name(self, key, value):
...
+
+        1. On the project itself in the ``version_scheme`` column.
+        2. The project's ecosystem default, if the project is part of one.
+        3. The project's backend default, if the backend defines one.
+        4. The global default defined in :data:`anitya.lib.versions.GLOBAL_DEFAULT`
+
+        Returns:
+            anitya.lib.versions.Version: A ``Version`` sub-class.
+        """
+        from .plugins import ECOSYSTEM_PLUGINS, BACKEND_PLUGINS, VERSION_PLUGINS
```
bowlofeggs: Ooof. Avoiding circular imports?

jeremycline: Yup. I have a plan to fix them, but this PR actually makes it easier to do later (as it rips out one of the things defined in ``anitya/__init__.py``. I started to do it in a separate branch before this gets merged, but realized that having this in would make untangling the imports easier. This is very much a short-term work-around. A purge of in-function imports is coming.

anitya/lib/backends/__init__.py
```diff
@@ -30,82 +30,9 @@
...
 import anitya.app
 from anitya.lib.exceptions import AnityaPluginException
+from anitya.lib.versions import RpmVersion
 import six
 
-try:
-    from rpm import labelCompare as _compare_rpm_labels
-except ImportError:
-    # Emulate RPM field comparisons as described in
-    # http://stackoverflow.com/questions/3206319/how-do-i-compare-rpm-versions-in-python/3206477#3206477
```
jeremycline: @ncoghlan actually implemented this based on the description in the referenced answer (which in turn is a summary of the RPM C implementation) and kindly posted his code to the SO answer. I don't think that's problematic.

anitya/lib/versions/rpm.py
```diff
@@ -0,0 +1,192 @@
...
+import re
+
+from .base import Version
+
+
+try:
+    from rpm import labelCompare as _compare_rpm_labels
+except ImportError:
+    # Emulate RPM field comparisons as described in
+    # http://stackoverflow.com/questions/3206319/how-do-i-compare-rpm-versions-in-python/3206477#3206477
```
pypingou: It does have a license by default:
````
You agree that all Subscriber Content that You contribute to the Network is perpetually and irrevocably licensed to Stack Exchange under the Creative Commons Attribution Share Alike license. 
````
source: https://stackexchange.com/legal

And CC-BY-SA is not compatible with the GPL license making taking code from SO a little problematic :(

File: anitya/__init__.py [hunk 4]
```diff
@@ -98,7 +72,9 @@ def check_release(project, session, test=False):
 
     odd_change = False
     if up_version and up_version != p_version:
-        max_version = order_versions([up_version, p_version])[-1]
+        version_class = project.get_version_class()
+        max_version = sorted([version_class(version=v) for v in (up_version, p_version)])[-1]
```
ncoghlan: `max_version = max(up_version, p_version, key=version_class)` should work here.

anitya/lib/backends/__init__.py
ncoghlan: Plus we only use it to let the test suite pass in the absence of the RPM C bindings, which is common in local testing under tox, but shouldn't happen in a production deployment using the system site-packages.

anitya/lib/versions/rpm.py
ncoghlan: @jeremycline noted this elsewhere, but putting it here explicitly:

- the pure Python fallback code was written by me, on Red Hat's time, to allow Anitya's test suite to run under the Fedora python 2.6 stack in tox, which doesn't have any of the system bindings available
- the part that came from Stack Overflow was the prose description of what the C level RPM code actually does
- I posted the code back to Stack Overflow as a new answer, since that's potentially easier for folks to grab than writing their own from scratch (plus it slightly increases the chance of someone else reviewing it in detail and spotting any errors in it)

From a licensing point of view, that means for most people, it's effectively dual licensed under GPLv2+ (via Anitya), and CC-BY-SA (via Stack Overflow), while Red Hat can do whatever it wants (since I wrote it on work time, for work-related development).

anitya/tests/lib/versions/test_rpm.py
```diff
@@ -0,0 +1,226 @@
...
+        """Assert two prereleases of the same type without versions are equal."""
+        old_version = rpm.RpmVersion(version='1.0.0beta')
+        new_version = rpm.RpmVersion(version='1.0.0beta')
+        self.assertTrue(new_version == old_version)
+
+    def test_eq_one_v_prefix(self):
+        """Assert versions that are the same except one has a v prefix are equal."""
+        old_version = rpm.RpmVersion(version='v1.0.0-1.fc26')
+        new_version = rpm.RpmVersion(version='1.0.0-1.fc26')
+        self.assertTrue(new_version == old_version)
```
ncoghlan: Huh, if these still pass when the system RPM bindings aren't available, I guess my fallback emulation code actually works as advertised and I can remove that caveat from the Stack Overflow answer :)

bowlofeggs:
Cool - I didn't realize the @ncoghlan had written the code. My copyright fears are assuaged!

PR #472, opened 2017-04-11
Add an EL7 requirements file so tox can test with EL7 dependency versions

jeremycline:
At the moment we have good test coverage on Python 2.7-3.6 with the latest versions of all our dependencies. However, release-monitoring.org currently runs on EL7. This makes it easy to test against the library versions available on EL7.


pypingou:
Code wise this seems fine but keeping updated the el7 list sounds a little painful.

Up to you.

👍 for me

jeremycline:
I expect the way it'll stay updated is we'll break something in that test, double check that is still the latest version in EPEL, and then either be sad or update the list.

If it turns out to be more trouble than it's worth, though, I'll drop it.

PR #473, opened 2017-04-12
Remove unused dependencies

jeremycline:
When we stopped building the RST files to docs and started using Sphinx, we no longer have dependencies on markupsafe or docutils. This removes them.


pypingou:
Fair 👍

PR #480, opened 2017-04-19
Adjust the path to the requirements.txt files in Ansible

jeremycline:
I moved all the requirements files to a directory, but neglected to
update the Ansible role.


pypingou:
👍

PR #481, opened 2017-04-19
Add a migration for the Project ``version_scheme`` column

jeremycline:
Somehow I missed the migration in #470 for the column I added. This adds
the necessary migration.


pypingou:
Easy enough, 👍

PR #487, opened 2017-05-04
Maven backend: Fix handling of projects with dots in artifactId

msimacek:
Dots in groupId need to be converted to slashes, but dots in artifactId
should stay intact.

The maven repository layout is documented at:
https://cwiki.apache.org/confluence/display/MAVENOLD/Repository+Layout+-+Final

### DIFF
--- a/anitya/lib/backends/maven.py
+++ b/anitya/lib/backends/maven.py
@@ -14,7 +14,10 @@ from anitya.lib.backends import BaseBackend, get_versions_by_regex
 from anitya.lib.exceptions import AnityaPluginException
 
 
-REGEX = r'\<a[^>]+\>(\d[^</]*)'
+VERSION_REGEX = re.compile(r'\<a[^>]+\>(\d[^</]*)')
+MAVEN_HOMEPAGE_RE = re.compile(r'https?://repo\d+.maven.org/')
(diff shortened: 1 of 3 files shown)


anitya/lib/backends/maven.py
```diff
@@ -62,11 +62,17 @@ def get_versions(cls, project):
             url = project.homepage
         else:
             coordinates = project.version_url or project.name
-            if ':' not in coordinates:
+            match = re.match(r'([^:]+):([^:]+)', coordinates)
```
jeremycline: Since this regular expression is not dynamically generated and will probably used fairly frequently, it's probably worth putting it at the module level and compiling it:
```
# At the module level
#: A regular expression used to validate that Maven ``version_url`` fields are actually coordinates,
#: which are just two strings, a group ID and an artifact ID separated by a colon. 
coordinates_regex = re.compile(r'([^:]+):([^:]+)')

...

# right here
match = coordinates_regex.match(coordinates)
```

jeremycline:
Looks good, thanks for writing tests!

@pkajaba, is this the same type of problem you're trying to solve in #452 with the URL probing?

jeremycline:
Great, thanks!

PR #490, opened 2017-05-30
Require the correct dateutil package in requirements.txt

### LINKED ISSUE #476: The dateutils in production is not the same as the dateutils in the dev environment
The specfile installs [python-dateutil](https://admin.fedoraproject.org/pkgdb/packages/python-dateutil). However, the upstream of that package is https://pypi.python.org/pypi/python-dateutil/ and our requirements file has [dateutils](https://pypi.python.org/pypi/dateutils/).

Presumably we want the one used in production, but we should also have tests that would have caught this particular problem. It may be the dependency isn't necessary at all.

- [ ] Find out if the dependency is necessary

- [ ] If it is necessary, write tests for the code that uses it

- [ ] Fix the requirements files to reference python-dateutil

jeremycline:
fixes #476

### DIFF
--- a/requirements/readthedocs.txt
+++ b/requirements/readthedocs.txt
@@ -4,7 +4,7 @@
 # the dependencies before readthedocs installs Anitya.
 alembic
 bunch
-dateutils
+python-dateutil
 fedmsg
 flask >= 0.10.1
 flask-openid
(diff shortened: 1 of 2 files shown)


pypingou:
👍

PR #499, opened 2017-06-16
allow the vagrant box to be provisioned with empty database

luto:
Sometimes a clean slate is better for local development. By default imports the production database so you've got something to work off of. With this change, it's possible to just creates the empty database schema.

Note that I'm not that deep into anitya yet. Creating an empty DB allows the service to start and also renders the main web views. I haven't tested anything more until now.


luto:
Note: fedora login works just fine with the empty DB.

PR #502, opened 2017-06-24
new backend: GitHub API

luto:
Anitya currently uses the GitHub releases HTML page to determine the versions for a project. While this works, it is likely to break with interface changes in the future. Parsing the HTML pages via regex is also probably less efficient than using the API.

This PR implements a simple backend, which reports a projects tags as versions.

Todo:
- [x] support authentication for much higher rate limits
- [ ] use [conditional requests](https://developer.github.com/v3/#conditional-requests) wherever possible to minimize the import on rate limits
- [ ] gracefully handle rate limits
- [ ] merge GitHub and GitHub API backends
  - [ ] fallback to HTML scraping if `github3` is not installed, or rate limit is hit

Please let me know what you think about this :)

### DIFF
--- a/anitya/lib/backends/githubapi.py
+++ b/anitya/lib/backends/githubapi.py
@@ -10,6 +10,7 @@
 
 import github3
 
+import anitya
 from anitya.lib.backends import BaseBackend
 from anitya.lib.exceptions import AnityaPluginException
 
@@ -72,7 +73,14 @@ class GithubApiBackend(BaseBackend):
             raise AnityaPluginException(
                 'Project %s was incorrectly set-up' % project.name)
 
-        repo = github3.repository(owner, repo)
+        github_access_token = anitya.app.APP.config.get('GITHUB_ACCESS_TOKEN')
+
+        if github_access_token:
+            gh = github3.login(token=github_access_token)
+        else:
+            gh = github3
+
+        repo = gh.repository(owner, repo)
 
         if not repo:
             raise AnityaPluginException(
--- a/anitya/lib/backends/githubapi.py
+++ b/anitya/lib/backends/githubapi.py
@@ -1,12 +1,23 @@
 # -*- coding: utf-8 -*-
-
-"""
- (c) 2014 - Copyright Red Hat Inc
-
- Authors:
-   Michael Lutonsky <m@luto.at>
-
-"""
+#
+# Copyright © 2014  Red Hat, Inc.
+#
+# This copyrighted material is made available to anyone wishing to use,
+# modify, copy, or redistribute it subject to the terms and conditions
+# of the GNU General Public License v.2, or (at your option) any later
+# version.  This program is distributed in the hope that it will be
+# useful, but WITHOUT ANY WARRANTY expressed or implied, including the
+# implied warranties of MERCHANTABILITY or FITNESS FOR A PARTICULAR
+# PURPOSE.  See the GNU General Public License for more details.  You
+# should have received a copy of the GNU General Public License along
+# with this program; if not, write to the Free Software Foundation,
+# Inc., 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301, USA.
+#
+# Any Red Hat trademarks that are incorporated in the source
+# code or documentation are not subject to the GNU General Public
+# License and may only be used or replicated with the express permission
+# of Red Hat, Inc.
+#
 
 import github3
 
--- a/anitya/lib/backends/githubapi.py
+++ b/anitya/lib/backends/githubapi.py
@@ -74,7 +74,7 @@ class GithubApiBackend(BaseBackend):
             url = project.homepage
         else:
             raise AnityaPluginException(
-                'Project %s was incorrectly set-up' % project.name)
+                'Project %s does not have a version URL or homepage' % project.name)
 
         url = url.replace('https://github.com/', '')
 
@@ -82,7 +82,7 @@ class GithubApiBackend(BaseBackend):
             owner, repo = url.split('/')
         except:
             raise AnityaPluginException(
-                'Project %s was incorrectly set-up' % project.name)
+                'The URL of project %s does not look like a github one' % project.name)
 
         github_access_token = anitya.app.APP.config.get('GITHUB_ACCESS_TOKEN')
 
@@ -95,7 +95,7 @@ class GithubApiBackend(BaseBackend):
 
         if not repo:
             raise AnityaPluginException(
-                'Project %s does not exists on GitHub' % project.name)
+                'Project %s does not exist on GitHub' % project.name)
 
         versions = []
 
--- a/anitya/lib/backends/githubapi.py
+++ b/anitya/lib/backends/githubapi.py
@@ -1,0 +1,94 @@
+# -*- coding: utf-8 -*-
+
+"""
+ (c) 2014 - Copyright Red Hat Inc
+
+ Authors:
+   Michael Lutonsky <m@luto.at>
+
+"""
+
+import github3
+
+from anitya.lib.backends import BaseBackend
+from anitya.lib.exceptions import AnityaPluginException
+
+
+class GithubApiBackend(BaseBackend):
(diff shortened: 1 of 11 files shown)


bowlofeggs:
I agree with the benefits, but I'll point out that GitHub API limits might present a problem if Anitya switches to using the API. I've not looked at this patch yet so apologies if it's implemented this way, but we might need some Anitya-side rate limiting to work around that.

anitya/lib/backends/githubapi.py
```diff
@@ -0,0 +1,102 @@
...
+        if github_access_token:
+            gh = github3.login(token=github_access_token)
+        else:
+            gh = github3
+
+        repo = gh.repository(owner, repo)
+
+        if not repo:
+            raise AnityaPluginException(
+                'Project %s does not exists on GitHub' % project.name)
```
bowlofeggs: s/exists/exist/

anitya/lib/backends/githubapi.py
```diff
@@ -0,0 +1,102 @@
+# -*- coding: utf-8 -*-
+
+"""
+ (c) 2014 - Copyright Red Hat Inc
+
+ Authors:
+   Michael Lutonsky <m@luto.at>
+
+"""
```
bowlofeggs: I noticed that the tests had a license declaration. Should we put one of those here too?

anitya/tests/lib/backends/test_githubapi.py
```diff
@@ -0,0 +1,157 @@
...
+            AnityaPluginException,
+            backend.GithubApiBackend.get_version,
+            project
+        )
+
+        pid = 3
+        project = model.Project.get(self.session, pid)
+        exp = '2.6.2'
+        obs = backend.GithubApiBackend.get_version(project)
+        self.assertEqual(obs, exp)
```
bowlofeggs: Are these tests connecting to GitHub over the Internet? I ask because I didn't notice any mocking. If they connect over the Internet, these tests will fail as the referenced project make new releases.

luto: The testing code within anitya uses [vcrpy](https://github.com/kevin1024/vcrpy) which automatically intercepts and caches all HTTP requests. The resulting files [like this one](https://github.com/release-monitoring/anitya/pull/502/files#diff-b7cdf64fa5606de2fc6c2c64c09cd94d) have been added to this PR. The tests do therefore not connect to github.

anitya/lib/backends/githubapi.py
luto: :white_check_mark:

anitya/lib/backends/githubapi.py
```diff
@@ -0,0 +1,102 @@
+# -*- coding: utf-8 -*-
+
+"""
+ (c) 2014 - Copyright Red Hat Inc
```
luto: :white_check_mark:

anitya/lib/backends/githubapi.py
luto: :white_check_mark:

anitya/lib/backends/githubapi.py
```diff
@@ -0,0 +1,102 @@
...
+            raise AnityaPluginException(
+                'Project %s was incorrectly set-up' % project.name)
+
+        url = url.replace('https://github.com/', '')
+
+        try:
+            owner, repo = url.split('/')
+        except:
+            raise AnityaPluginException(
+                'Project %s was incorrectly set-up' % project.name)
```
luto: :white_check_mark:

anitya/lib/backends/githubapi.py
```diff
@@ -0,0 +1,102 @@
...
+            when the versions cannot be retrieved correctly
+
+        '''
+        if project.version_url:
+            url = project.version_url
+        elif project.homepage.startswith('https://github.com'):
+            url = project.homepage
+        else:
+            raise AnityaPluginException(
+                'Project %s was incorrectly set-up' % project.name)
```
luto: :white_check_mark:

luto:
> I agree with the benefits, but I'll point out that GitHub API limits might present a problem if Anitya switches to using the API.

Assuming release-monitoring.org uses authentication on github.com plus the [conditional requests](https://developer.github.com/v3/#conditional-requests), which I'll implement soon, this requires antiya to discover 5.000 _new releases_ within a single hour. API requests which do not discover new data do not contribute to the rate limit. While I agree that this might be a problem in an abuse case (e.g. a malicious user creates 5.000+ new projects), this should not happen with normal usage. I'm assuming that GitHub will also take action, when antiya hits the HTML pages with such rates.

luto:
Thank you. I've changed the code to incorporate your notes. Let me know if you notice anything else!

bowlofeggs:
On Mon, 2017-06-26 at 11:17 -0700, luto wrote:
> The testing code within anitya uses vcrpy which automatically
> intercepts and caches all HTTP requests. The resulting files like
> this one have been added to this PR. The tests do therefore not
> connect to github.

Ah yes. I've been meaning to look into integrating that with Bodhi's
tests too.

luto:
Glad to hear you like it!

> I think we should have this be part of the existing GitHub backend.

Alright. I'll look into merging them.

> On the topic of rate limiting, I think it would be good to make sure we fail gracefully if we hit our rate limit. Does github3 raise an exception when an API request is denied due to rate limiting? Or is the call to github.rate_limit() cheap (e.g. it's tracked locally by the library) so we can call it before each request?

I didn't look into the specifics of handling the rate limit within github3 yet. But I agree that a graceful fail would be preferable (graceful meaning raising a proper `AnityaPluginException`). This will be implemented after I've got the conditional requests working.

ishitatsuyuki:
This seems to be inactive for a while, but FYI you can fetch at most 500k nodes per hour with GraphQL: https://developer.github.com/v4/guides/resource-limitations/

yarons:
@ishitatsuyuki Requires a user, most likely will be block after reaching rate limit for several times a week or even before that.

ishitatsuyuki:
@yarons Huh? Can you elaborate? With ~10k projects it's unlikely that we'll hit the limit, and if we get more, we can just split the work over hours.

yarons:
@ishitatsuyuki 
Well, we can have up to 5,000 calls per hour:

> ### Rate limit
> 
> The GraphQL API v4 limit is different from the REST API v3's limit of 5,000 requests per hour for authenticated requests.

So I guess it's a non-issue.
In this case I've already wrote the query in #567.

anitya/tests/lib/backends/test_githubapi.py
```diff
@@ -0,0 +1,157 @@
+# -*- coding: utf-8 -*-
+#
+# Copyright © 2014  Red Hat, Inc.
```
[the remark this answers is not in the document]
yarons: Maybe 2018 by now :smile:

ishitatsuyuki:
I don't think so. This project is not super active, so it's likely left in the current state until an alternative backend is implemented.

yarons:
So there will be no update until the backend will be updated?
From what I understood, Alpine is relying on this system.

jeremycline:
I'm happy to review and help get a new backend in, but the time I can devote to implementation is limited. It's on my to-do list, but it's not likely to get done anytime soon.

I'm not responsible (nor do I have the access) for the production deployments of anitya to release-monitoring.org. The Fedora infrastructure team (infrastructure@lists.fedoraproject.org or https://pagure.io/fedora-infrastructure/issues) handles it. There's currently a staging deployment of 0.12.0 at https://release-monitoring.app.os.stg.fedoraproject.org/.

yarons:
Hi @jeremycline, I'll try to pick up some small bugs myself, I'm not much of an expert but I can handle Regex and JSON selectors pretty easily so maybe I can help with this part.

I see regarding the production, the staging instance is running into an error when trying to login, yet, I'm really happy things are going the right way.

I'll try filing a bug on pagure.

Thank you!

PR #503, opened 2017-07-11
Remove functions from __init__.py that don't belong there

jeremycline:
Anitya suffers from very circular imports. These are caused by defining
inter-dependent functions in various ``__init__.py``. These modules are
for package initialization. This moves all the code from
``anitya/__init__.py`` and ``anitya/lib/__init__.py`` into
``anitya/lib/utilities.py`` (not a great name, but most of the functions
inside it need to be factored out so I don't expect it to remain for too
long).

This should make working on the backend much less painful since it can
now import ``anitya.config`` without causing the entire backend to be
imported before the configuration can be imported.


anitya/__init__.py
```diff
@@ -1,172 +1,4 @@
...
-    ))
-
-    model.Log.insert(
-        session,
-        user=message['agent'],
-        project=project,
-        distro=distro,
-        description=final_msg)
-
-    return final_msg
```
bowlofeggs: Are we confident that third party programs wouldn't be importing these from here? Some of them were named with an underscore, but some were not. If we aren't sure, should we import those things here so they can still be imported? Or maybe bump the X version?

anitya/lib/utilities.py
```diff
@@ -0,0 +1,508 @@
+# -*- coding: utf-8 -*-
```
bowlofeggs: The project doesn't seem to have an overall copyright claim. I recommend claiming copyright on this file (and I recommend correcting this for the rest of the project too). Don't forget that this code came from existing files, so the copyright claim should take the history of those files into account for ownership and dates.

anitya/__init__.py
jeremycline: Nope, this breaks APIs, but there's not a good way around it. We're still versioned 0.y.z and I intend that to mean its API is not stable. There are going to be further breakages as I untangle the modules, unfortunately.

I'll make nice release notes about it, though.

jeremycline:
Okay, the copyright claim has been fixed and I wrote some release notes. I'd appreciate it if you cast your eye over the notes to see if they're nice enough.

bowlofeggs:
On Tue, 2017-07-11 at 18:18 +0000, Jeremy Cline wrote:
> I'd appreciate it if you cast your eye over the notes to see if
> they're nice enough.

LGTM!

PR #506, opened 2017-07-21
Refactor the Flask application to use an app factory

jeremycline:
This refactor removes the module-level ``anitya.app.APP`` object and replaces it with a function that will create and configure a complete Flask application. This, among other things, makes testing easier since each test can create, configure, and modify its own Flask application object. Check out http://flask.pocoo.org/docs/0.12/patterns/appfactories/ for an overview of the general pattern.

I did my best to break this up into a series of commits that are atomic and make sense. They're probably not as neat as I'd like, but hopefully they make this PR a little less painful to review (sorry!).

### DIFF
--- a/anitya/app.py
+++ b/anitya/app.py
@@ -10,7 +10,6 @@ User-facing Flask routes should be placed in the ``anitya.ui`` module and API
 routes should be placed in ``anitya.api_v2``.
 """
 
-import functools
 import logging
 import logging.config
 import logging.handlers
@@ -69,7 +68,6 @@ def create(config=None):
     app.oid.loginhandler(ui.yahoo_login)
     app.oid.loginhandler(ui.google_login)
 
-
     if app.config.get('EMAIL_ERRORS'):
         # If email logging is configured, set up the anitya logger with an email
         # handler for any ERROR-level logs.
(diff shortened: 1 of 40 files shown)


anitya/app.py
```diff
@@ -79,37 +77,12 @@ def create(config=None):
             mail_admin=app.config.get('ADMIN_EMAIL')
         ))
 
-
```
bowlofeggs: These space changes are good, but would be better to put into an atomic commit.

anitya/app.py
```diff
@@ -50,26 +49,20 @@ def create(config=None):
 
     if config is None:
         config = anitya_config
-
```
bowlofeggs: This change would be better to do in its own commit as well.

anitya/app.py
```diff
@@ -50,26 +49,20 @@ def create(config=None):
...
 
-    # Mark login handlers
-    app.oid.loginhandler(ui.login)
-    app.oid.loginhandler(ui.fedora_login)
-    app.oid.loginhandler(ui.yahoo_login)
-    app.oid.loginhandler(ui.google_login)
-
     app.before_request(check_auth)
     app.teardown_request(shutdown_session)
+
```
bowlofeggs: Here as well.

anitya/ui.py
```diff
@@ -10,8 +11,26 @@
...
 from anitya.lib import utilities
+from anitya.lib.model import Session as SESSION
+
+
+ui_blueprint = flask.Blueprint(
+    'anitya_ui', __name__, static_folder='static', template_folder='templates')
+
+
+def login_required(function):
+    ''' Flask decorator to retrict access to logged-in users. '''
```
jeremycline: Normally, I'd be all for this (you know how much I like documentation), but I also plan on replacing this with [flask-login's version](https://flask-login.readthedocs.io/en/latest/#flask_login.login_required) because it integrates nicely with [python-social-auth](http://python-social-auth.readthedocs.io/en/latest/configuration/flask.html)

ncoghlan:
I don't have an opinion on the exact details of the changes, but a big +1 for the general idea (especially if it gets you closer to enabling full social auth support).

PR #507, opened 2017-08-03
Add users to Anitya and provide authenticated API access

jeremycline:
This is still very much a work in progress, but just in case people are interested in what I've been up to...

This adds users via python-social-auth and starts using flask-login to manage logins. It _also_ includes API tokens. What it doesn't include is a tidy login page, a UI for generating API tokens, or a way to blacklist users.

Early feedback very much appreciated 😃

### DIFF
--- a/alembic/versions/feeaa70ead67_social_authentication_table.py
+++ b/alembic/versions/feeaa70ead67_social_authentication_table.py
@@ -1,0 +1,89 @@
+"""Social authentication table
+
+Revision ID: feeaa70ead67
+Revises: a52d2fe99d4f
+Create Date: 2017-07-07 14:23:30.605238
+"""
+
+from alembic import op
+from social_sqlalchemy import storage
+import sqlalchemy as sa
+
+from anitya.lib.model import GUID
+
+
+# revision identifiers, used by Alembic.
+revision = 'feeaa70ead67'
+down_revision = 'a52d2fe99d4f'
+
+
+def upgrade():
+    op.create_table(
+        'social_auth_association',
+        sa.Column('id', sa.Integer(), nullable=False),
+        sa.Column('server_url', sa.String(length=255), nullable=True),
+        sa.Column('handle', sa.String(length=255), nullable=True),
+        sa.Column('secret', sa.String(length=255), nullable=True),
+        sa.Column('issued', sa.Integer(), nullable=True),
+        sa.Column('lifetime', sa.Integer(), nullable=True),
+        sa.Column('assoc_type', sa.String(length=64), nullable=True),
+        sa.PrimaryKeyConstraint('id'),
+        sa.UniqueConstraint('server_url', 'handle')
+    )
+    op.create_table(
+        'social_auth_code',
+        sa.Column('id', sa.Integer(), nullable=False),
+        sa.Column('email', sa.String(length=200), nullable=True),
+        sa.Column('code', sa.String(length=32), nullable=True),
+        sa.PrimaryKeyConstraint('id'),
+        sa.UniqueConstraint('code', 'email')
+    )
+    op.create_index(op.f('ix_social_auth_code_code'), 'social_auth_code', ['code'], unique=False)
+    op.create_table(
+        'social_auth_nonce',
+        sa.Column('id', sa.Integer(), nullable=False),
+        sa.Column('server_url', sa.String(length=255), nullable=True),
+        sa.Column('timestamp', sa.Integer(), nullable=True),
+        sa.Column('salt', sa.String(length=40), nullable=True),
+        sa.PrimaryKeyConstraint('id'),
+        sa.UniqueConstraint('server_url', 'timestamp', 'salt')
+    )
+    op.create_table(
+        'social_auth_partial',
+        sa.Column('id', sa.Integer(), nullable=False),
+        sa.Column('token', sa.String(length=32), nullable=True),
+        sa.Column('data', storage.JSONType(), nullable=True),
+        sa.Column('next_step', sa.Integer(), nullable=True),
+        sa.Column('backend', sa.String(length=32), nullable=True),
+        sa.PrimaryKeyConstraint('id')
+    )
+    op.create_index(
+        op.f('ix_social_auth_partial_token'), 'social_auth_partial', ['token'], unique=False)
+    op.create_table(
+        'social_auth_usersocialauth',
+        sa.Column('id', sa.Integer(), nullable=False),
+        sa.Column('provider', sa.String(length=32), nullable=True),
+        sa.Column('extra_data', storage.JSONType(), nullable=True),
+        sa.Column('uid', sa.String(length=255), nullable=True),
+        sa.Column('user_id', GUID, nullable=False),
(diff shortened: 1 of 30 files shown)


anitya/authentication.py
```diff
@@ -17,129 +17,104 @@
...
 
-def register_oidc_client(code, state):
-    """Accept OpenIDConnect callback from authentication server"""
-    # TODO: Actually handle online token generation properly
-    #       https://github.com/release-monitoring/anitya/issues/442
-    token_data = {}
-    result = flask.jsonify(token_data)
-    result.status_code = 200
-    return result
+    This implements the interface required by :meth:`flask_login.LoginManager.user_loader`.
```
bowlofeggs: You shouldn't use ```meth```.

anitya/config.py
```diff
@@ -81,19 +76,21 @@
...
-    OIDC_CLIENT_SECRETS=os.path.join(
-        os.path.dirname(os.path.dirname(os.path.abspath(__file__))),
-        'client_secrets.json'
+    SESSION_PROTECTION='strong',
+    SOCIAL_AUTH_AUTHENTICATION_BACKENDS=(
+        'social_core.backends.fedora.FedoraOpenId',
+        'social_core.backends.yahoo.YahooOpenId',
+        'social_core.backends.open_id.OpenIdAuth',
+        # 'social_core.backends.google_openidconnect.GoogleOpenIdConnect',
+        # 'social_core.backends.github.GithubOAuth2',
```
bowlofeggs: Might as well drop these instead of commenting them.

anitya/lib/model.py
```diff
@@ -20,17 +20,28 @@
 anitya mapping of python classes to Database Tables.
 """
 
+try:
+    # The Python 3.6+ API
+    from secrets import choice as random_choice
+except ImportError:
+    from random import choice as random_choice
```
bowlofeggs: dear python, pls to follow semantic versioning. kthxbye

anitya/lib/model.py
```diff
@@ -20,17 +20,28 @@
...
 import sqlalchemy as sa
-from sqlalchemy.orm import validates
+from sqlalchemy.dialects.postgresql import UUID
+from sqlalchemy.orm import (
+    sessionmaker, scoped_session, query as sa_query, validates)
 from sqlalchemy.ext.declarative import declarative_base
 from sqlalchemy.orm.exc import NoResultFound
-from sqlalchemy.orm import sessionmaker, scoped_session, query as sa_query
+from sqlalchemy.types import TypeDecorator, CHAR
+import uuid
```
bowlofeggs: ```uuid``` is part of the standard library, so this should be moved to the first import group for PEP-8.

anitya/lib/model.py
```diff
@@ -908,3 +919,191 @@ def last_entry(cls, session):
...
+    Generate an API token of a given length using the provided character set.
+
+    Args:
+        charset (str): A string of characters to choose from in the token.
+        length (int): The number of characters to use in the token.
+
+    Returns:
+        str: The API token as a unicode string.
+    """
+    return u''.join(random_choice(charset) for __ in range(length))
```
bowlofeggs: Note the warning in the [random](https://docs.python.org/3.5/library/random.html) docs:

```
Warning The pseudo-random generators of this module should not be used for security purposes.
```

I saw at the top that you use the Python 3.6 [```secrets```](https://docs.python.org/3.6/library/secrets.html) module if available, which does say that it can be used for "generating cryptographically strong random numbers suitable for managing data such as passwords, account authentication, security tokens, and related secrets."

anitya/tests/test_flask_admin.py
```diff
@@ -17,566 +17,737 @@
...
 
-            output = c.post(
-                '/distro/add', data=data, follow_redirects=True)
-            self.assertEqual(output.status_code, 200)
-            self.assertTrue(
-                b'<h1>Distributions participating</h1>' in output.data)
-            self.assertTrue(
-                b'<a href="/distro/Debian/edit">' in output.data)
+            # self.assertEqual(201, create_output.status_code)
+            # self.assertEqual(409, dup_output.status_code)
```
bowlofeggs: Might want to drop or uncomment these two lines.

pypingou:
> a way to blacklist users.

Note that this was a feature that was added because someone, despite multiple discussion over emails, was not being friendly, but I don't think it has ever been used (in other words, you could postpone it for later if you want)

anitya/authentication.py
jeremycline: Not even once?

anitya/lib/model.py
jeremycline: ``random.choice`` is still present in 3.6 so it actually does.

bowlofeggs:
On 09/18/2017 01:47 PM, Jeremy Cline wrote:
> Not even once?

I guess it comes down to a question: how much do you value your teeth?

anitya/lib/model.py
bowlofeggs: Ah, then I stand happily corrected.

anitya/lib/model.py
jeremycline: Good catch, I added that and meant to come back and find the equivalent non-Python 3.6 API, but clearly didn't. I changed it so that if secrets isn't available, it falls back to [SystemRandom](https://docs.python.org/3.5/library/random.html#random.SystemRandom) which uses ``os.urandom`` rather than the Mersenne Twister for random numbers.

This is in line with the [recommendations of cryptography](https://cryptography.io/en/stable/random-numbers/) and is what Django does as well.

jeremycline:
Okay, I think I covered all your comments @bowlofeggs, thanks for the review! Take a look at things (especially the random bit) and let me know what you think.

@pypingou, that's good to know. I don't think it'll be difficult to implement (Flask-Login has a way to disable accounts by default), I just didn't want this PR to be any bigger than it already is :smile:.

alembic/versions/feeaa70ead67_social_authentication_table.py
```diff
@@ -1,23 +1,109 @@
...
+            value (object): The database value.
+            dialect (sqlalchemy.engine.interfaces.Dialect): The dialect in use.
+
+        Returns:
+            uuid.UUID: The value as a Python :class:`uuid.UUID`.
+        """
+        if value is None:
+            return value
+        else:
+            return uuid.UUID(value)
```
bowlofeggs: It might be a nice compromise to have only the first migration that needs this copy of ```GUID``` to copy it from the models, and have the other migrations that use the exact same code import it from the migration. This way you avoid the "what if GUID changes in some way that breaks the migration" problem, but you also don't have to have n copies of it.

anitya/lib/model.py
```diff
@@ -24,12 +24,16 @@
     # The Python 3.6+ API
     from secrets import choice as random_choice
 except ImportError:
-    from random import choice as random_choice
+    # Fall back to random with os.urandom
+    import random
+    random = random.SystemRandom()
+    random_choice = random.choice
```
bowlofeggs: LGTM.

PR #515, opened 2017-11-14
Drop the dependency on bunch

jeremycline:
Bunch isn't used, so we shouldn't depend on it.

### DIFF
--- a/CHANGELOG.rst
+++ b/CHANGELOG.rst
@@ -4,6 +4,12 @@ Changes
 dev (master)
 ------------
 
+Dependencies
+------------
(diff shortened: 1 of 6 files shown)


pypingou:
IIRC it was/is used by flask-openid, it could be that it was added because of a missing dep back then.

jeremycline:
> IIRC it was/is used by flask-openid, it could be that it was added because of a missing dep back then.

Ah, good to know. #518 drops flask-openid since we no longer use it so we should be okay even if the missing dep is still a problem.

PR #517, opened 2017-11-14
Unify the various WSGI applications

jeremycline:
For some reason there were several WSGI app files. This unifies them
inside the Anitya package so users can do something like

```
gunicorn --bind 0.0.0.0:8080 anitya.wsgi:application
```


anitya/tests/test_wsgi.py
```diff
@@ -0,0 +1,12 @@
+import unittest
```
bowlofeggs: You might put a copyright claim on the test file too.

PR #523, opened 2018-01-10
Make the login page less ugly

jeremycline:
When I switched to python-social-auth a made a really ugly login page. This is still pretty ugly, but less so.


pypingou:
Didn't check how it renders but the code looks fine to me :)

jeremycline:
Render

It's not very pretty, but it's a little better than what I had before.

PR #524, opened 2018-01-16
Move the database-related code into anitya.db

jeremycline:
Move the models and session setup to ``anitya.db`` and then remove all references to ``anitya.lib.model``.

This contains no new code, just search-and-replace or module splitting (model.py because meta.py and models.py).


anitya/db/__init__.py
```diff
@@ -0,0 +1,9 @@
+"""
```
bowlofeggs: Might be good to put a Copyright header in this and other new files. Unless the project has a global copyright claim?

PR #525, opened 2018-01-17
Add a basic v2 API for packages

jeremycline:
This adds an API to create and see packages in Anitya. 

There's some prep work here that makes the ecosystem field non-nullable  and to automatically set it for projects. I'm setting it to a project's homepage if it's not part of an ecosystem (such as PyPI). This lets users uniquely identify projects using the name and ecosystem. 

In most cases this would be something like ("PyPI", "requests") so that's nice and simple. GCC would be ("https://gcc.gnu.org/", "gcc").

### DIFF
--- a/anitya/api_v2.py
+++ b/anitya/api_v2.py
@@ -10,6 +10,8 @@ from gettext import gettext as _
 import flask_login
 from flask import jsonify
 from flask_restful import Resource, reqparse
+from sqlalchemy.exc import IntegrityError
+from sqlalchemy.orm.exc import NoResultFound
 
 from anitya import authentication
 from anitya.db import Session, models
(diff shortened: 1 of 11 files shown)


anitya/api_v2.py
```diff
@@ -63,6 +65,183 @@ def _items_per_page_validator(arg):
     return arg
 
 
+class PackagesResource(Resource):
+    """The ``api/v2/packages/`` API endpoint."""
+
+    def get(self):
+        """
+        List all packages
```
bowlofeggs: s/packages/packages./

anitya/api_v2.py
```diff
@@ -63,6 +65,183 @@ def _items_per_page_validator(arg):
...
+        **Example request**:
+
+        .. sourcecode:: http
+
+            GET /api/v2/packages/?name=0ad&distribution=Fedora HTTP/1.1
+            Accept: application/json
+            Accept-Encoding: gzip, deflate
+            Connection: keep-alive
+            Host: localhost:5000
+            User-Agent: HTTPie/0.9.4
```
bowlofeggs: the ie is for internet explorer

anitya/db/events.py
```diff
@@ -0,0 +1,64 @@
...
+# implied warranties of MERCHANTABILITY or FITNESS FOR A PARTICULAR
+# PURPOSE.  See the GNU General Public License for more details.  You
+# should have received a copy of the GNU General Public License along
+# with this program; if not, write to the Free Software Foundation,
+# Inc., 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301, USA.
+#
+# Any Red Hat trademarks that are incorporated in the source
+# code or documentation are not subject to the GNU General Public
+# License and may only be used or replicated with the express permission
+# of Red Hat, Inc.
```
bowlofeggs: I've not seen this clause before in a file header.

anitya/db/migrations/versions/34b9bb5fa388_make_ecosystem_non_nullable.py
```diff
@@ -0,0 +1,36 @@
+"""Make ecosystem non-nullable
```
bowlofeggs: Copyright/license block might should go here.

anitya/tests/test_flask_api_v2.py
```diff
@@ -35,6 +35,475 @@ def _read_json(output):
...
+    def setUp(self):
+        super(PackagesResourceGetTests, self).setUp()
+        self.app = self.flask_app.test_client()
+        session = Session()
+        self.user = models.User(email='user@example.com', username='user')
+        self.api_token = models.ApiToken(user=self.user)
+        fedora = models.Distro('Fedora')
+        debian = models.Distro('Debian')
+        jcline_linux = models.Distro('jcline linux')
+        session.add_all([self.user, self.api_token, fedora, debian, jcline_linux])
```
bowlofeggs: where can i download jcline_linux? how does it compare to [dros](https://github.com/jeremycline/dros)?

anitya/tests/test_flask_api_v2.py
```diff
@@ -35,6 +35,475 @@ def _read_json(output):
...
+    def test_no_packages(self):
+        """Assert querying packages works, even if there are no packages."""
+        output = self.app.get('/api/v2/packages/')
+        self.assertEqual(output.status_code, 200)
+        data = _read_json(output)
+
+        self.assertEqual(data, {'page': 1, 'items_per_page': 25, 'total_items': 0, 'items': []})
+
+    def test_authenticated(self):
+        """Assert the API works when authenticated."""
```
bowlofeggs: Should there also be a ```test_unauthenticated```?

anitya/api_v2.py
jeremycline: You caught me

anitya/tests/test_flask_api_v2.py
jeremycline: The rest of the tests don't use authentication so it's tested in the basic success test

PR #527, opened 2018-01-24
Include the ecosystem in the Project dict from __json__

jeremycline:
When https://github.com/release-monitoring/anitya/pull/525 is accepted, people will want to use the ecosystem for a project to create packages via the API, so include it in the response.


anitya/tests/db/test_models.py
```diff
@@ -94,6 +94,17 @@ def test_validate_ecosystem_good(self):
...
+    def test_ecosystem_in_json(self):
+        """Assert the ecosystem is included in the dict returned from ``__json__``"""
+        project = models.Project(
+            name='test',
+            homepage='http://example.com',
+            backend='custom',
+            ecosystem_name='pypi',
+        )
+        self.assertTrue('ecosystem' in project.__json__())
+        self.assertEqual('pypi', project.__json__()['ecosystem'])
```
bowlofeggs: Doesn't the success of the second assertion imply the success of the first?

PR #529, opened 2018-02-07
Use a platform whitelist for libraries.io

jeremycline:
There's over 600,000 Javascript projects we don't care about tracking,
so allow platform whitelisting for automatic creation.


pypingou:
Code looks fine to me

PR #532, opened 2018-03-01
Switch over to accessing pypi.org

### LINKED ISSUE #531: Switch to querying pypi.org instead of pypi.python.org
PyPI is in the progress of migrating away from the existing implementation at pypi.python.org to the new implementation at pypi.org.

For Anitya, the core changes for this should be limited to https://github.com/release-monitoring/anitya/edit/master/anitya/lib/backends/pypi.py, where we'll want to:

* switch at least one example URL over to the new site layout
* query the new service for version information
* query the new service for the RSS feed

We'll also want to ensure that the tests continue to cover registering with both the old-style and new-style project URLs.

The notification issue is at https://github.com/pypa/warehouse/issues/2935, so I'll post any question we have there.

ncoghlan:
The PSF & PyPA are migrating PyPI to a new implementation hosted
at pypi.org, and will eventually redirect all pypi.python.org acesses to
that new version.

This switches the PyPI backend over to using the new implementation
(they're backed by the same database, so all the information will be the
same).

Closes #531


ncoghlan:
Ah, because I changed the default home page for feed-created projects, vcrpy's existing recordings for affected tests have been invalidated. I'll need to regenerate them before this PR will pass CI.

ncoghlan:
@jeremycline Thanks for fixing those!

PR #534, opened 2018-03-09
Fix readthedocs requirements

jeremycline:
sphinx no longer needs pinning and the build needs social-auth installed.

Build tested with http://jcline-anitya.readthedocs.io/en/latest/


jeremycline:
always :heart:

PR #535, opened 2018-03-12
Update npmjs backend

QuLogic:
The feed is deprecated and now returning nothing. Based on this [blog post](http://blog.npmjs.org/post/142077474335/npm-registry-is-now-fully-https), I think I've found a suitable replacement, though it's not exactly obvious.

Also, npmjs uses https by default, so update the links and then update the tapes with current results.


anitya/lib/backends/npmjs.py
```diff
@@ -112,8 +117,11 @@ def check_feed(cls):
         except Exception:  # pragma: no cover
             raise AnityaPluginException('No JSON returned by %s' % url)
 
-        for item in data[:40]:
+        for item in data['results']:
+            if item.get('deleted'):
+                continue
+            item = item['doc']
```
jeremycline: Minor nitpick, but this caught me a bit off-guard. Could you either change the variable name here to something like ``item_doc`` or just do ``item['doc']['name']``? At first I expected to find the 'name' key in the 'results' object.

QuLogic: Fixed.

PR #541, opened 2018-03-12
Use https links wherever else available

QuLogic:
This covers using https links everywhere that doesn't touch the tapes (and thus would be in my other PRs).


QuLogic:
Another thing that should probably be changed is `*.id.fedoraproject.org`, but I wasn't sure if I should change the config or not.

jeremycline:
> Another thing that should probably be changed is *.id.fedoraproject.org, but I wasn't sure if I should change the config or not.

Ah, that configuration needs to be updated since it now uses UUIDs to identify users as admins/blacklisted users. I'll file a ticket about that.

PR #543, opened 2018-03-13
Add a CRAN backend.

### LINKED ISSUE #510: CRAN backend
Many R packages are published on CRAN. It would be useful to have a backend that works with it. An additional wrinkle is that R versions tend to include dashes like 1.2-3, but on say Fedora, that gets converted to 1.2.3. I'm not sure how it works out on other distros.

QuLogic:
Instead of trying to parse CRAN webpages, this uses MetaCRAN to determine existing versions of packages.

This also replaces dashes with dots in versions as that's what's needed in Fedora. I'm not sure if this is the correct place to be doing that change (maybe it should go in the-new-hotness somehow.) I also see an ecosystem thing and I'm not sure if this should be added there too.

Fixes #510.

### DIFF
--- a/anitya/lib/backends/cran.py
+++ b/anitya/lib/backends/cran.py
@@ -1,0 +1,175 @@
+# -*- coding: utf-8 -*-
+# This file is part of the Anitya project.
+# Copyright (C) 2018 Elliott Sales de Andrade <quantum.analyst@gmail.com>
+#
+# This program is free software; you can redistribute it and/or modify
+# it under the terms of the GNU General Public License as published by
+# the Free Software Foundation; either version 2 of the License, or
+# (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program; if not, write to the Free Software
+# Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301 USA
+
+"""
+A backend for the R package host, `CRAN <https://cran.r-project.org/>`_.
+
+CRAN does not provide a specific API, so this uses the `MetaCRAN <https://www.r-pkg.org/>`_
+JSON API instead:
+
+  * An HTTP GET request to ``https://crandb.r-pkg.org/{name}`` returns information about the `latest
+    version of the package <https://github.com/metacran/crandb#pkg-latest-version-of-a-package>`_.
+    We read the ``Version`` key from this response.
+  * An HTTP GET request to ``https://crandb.r-pkg.org/{name}/all`` returns information about `all
+    versions <https://github.com/metacran/crandb#pkgall-all-versions-of-a-package>`_. The
+    ``versions`` key in the response contains a dictionary keyed by version and with values
+    corresponding to the previous request above.
+  * An HTTP GET request to ``https://crandb.r-pkg.org/-/pkgreleases?descending=true&limit=50``
+    returns information on `all package releases
+    <https://github.com/metacran/crandb#-pkgreleases-package-releases>`_, in order of release.
+    The response is a list of dictionaries with the ``package`` key containing the same results as
+    the first request above.
+"""
+
+import json
+
+import requests
+
+from anitya.lib.backends import BaseBackend
+from anitya.lib.exceptions import AnityaPluginException
+
+
+class CranBackend(BaseBackend):
+    """The custom class for projects hosted on CRAN."""
+
+    name = 'CRAN (R)'
+    examples = [
+        'https://cran.r-project.org/web/packages/tidyverse/',
+        'https://cran.r-project.org/web/packages/devtools/',
+    ]
+
+    @classmethod
+    def get_version(cls, project):
+        """
+        Retrieve the latest version of the provided project from this backend.
+
+        Args:
+            project (anitya.db.models.Project): The CRAN project to retrieve the version for.
+
+        Returns:
+            str: The latest version found upstream.
+
+        Raises:
+            AnityaPluginException: If the URL was unreachable or the response was in an unexpected
+                format.
+
+        """
+        url = 'https://crandb.r-pkg.org/{name}'.format(name=project.name)
+
+        try:
+            response = cls.call_url(url)
+        except requests.RequestException as e:  # pragma: no cover
+            raise AnityaPluginException('Could not contact {}: {}'.format(url, str(e)))
+
+        if response.status_code != 200:
+            raise AnityaPluginException(
+                'Failed to download from {}: {} {}'.format(url,
+                                                           response.status_code,
+                                                           response.reason))
+
+        try:
+            data = response.json()
+        except json.JSONDecodeError:  # pragma: no cover
+            raise AnityaPluginException('No JSON returned by {}'.format(url))
+
+        if 'error' in data or 'Version' not in data:  # pragma: no cover
+            raise AnityaPluginException('No versions found at {}'.format(url))
+
+        return data['Version']
+
+    @classmethod
+    def get_versions(cls, project):
+        """
+        Retrieve all the versions (that can be found) of the provided project from this backend.
+
+        Args:
+            project (anitya.db.models.Project): The project to retrieve the versions for.
+
+        Returns:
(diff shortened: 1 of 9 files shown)


anitya/lib/backends/cran.py
```diff
@@ -0,0 +1,120 @@
+# -*- coding: utf-8 -*-
+
+"""
+ (c) 2018 - Copyright Elliott Sales de Andrade
```
jeremycline: Once you move your copyright claim, can you use this docblock to provide an overview of the module? Just an introduction CRAN with pointers to more detailed documentation is fine. Something similar to https://github.com/release-monitoring/anitya/blob/e5e199a7b9d88df1465a8a1041140e93a3b3f2eb/anitya/lib/backends/crates.py#L19 if possible.

anitya/lib/backends/cran.py
```diff
@@ -0,0 +1,120 @@
...
+        'https://cran.r-project.org/web/packages/tidyverse/',
+        'https://cran.r-project.org/web/packages/devtools/',
+    ]
+
+    @classmethod
+    def get_version(cls, project):
+        ''' Method called to retrieve the latest version of the projects
+        provided, project that relies on the backend of this plugin.
+
+        :arg Project project: a :class:`anitya.db.models.Project` object whose backend
```
jeremycline: I've been using the ["google style" docbocks](http://www.sphinx-doc.org/en/1.6.3/ext/example_google.html) since I think they look a little nicer in the code. It's not a huge deal since the code base isn't consistent at the moment, but it's what I prefer.

anitya/lib/backends/cran.py
```diff
@@ -0,0 +1,120 @@
...
+        :raise AnityaPluginException: a
+            :class:`anitya.lib.exceptions.AnityaPluginException` exception
+            when the version cannot be retrieved correctly
+
+        '''
+        url = 'https://crandb.r-pkg.org/{name}'.format(name=project.name)
+
+        try:
+            response = cls.call_url(url)
+        except Exception:  # pragma: no cover
```
jeremycline: Since this is always an https url you can just except ``requests.RequestException`` here. It'd also be good to include the exception's message in the AnityaPluginException message. Dropping the ``# pragma: no cover`` and adding a test would be good, too.

anitya/lib/backends/cran.py
```diff
@@ -0,0 +1,120 @@
...
+            raise AnityaPluginException('Could not contact %s' % url)
+
+        try:
+            data = response.json()
+        except Exception:  # pragma: no cover
```
jeremycline: It'd be good to catch the specific exception here, which should be ``ValueError`` (or a subclass of that in Python 3).

anitya/lib/backends/cran.py
```diff
@@ -0,0 +1,120 @@
...
+            raise AnityaPluginException('No JSON returned by %s' % url)
```
jeremycline: Maybe add the ``response.status_code`` to the message.

anitya/lib/backends/cran.py
```diff
@@ -0,0 +1,120 @@
...
+
+        if 'error' in data or 'Version' not in data:
```
jeremycline: Instead of checking for the 'error' key (which appears to be in the non-200 responses), it might be better to check the status code earlier and report that.

anitya/lib/backends/cran.py
```diff
@@ -0,0 +1,120 @@
...
+        :raise AnityaPluginException: a
+            :class:`anitya.lib.exceptions.AnityaPluginException` exception
+            when the versions cannot be retrieved correctly
+
+        '''
+        url = 'https://crandb.r-pkg.org/{name}/all'.format(name=project.name)
+
+        try:
+            response = cls.call_url(url)
+        except Exception:  # pragma: no cover
```
jeremycline: Same comment as above for the error handling here

anitya/lib/backends/cran.py
```diff
@@ -0,0 +1,120 @@
...
+        ''' Return a generator over the latest 50 uploads to CRAN
+
+        by querying the MetaCRAN JSON API.
+        '''
+
+        url = 'https://crandb.r-pkg.org/-/pkgreleases?descending=true&limit=50'
+
+        try:
+            response = cls.call_url(url)
+        except Exception:  # pragma: no cover
```
jeremycline: Same error handling deal here :)

anitya/tests/lib/backends/test_cran.py
```diff
@@ -0,0 +1,115 @@
...
+# of the GNU General Public License v.2, or (at your option) any later
+# version.  This program is distributed in the hope that it will be
+# useful, but WITHOUT ANY WARRANTY expressed or implied, including the
+# implied warranties of MERCHANTABILITY or FITNESS FOR A PARTICULAR
+# PURPOSE.  See the GNU General Public License for more details.  You
+# should have received a copy of the GNU General Public License along
+# with this program; if not, write to the Free Software Foundation,
+# Inc., 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301, USA.
+#
+# Any Red Hat trademarks that are incorporated in the source
```
jeremycline: Since there aren't any Red Hat trademarks in this code I think you can safely drop this

anitya/tests/lib/backends/test_cran.py
```diff
@@ -0,0 +1,115 @@
...
+            'https://github.com/rmendels/xtractomatic',
+            'CRAN (R)', '3.4.2'))
+        self.assertEqual(items[1], (
+            'FedData',
+            'https://github.com/ropensci/FedData',
+            'CRAN (R)', '2.5.2'))
+        # etc...
+
+
+if __name__ == '__main__':
```
jeremycline: Could you remove this? I've been meaning to remove them from all the existing tests since we use pytest as the test runner.

anitya/tests/lib/backends/test_cran.py
```diff
@@ -0,0 +1,115 @@
...
+
+    def test_cran_get_version(self):
+        """ Test the get_version function of the CRAN backend. """
+        pid = 1
+        project = models.Project.get(self.session, pid)
+        exp = '0.3.2'
+        obs = backend.CranBackend.get_version(project)
+        self.assertEqual(obs, exp)
+
+        pid = 2
```
jeremycline: I know you've following existing tests, but unfortunately they're not great and should be refactored, sorry!

Tests should only test one thing, so this should be broken into two tests, the first checking the successful path, and the second checking that the exception is raised. I did some refactoring like that in https://github.com/release-monitoring/anitya/blob/535283835a1edae820470579cd7e55e76246cef7/anitya/tests/lib/backends/test_pypi.py, but I haven't fixed all the tests yet.

anitya/tests/lib/backends/test_cran.py
```diff
@@ -0,0 +1,115 @@
...
+
+    def test_cran_get_versions(self):
+        """ Test the get_versions function of the CRAN backend. """
+        pid = 1
+        project = models.Project.get(self.session, pid)
+        exp = ['0.1', '0.3.2']
+        obs = backend.CranBackend.get_ordered_versions(project)
+        self.assertEqual(obs, exp)
+
+        pid = 2
```
jeremycline: If you could also split this test up that'd be great, thanks!

QuLogic:
Fixed everything except the `no cover` lines. I'm not sure how to get those tested without the server returning some garbage.

anitya/lib/backends/cran.py
```diff
@@ -0,0 +1,182 @@
...
+
+import json
+
+import requests
+
+from anitya.lib.backends import BaseBackend
+from anitya.lib.exceptions import AnityaPluginException
+
+
+def _normalize_version(version):
```
jeremycline: Sorry, I meant to comment on this first time round, but I think you're right that this should go in the-new-hotness. Anitya shouldn't enforce Fedora-specific version stuff.

QuLogic: OK, removed.

QuLogic:
Any idea when this might get in a release? It's been over a year since the last one...

jeremycline:
Unfortunately I can't promise any real timeline. I asked from resources from Fedora's infrastructure team which hosts release-monitoring.org for a staging instance since I don't want to deploy directly to production, but it's been 4 months and there's been no movement. 

It might just be better to deploy it straight to production and deal with the fallout, but the host still needs to get rebased to Fedora 27 or I need to do a lot of packaging work. Sorry :disappointed:

Issue #533, opened 2018-03-09
query for all projects retrieves 2 projects with same name in different order


ricardo-martincoski:
Hello,

I am reviewing a local script to be used by Buildroot distro to generate a table listing projects, current version in Buildroot, and latest available version based on release-monitoring (and nice colors to mark projects that are up-to-date or not).
While we don't populate all the mappings for the distro, we would like to have guesses for projects without mapping.

Here is a small and clumsy script that demonstrates what I found:

```
import json
import os
import subprocess
def print_id(list):
    for i in list['projects']:
        if i['name'] == 'readline':
            print(i['id'])
API = "http://release-monitoring.org/api"
url1 = os.path.join(API, "projects", "?pattern=%s" % 'readline')
url2 = os.path.join(API, "projects")
subprocess.call(['wget', '-O', 'readline', url1])
subprocess.call(['wget', '-O', 'projects', url2])
print_id(json.load(open('readline')))
print_id(json.load(open('projects')))

```
Sometimes it gives me:
913
4173
4173
913
and sometimes it gives me:
913
4173
913
4173
It seems to not always return the same order. If I run the above script few times in a row the order is consistent. If I wait few hours the result can change.

The same occurs for zlib.

ricardo-martincoski:
I don't really develop using flask, so forgive me if my suggestion is stupid.
Couldn't a more consistent result be achieved with a simple change like this?

```
anitya/db/models.py @ Project.all(...)
         ).order_by(
             sa.func.lower(Project.name)
+        ).order_by(
+            Project.id
         )

```
Unfortunately I am using an old distro (Ubuntu 16.04) so the setup to test this code was not quick (missing vagrant extensions).
I plan to upgrade my distro but it takes time, so it's unlikely I can try to contribute code any time soon.

jeremycline:
Hi @ricardo-martincoski, thanks for the bug report.

Yes, ordering by id in addition to the name should produce consistent results. We're trying to retire the v1 API, though, since it has a lot of other problems (lots of endpoints don't paginate, for example). The new API doesn't appear to be ordering by anything beyond name (https://github.com/release-monitoring/anitya/blob/a0a9343dd86d9b7d41487b241f8e102d3321deee/anitya/api_v2.py#L328) so it'd probably be a good time to change that. It could order on ecosystem rather than project ID since projects should now have that set.

[...]

PR #546, opened 2018-03-18
Issue #533: sort all projects by (name, ecosystem)

ricardo-martincoski:
The list of all projects is already sorted by the project name. But it
results in inconsistent order over long periods of time (~8 hours) as
there is no explicit rule to sort two projects with the same name.
Two examples of multiple projects with the same name are readline and
zlib.

For api v2, add a second sort rule, based on the ecosystem to make the
results predictable.

Signed-off-by: Ricardo Martincoski <ricardo.martincoski@gmail.com>

### DIFF
--- a/anitya/tests/test_flask_api_v2.py
+++ b/anitya/tests/test_flask_api_v2.py
@@ -587,6 +587,66 @@ class ProjectsResourceGetTests(DatabaseTestCase):
 
         self.assertEqual(data, exp)
 
+    def test_list_projects_with_same_name(self):
+        """Assert two projects with the same name are sorted by the ecosystem."""
+        self.maxDiff = None
+        session = Session()
+        project1 = models.Project(
+            name="zlib",
+            homepage="https://hackage.haskell.org/package/zlib",
+            backend='GitHub',
+            ecosystem_name="https://hackage.haskell.org/package/zlib",
+        )
+        project2 = models.Project(
+            name="zlib",
+            homepage="http://www.zlib.net/",
+            backend="custom",
+            regex="DEFAULT",
+            ecosystem_name="http://www.zlib.net/",
+        )
+        session.add_all([project1, project2])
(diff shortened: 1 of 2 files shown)


anitya/tests/test_flask_api_v2.py
```diff
@@ -587,6 +587,66 @@ def test_list_projects(self):
...
+                    "regex": None,
+                    "version": None,
+                    "version_url": None,
+                    "versions": [],
+                    "ecosystem": "https://hackage.haskell.org/package/zlib",
+                }
+            ]
+        }
+
+        self.maxDiff = None
```
pypingou: You probably want to set this elsewhere :)

jeremycline: @pypingou Did you have a particular place in mind? I don't really mind it where it is. Maybe at the top of the test would be nicer, but that's a very minor nitpick.

ricardo-martincoski: Thank you both for your review.

Indeed there are other test cases in the same file using it at the top of the test:
https://github.com/release-monitoring/anitya/blob/3433f08242d2ed542f9ba0d61614659dc8152a9d/anitya/tests/test_flask_api_v2.py#L357-L359

I will do the same.

I will also move `session = Session()` to be consistent with
https://github.com/release-monitoring/anitya/blob/3433f08242d2ed542f9ba0d61614659dc8152a9d/anitya/tests/test_flask_api_v2.py#L512-L516

pypingou: I'd even put it at the top of the class since it may influence other tests run after this one

jeremycline: I'm not sure what you mean here, but if you're concerned setting it in an individual test affects any test run afterwards, that doesn't seem to be the case. That would be a very weird behavior.

PR #556, opened 2018-05-15
Link to backend info from project edit form

abitrolly:


nqb:
Not agree with this commit. 

`<a href="https://release-monitoring.org/about#backends">` : will work only for release-monitoring.org instance. 

Please use relative path.

abitrolly:
You mean absolute to the root of the web-site? Because I don't remember URL for this form to calculate relative path.

If the path is relative to web-site root, then can you point me to documentation to find the helper for this site root, because I am not familiar with [this framework](https://github.com/release-monitoring/anitya/blob/master/requirements.txt)?

abitrolly:
@nqb I just copy/pasted what seems to be logical for me. Should work.

nqb:
Hello @abitrolly,

Seems better but I can't test. @jeremycline, your opinion ?

abitrolly:
> Does this fix an open issue on the issue tracker?

No, only self.

abitrolly:
@jeremycline so did you get the time to review this?

jeremycline:
Sorry, it dropped off my radar. I just gave this a test, it looks like it's not accounting for the documentation having moved away from ``/about``. The new href would be ``href="{{ url_for('anitya_ui.static', filename='docs/user-guide.html') }}#backends"``, but perhaps it would be better to just link to the user guide as a whole from the new project page (in other words, use the ``#creating-a-new-project`` fragment and put it at the top somewhere as "Help" or something). Having just the backends linked is a bit odd when there's a whole guide.

abitrolly:
The link is for helping users quickly choose proper backend. User guide is already available from the top menu.

jeremycline:
What I'm suggesting is that if someone doesn't know what backend they want, they probably want all the documentation on creating a new project. The documentation generally is linked at the top, but linking directly to the "create a project" documentation is more helpful. If just the backend field is linked, I (as a user) will wonder why the rest of the fields aren't linked or don't have documentation.

The adjustment you've made doesn't look like it'll render, however. You've got ``}}#backends }}#backends``

abitrolly:
If fixed the link, but until 0.12 is released https://github.com/release-monitoring/anitya/issues/561 there it not much motivation to add other documentation links. I don't like layout of Sphinx docs and they are hard to customize, so nobody does it.

Zlopez:
Now when there is version 0.13.1 live I want to add this PR.

Just one note, this change is made in `project.html` template, but according to what you wrote, you want this link to be visible when the user is creating new project.
So you should do the change in `project_new.html` template, but I left this up to you.

abitrolly:
Rebased. Making a link for edit/new form is doable. I can do this right away. Without testing as I am not sure I have time to setup testing environment. https://github.com/release-monitoring/anitya/blob/master/anitya/templates/functions.html

Zlopez:
To test it in current state, you need to delete `check_user` event in `anitya/db/events.py`.
This will be fixed by #669

abitrolly:
Just commenting decorator helped. PR works ok.

abitrolly:
Squashed commits.

Zlopez:
Looks good. Could you just add news file for townscryer? Just add a new file to `news` folder.

abitrolly:
It is possible to autogenerate links for adding `towncrier` files. Like https://github.com/abitrolly/anitya/new/patch-1?filename=news/556.feature

Zlopez:
Link isn't working

abitrolly:
Because I removed my branch.

PR #570, opened 2018-08-01
WIP: Initial message schema for Anitya

jeremycline:
This defines message schema using json-schema for a subset of messages
sent by Anitya.

Signed-off-by: Jeremy Cline <jcline@redhat.com>


Zlopez:
As [Fedora-messaging](https://github.com/fedora-infra/fedora-messaging) is released with version 1.0.0, this could be a good addition to the project.

@jeremycline 
Do you need any help with this PR?

jeremycline:
I'm going to take a bit of time today or tomorrow to clean it up, it'll be a good first app to move over

jeremycline:
Yup, I'm not done with this yet, sorry for the noise

jeremycline:
This depends on #635 which I think depended on #633. Tests are failing, but a lot of them have to do with the logging code that's about to be removed, so I don't want to spend time fixing them just to remove them.

Zlopez:
Ok, I will than fix #635 first than.

anitya_schema/anitya_schema/project_messages.py
```diff
@@ -0,0 +1,544 @@
...
+
+class ProjectMapDeleted(message.Message):
+    topic = "org.release-monitoring.prod.anitya.project.map.remove"
+    body_schema = {
+        "id": "https://fedoraproject.org/jsonschema/anitya.project.map.deleted.json",
+        "$schema": "http://json-schema.org/draft-04/schema#",
+        "type": "object",
+        "required": ["project", "message", "distro"],
+        "properties": {
+            "distro": {"type": "null"},
```
Zlopez: Why are we sending this in project related messages?

anitya_schema/anitya_schema/project_messages.py
```diff
@@ -0,0 +1,544 @@
...
+            "message": {
+                "type": "object",
+                "properties": {
+                    "agent": {"type": "string"},
+                    "odd_change": {"type": "boolean"},
```
Zlopez: This is no longer happening. I removed this error when I introduced check for multiple versions.

anitya_schema/anitya_schema/tests/__init__.py
```diff
@@ -0,0 +1,20 @@
+# Copyright (C) 2018  Red Hat, Inc.
```
Zlopez: Fixtures are really nice solution for testing.

anitya_schema/anitya_schema/tests/test_schema.py
```diff
@@ -0,0 +1,45 @@
...
+
+
+class ExistingMessagesTests(unittest.TestCase):
+    """
+    Assert schema successfully handle legacy fedmsg messages.
+
+    The fixtures are generated by downloading messages from datagrepper.
+    """
+
+    def test_valid(self):
```
Zlopez: I don't see any assert in this method.

anitya_schema/setup.py
```diff
@@ -0,0 +1,73 @@
...
+from setuptools import setup, find_packages
+
+
+here = os.path.abspath(os.path.dirname(__file__))
+with open(os.path.join(here, "README.rst")) as fd:
+    README = fd.read()
+
+
+setup(
+    name="anitya_schema",
```
Zlopez: I expect that this should be the package I have to import in the-new-hotness.

jeremycline: Correct, it'll be published to PyPI/packaged separately from anitya so consumers can just pull in the schema

anitya_schema/anitya_schema/tests/test_schema.py
jeremycline: Yep, that's expected, although I should probably add a comment to that effect. ``validate()`` raises an exception if the message is invalid which would cause the test to fail. The assertion is a basically there there isn't an exception.

anitya_schema/anitya_schema/project_messages.py
jeremycline: So these schema aren't what we *should* be sending, it's just what the current messages look like. Most of the message schema are crazy and make no sense. This is more about documenting the madness so we can fix it than it is about laying out a reasonable set of messages. Once we've done that you can come up with reasonable messages and a migration plan for consumers.

anitya_schema/anitya_schema/project_messages.py
jeremycline: That's not good :(

The message schema changed and it's possible consumers were relying on that field. If, for example, the-new-hotness expects this key to be present it will crash. The best path forward is likely to make sure that key remains the message getting sent out and is always false or something.

anitya/lib/utilities.py
```diff
@@ -38,19 +40,29 @@
 
 def fedmsg_publish(*args, **kwargs):  # pragma: no cover
```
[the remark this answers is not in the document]
jeremycline: This is a interim solution. The most reasonable refactor is likely to get rid of this function entirely. The reason is that different messages almost certainly have different failure scenarios and once you get rid of the fedmsg portion of this function, all it's doing is squashing exceptions.

anitya_schema/anitya_schema/project_messages.py
Zlopez: Oh, I didn't knew this could have impact like this. The change is only on master right now. It wasn't released yet, so I can add the field back.
It's hard to know what is required.

Zlopez:
@jeremycline All the blocking PRs should be merged now, so you can continue on this PR when you have time.

anitya_schema/anitya_schema/project_messages.py
Zlopez: The odd_change is now always false.

Zlopez:
I'm merging this and continue working where @jeremycline ended.

Issue #581, opened 2018-08-17
GitHub backend does not see any tags


AloisMahdal:
I just logged in to  https://release-monitoring.org/ (anitya v0.11, at this moment) and added 2 GitHub projects: ymattw/ydiff and naelstrof/slop.  I tried to test them and both end with error:

    From https://release-monitoring.org/:
    Unable to retrieve the latest version from upstream! ERROR: ydiff: no upstream version found. - https://github.com/ymattw/ydiff/tags - class="tag-name">([^<]*)</span

It looks like the error contains pieces of HTML regex; however the regex cannot match; this is how HTML looks now:

    [...]
    <h4 class="flex-auto min-width-0 pr-2 pb-1 commit-title">
      <a href="/ymattw/ydiff/releases/tag/1.1">
        1.1
      </a>
      <span class="hidden-text-expander inline"><button type="button" aria-expanded="false" class="ellipsis-expander js-details-target" aria-label="Toggle commit message">&hellip;</button></span>
    </h4>
    [...]

As I mentioned, the site uses old anitya version, but I haven't found this issue in release notes nor in Issues, so this is probably broken with v0.12 as well.

AloisMahdal:
Needless to say, matching against HTML is rather questionable; especially in case of GitHub it looks like Anitya could do way better by just querying the git repos directly.

Heidistein:
This is also breaking other existing packages (ie phpMyAdmin)

Zlopez:
I'm already started working on implementation of Github API. Hopefully, it will be delivered in next version.

cheese1:
+1

Zlopez:
Github API was already merged. Closing this issue.

abitrolly:
@Zlopez let's close the issue when it is released, because it is not fixed for users who report the problem.

abitrolly:
The error is caused by changed GitHub markup. I can not find `tag-name` in current version of tags page markup - http://web.archive.org/web/20180901082326/https://github.com/sdispater/poetry/tags

In case GitHub API approach fails because of stricter rate limits, just updating the filter may fix the problem.

Zlopez:
Reopening because the new version is still not released.

Zlopez:
Closing after release of 0.13.1 in production.

Issue #567, opened 2018-06-25
Git(Hub|Lab) APIs and GNOME


yarons:
I looked at the current plugin and maybe fetching the entire HTML is a little counter intuitive.

With the latest GitHub GraphQL I can fetch the latest tag like this:

```
query {
  repository(owner: "release-monitoring", name: "anitya") {
    tags: refs(refPrefix: "refs/tags/", first: 1, direction: DESC) {
      edges {
        tag: node {
          name
        }
      }
    }
  }
}
```
Which is very efficient but requires a user.
With v3 it's possible to fetch all tags like this:
https://api.github.com/repos/release-monitoring/anitya/tags

It works pretty similar with GitLab:
https://gitlab.gnome.org/api/v4/projects/GNOME%2Fgnome-video-arcade/repository/tags

GNOME recently switch to GitLab to it's better to remove that in favor of a native GitLab plugin as well (Or just redirect it properly).

jeremycline:
Indeed, and it'd be much better to use APIs. https://github.com/release-monitoring/anitya/pull/502 started on an implementation using the v3 (I think) API. I think someone could pick that PR up and make a few changes and it'd be good to go. Same deal with GitLab.

Zlopez:
I will start on implementation of GitHub API [v4](https://developer.github.com/v4/) instead of using the older v3.

I see that the GitLab API is now only supporting REST and the GitHub API is GraphQL. So this looks like this will be two different APIs to implement.

Zlopez:
Github and Gitlab API were already merged. Closing this issue.

yarons:
So now we'll just have to wait for the production version upgrade?

Zlopez:
I need to release new version first and deploy it in staging.

But yes, we need to wait till upgrade on production.

yarons:
Is it in 0.12.1?

Zlopez:
Just now only in master, it will be released in 0.13.0.

yarons:
Regardless, thank you so much for this, keep up the great work!

PR #582, opened 2018-08-21
Update Github backend to use Github API v4

Zlopez:
Implementation of [Github API v4](https://developer.github.com/v4/) as requested in https://github.com/release-monitoring/anitya/issues/567
This should also solve https://github.com/release-monitoring/anitya/issues/581

### DIFF
--- a/Vagrantfile.example
+++ b/Vagrantfile.example
@@ -4,7 +4,7 @@
 VAGRANTFILE_API_VERSION = "2"
 
 Vagrant.configure(VAGRANTFILE_API_VERSION) do |config|
- config.vm.box = "fedora/27-cloud-base"
+ config.vm.box = "fedora/28-cloud-base"
 
  # Forward traffic on the host to the development server on the guest
  config.vm.network "forwarded_port", guest: 5000, host: 5000
--- a/Vagrantfile.example
+++ b/Vagrantfile.example
@@ -4,7 +4,7 @@
 VAGRANTFILE_API_VERSION = "2"
 
 Vagrant.configure(VAGRANTFILE_API_VERSION) do |config|
- config.vm.box = "fedora/28-cloud-base"
+ config.vm.box = "fedora/27-cloud-base"
 
  # Forward traffic on the host to the development server on the guest
  config.vm.network "forwarded_port", guest: 5000, host: 5000
--- a/anitya/lib/backends/__init__.py
+++ b/anitya/lib/backends/__init__.py
@@ -199,13 +199,13 @@ class BaseBackend(object):
 
         :arg url: the url to request (get).
         :type url: str
+        :arg insecure: flag for secure/insecure connection (default False)
+        :type insecure: bool
+        :type use_token: bool
         :return: the request object corresponding to the request made
         :return type: Request
         '''
-        user_agent = 'Anitya %s at upstream-monitoring.org' % \
-            pkg_resources.get_distribution('anitya').version
-        from_email = anitya_config.get('ADMIN_EMAIL')
-
+        headers = self.get_header()
         if '*' in url:
             url = self.expand_subdirs(url)
 
@@ -213,19 +213,14 @@ class BaseBackend(object):
             socket.setdefaulttimeout(30)
 
             req = urllib2.Request(url)
-            req.add_header('User-Agent', user_agent)
-            req.add_header('From', from_email)
+            req.add_header('User-Agent', headers.user_agent)
+            req.add_header('From', headers.from_email)
             resp = urllib2.urlopen(req)
             content = resp.read()
 
             return content
 
         else:
-            headers = {
-                'User-Agent': user_agent,
-                'From': from_email,
-            }
-
             # Works around https://github.com/kennethreitz/requests/issues/2863
             # Currently, requests does not start new TCP connections based on
             # TLS settings. This means that if a connection is ever started to
@@ -246,6 +241,27 @@ class BaseBackend(object):
                     url, headers=headers, timeout=60, verify=True)
 
             return resp
+
+    @classmethod
+    def get_header(self):
+        ''' Get HTTP header.
+
+        This method returns predefined header for for Anitya.
+        The header will contain user-agent and e-mail.
+
+        :return: header dictionary
+        :return type: dict
+        '''
+        user_agent = 'Anitya %s at upstream-monitoring.org' % \
+            pkg_resources.get_distribution('anitya').version
+        from_email = anitya_config.get('ADMIN_EMAIL')
+
+        headers = {
+            'User-Agent': user_agent,
+            'From': from_email,
+        }
+
+        return headers
 
 
 def get_versions_by_regex(url, regex, project, insecure=False):
--- a/anitya/lib/backends/__init__.py
+++ b/anitya/lib/backends/__init__.py
@@ -36,6 +36,13 @@ import six
 
 REGEX = anitya_config['DEFAULT_REGEX']
 
+# Default headers for requests
+REQUEST_HEADERS = {
+    'User-Agent': 'Anitya %s at upstream-monitoring.org' %
+    pkg_resources.get_distribution('anitya').version,
+    'From': anitya_config.get('ADMIN_EMAIL'),
+    }
+
 _log = logging.getLogger(__name__)
 
 
@@ -207,11 +214,10 @@ class BaseBackend(object):
         :type url: str
         :arg insecure: flag for secure/insecure connection (default False)
         :type insecure: bool
-        :type use_token: bool
         :return: the request object corresponding to the request made
         :return type: Request
         '''
-        headers = self.get_header()
+        headers = REQUEST_HEADERS.copy()
         if '*' in url:
             url = self.expand_subdirs(url)
 
@@ -247,27 +253,6 @@ class BaseBackend(object):
                     url, headers=headers, timeout=60, verify=True)
 
             return resp
-
-    @classmethod
-    def get_header(self):
-        ''' Get HTTP header.
-
-        This method returns predefined header for for Anitya.
-        The header will contain user-agent and e-mail.
-
-        :return: header dictionary
-        :return type: dict
-        '''
-        user_agent = 'Anitya %s at upstream-monitoring.org' % \
-            pkg_resources.get_distribution('anitya').version
-        from_email = anitya_config.get('ADMIN_EMAIL')
-
-        headers = {
-            'User-Agent': user_agent,
-            'From': from_email,
-        }
-
-        return headers
 
 
 def get_versions_by_regex(url, regex, project, insecure=False):
--- a/anitya/lib/backends/__init__.py
+++ b/anitya/lib/backends/__init__.py
@@ -135,13 +135,17 @@ class BaseBackend(object):
         ''' Method called to retrieve the latest version of the projects
         provided, project that relies on the backend of this plugin.
 
-        :arg Project project: a :class:`anitya.db.models.Project` object whose backend
-            corresponds to the current plugin.
-        :return: the latest version found upstream
-        :return type: str
-        :raise AnityaPluginException: a
-            :class:`anitya.lib.exceptions.AnityaPluginException` exception
-            when the version cannot be retrieved correctly
+        Attributes:
+            project (:obj:`anitya.db.models.Project`): Project object whose backend
+                corresponds to the current plugin.
+
+        Returns:
+            str: Latest version found upstream
+
+        Raises:
+            AnityaPluginException: A
+                :obj:`anitya.lib.exceptions.AnityaPluginException` exception
+                when the versions cannot be retrieved correctly
 
         '''
         pass
@@ -152,13 +156,17 @@ class BaseBackend(object):
         of the projects provided, project that relies on the backend of
         this plugin.
 
-        :arg Project project: a :class:`anitya.db.models.Project` object whose backend
-            corresponds to the current plugin.
-        :return: a list of all the possible releases found
-        :return type: list
-        :raise AnityaPluginException: a
-            :class:`anitya.lib.exceptions.AnityaPluginException` exception
-            when the versions cannot be retrieved correctly
+        Attributes:
+            project (:obj:`anitya.db.models.Project`): Project object whose backend
+                corresponds to the current plugin.
+
+        Returns:
+            :obj:`list`: A list of all the possible releases found
+
+        Raises:
+            AnityaPluginException: A
+                :obj:`anitya.lib.exceptions.AnityaPluginException` exception
+                when the versions cannot be retrieved correctly
 
         '''
         pass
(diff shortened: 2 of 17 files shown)


Vagrantfile.example
```diff
@@ -4,7 +4,7 @@
 VAGRANTFILE_API_VERSION = "2"
 
 Vagrant.configure(VAGRANTFILE_API_VERSION) do |config|
- config.vm.box = "fedora/27-cloud-base"
+ config.vm.box = "fedora/28-cloud-base"
```
jeremycline: It's generally good practice to break this sort of stuff out into a separate commit. I don't see anyone wanting to revert this or anything, but it's a nice separation of concerns thing. Not a huge deal, don't feel like you need to fix it here.

anitya/lib/backends/__init__.py
```diff
@@ -199,33 +199,28 @@ def call_url(self, url, insecure=False):
...
             url = self.expand_subdirs(url)
 
         if url.startswith('ftp://') or url.startswith('ftps://'):
             socket.setdefaulttimeout(30)
 
             req = urllib2.Request(url)
-            req.add_header('User-Agent', user_agent)
-            req.add_header('From', from_email)
+            req.add_header('User-Agent', headers.user_agent)
+            req.add_header('From', headers.from_email)
```
jeremycline: ``headers`` is a dictionary, right? I don't think this works.

anitya/lib/backends/github.py
```diff
@@ -59,18 +65,168 @@ def get_versions(cls, project):
...
+                url = url[:-1]
         else:
             raise AnityaPluginException(
                 'Project %s was incorrectly set-up' % project.name)
 
-        return get_versions_by_regex(url, REGEX, project)
+        if url:
+            (owner, repo) = url.split('/')
+
+        if (not owner) or (not repo):
```
jeremycline: It's exactly equivalent, but ``if not (owner and repo)`` might read a little cleaner.

Also, I know this is just a refactor of what was there before, but it makes more sense to validate this *before* it goes in the database, rather than right here. It'd tell the user up-front they've put in incorrect data. Might be a nice refactor for a future PR.

anitya/lib/backends/github.py
```diff
@@ -59,18 +65,168 @@ def get_versions(cls, project):
...
+            elif version.startswith('v'):
+                version = version[1:]
+            # TODO: Save commit url to DB and make versions clickable on frontend
+            # commit_url = edge['node']['target']['commitUrl']
+            versions.append(version)
+
+        return versions
+
+    @classmethod
+    def prepare_query(cls, owner, repo, after=''):
```
jeremycline: There's no reason for this to be a class method. There's also no reason for the class inheritance structure in the backends generally, so while this matches the general style, it's not a good style. You can either make it a private instance method or just rip this out into a private module-level function.

anitya/lib/backends/github.py
```diff
@@ -59,18 +65,168 @@ def get_versions(cls, project):
...
+        ''' Method for preparing GraphQL query for specified repository
+
+        :arg owner: owner of the repository
```
jeremycline: I find the ["Google"](http://www.sphinx-doc.org/en/1.7/ext/example_google.html) style of comments more pleasant to write, you might as well.

anitya/lib/backends/github.py
```diff
@@ -59,18 +65,168 @@ def get_versions(cls, project):
...
+                remaining
+                resetAt
+            }
+        }
+        ''' % (owner, repo, after_str)
+
+        return query
+
+    @classmethod
+    def call_url_post(cls, url, json, insecure=False, use_token=False):
```
jeremycline: I think you can just factor this into the ``get_versions`` call. I think it's reasonable for the token to be required, and we should never use ``insecure=True`` with GitHub so it boils down to a single call to another function.

The ``get_headers`` interface (while matching the existing style) is more complicated than just setting a default set of headers somewhere, then making a copy here and injecting the access token. I recommend doing that.

anitya/lib/backends/github.py
```diff
@@ -59,18 +65,168 @@ def get_versions(cls, project):
...
+                    url, project.name, str(err)))
+
+        if resp.ok:
+            json = resp.json()
+        else:
+            raise AnityaPluginException(
+                '%s: Server responded with status "%s": "%s"' % (
+                    project.name, resp.status_code, resp.reason))
+
+        if 'errors' in json:
```
jeremycline: Consider making a function that accepts the json and spits out the versions (or raises an exception). That'll make testing easier, and make this function a digestible length.

anitya/lib/backends/github.py
```diff
@@ -59,18 +65,168 @@ def get_versions(cls, project):
...
+
+        total_count = json['data']['repository']['refs']['totalCount']
+
+        _log.debug('Received %s tags for %s' % (total_count, project.name))
+
+        # TODO: Save cursor to project and use it to fetch only newer versions
+        # cursor = json['data']['repository']['refs']['edges']['cursor']
+
+        remaining = json['data']['rateLimit']['remaining']
+        reset_time = json['data']['rateLimit']['resetAt']
```
jeremycline: We should handle the rate limiting gracefully. If we're getting rate limited, it should raise an exception (RateLimitError or similar) that contains the time the rate limit is over so that the caller can handle it by rescheduling for later (or exploding)

anitya/lib/backends/__init__.py
Zlopez: Yes, you are right. This means, that there was not test for this, I will add it.

anitya/lib/backends/github.py
Zlopez: Yes, this should be probably validated on frontend in javascript, but you also need to validate it before saving it to db (in case somebody is trying something nasty). This could be different for every backend, so it will be good to have validation method as part of backend and the db method could call it before inserting the data.

What do you think?

anitya/lib/backends/github.py
Zlopez: I will change it to private method. It doesn't need to be public.

anitya/lib/backends/github.py
Zlopez: I used the style, that was there before, but I will change it.

anitya/lib/backends/github.py
Zlopez: I didn't understood the second paragraph in your comment. You want me to copy the `get_headers` method to this backend.
The `get_headers` method is new method in `BaseBackend` and it only fills basic header fields.

anitya/lib/backends/github.py
Zlopez: Ok, I will do this. It sounds good.

anitya/lib/backends/github.py
Zlopez: This looks like elegant solution to handle the rate limit.

Vagrantfile.example
Zlopez: I will move it to separate commit.

anitya/lib/backends/github.py
jeremycline: Ah, what I'm saying is I don't like the use of inheritance in the backends. It's complicated and adds no value, but I never got a chance to refactor it. Rather than add to it, this could be implemented as:

In constants or BaseBackend's module or something:
```
REQUEST_HEADERS = <default_headers>
```

Then for your Github backend you just make a copy of REQUEST_HEADERS and add your auth token:

```
github_headers = REQUEST_HEADERS.copy()
github_headers['Authentication'] = ...
```

There's no need for a get_headers method that subclasses override. The whole ``call_url`` function has no business being a class method at all. If should just be a module-level function that acts as a wrapper that handles the special FTP case and otherwise is the requests API.

Zlopez: I thought that the inheritance is intentional, otherwise `BaseBackend` doesn't have much sense.

jeremycline: Inheritance is fine to define the interface, but the only interface we're really interested in is for a given project, new versions are returned. Everything else is an implementation detail of the particular backend and doesn't need to be part of the class.

anitya/lib/backends/github.py
```diff
@@ -50,27 +59,167 @@ def get_versions(cls, project):
...
+        except ValueError:
+            raise AnityaPluginException(
+                'Project %s was incorrectly set-up. \
+                Can\'t parse owner and repo' % project.name)
+
+        query = prepare_query(owner, repo)
+
+        try:
+            headers = REQUEST_HEADERS.copy()
+            token = config.get('GITHUB_ACCESS_TOKEN')
```
jeremycline: Since you set a default for the token, you don't need to use ``get``.

anitya/lib/exceptions.py
```diff
@@ -102,3 +102,22 @@ def __str__(self):
...
+    Raised when the rate limit for requests is reached.
+
+    Args:
+        reset_time (str): Time when limit will be reseted (UTC time encoded in ISO-8601).
+    """
+
+    def __init__(self, reset_time):
+        self._reset_time = reset_time
+
+    @property
```
jeremycline: I recommend just calling ``self._reset_time`` ``self.reset_time`` and not making a property. Properties are handy if you need to do some sort of transformation or you want your internal representation to differ from the public API in some way, but here it's the same.

anitya/tests/lib/backends/test_init.py
```diff
@@ -24,19 +24,22 @@
 from __future__ import absolute_import, unicode_literals
 
 import re
-import unittest
+import six.moves.urllib.request as urllib2
 
+import unittest
```
jeremycline: Following the PEP 8 rules, unittest should stay up in the top group and six should move to this group

anitya/tests/request-data/anitya.tests.lib.backends.test_init.BaseBackendTests.test_expand_subdirs
```diff
@@ -0,0 +1,74 @@
...
+        bWU9InEiIGlkPSJ0ZXh0LWYiIG1heGxlbmd0aD0iMjU1IiAvPgogICAgICAgICAgICA8aW5wdXQg
+        dHlwZT0ic3VibWl0IiB2YWx1ZT0iaGxlZGF0IiBpZD0ic3VibWl0LWYiIC8+CiAgICAgICAgICA8
+        L3A+CiAgICAgICAgPC9mb3JtPgotLT4KPC9ib2R5PjwvaHRtbD4K
+    headers:
+      connection: [Keep-Alive]
+      content-type: [text/html;charset=UTF-8]
+      date: ['Thu, 23 Aug 2018 09:39:54 GMT']
+      keep-alive: ['timeout=5, max=100']
+      server: [Apache/2.2.22 (Fedora)]
+      x-clacks-overhead: [GNU Terry Pratchett]
```
jeremycline: Nice

anitya/lib/backends/github.py
```diff
@@ -50,27 +59,167 @@ def get_versions(cls, project):
...
+                error['type'], error['message'])
+        raise AnityaPluginException(
+            '%s: Server responded with following errors\n%s' % (
+                project.name, error_str))
+
+    total_count = json['data']['repository']['refs']['totalCount']
+
+    _log.debug('Received %s tags for %s' % (total_count, project.name))
+
+    # TODO: Save cursor to project and use it to fetch only newer versions
```
jeremycline: I recommend turning these TODOs into tickets and removing them from the code

anitya/lib/backends/__init__.py
```diff
@@ -98,7 +105,13 @@ def expand_subdirs(self, url, glob_char="*"):
...
 
         if url_prefix != "":
-            dir_listing = self.call_url(url_prefix).text
+            resp = self.call_url(url_prefix)
+            # When FTP server is called, Response object is not created
+            # and we get binary string instead
+            try:
+                dir_listing = resp.text
+            except AttributeError:
+                dir_listing = resp.decode('utf-8')
```
jeremycline: I just skimmed https://tools.ietf.org/html/rfc2640, but it sounds like there's no guarantees that the response will be UTF-8 encoded, and even if the spec said it MUST be, I'd be willing to bet there are some non-conforming servers out there. If it *should* only be ASCII/UTF-8 then this is fine, but this should handle an Unicode decoding exception and raise the appropriate Anitya exception explaining that the FTP service is misbehaving.

anitya/lib/backends/github.py
Zlopez: Will do

anitya/tests/lib/backends/test_init.py
Zlopez: Didn't know this

anitya/lib/backends/github.py
Zlopez: You are right

anitya/lib/backends/__init__.py
Zlopez: I will add another try-except when trying to decode the binary string and if it's not UTF-8, I will just raise `AnityaPluginException`.
Maybe there is some way, how to get the encodings of the binary string and then use the appropriate decode method. This will be nice way how to do it.
I will check this.
It will be much easier, if `Requests` library could work with FTP.

anitya/tests/request-data/anitya.tests.lib.backends.test_init.BaseBackendTests.test_expand_subdirs
Zlopez: I was surprised by this value when I fetched the URL.

jeremycline: In case you're not familiar with Terry Pratchett's Discworld books, one of his books includes a lot of amusing references to the Free Software movement and this HTTP header is in [reference to one of his books](https://xclacksoverhead.org/home/about).

Zlopez: I know his work. I read almost all his books from Discworld, but I was surprised to see the header field, I never saw it before.

anitya/lib/backends/__init__.py
jeremycline: There is a requests-ftp extension, although I'm not sure if it works with current requests. It'd be nice to drop the FTP support, but I'm not sure if there's still a number of projects only available over FTP. I think just bailing is a reasonable approach. The specification indicates it should only ever be ASCII or UTF-8, but because FTP is so old I have no doubt there are servers out there not following the latest RFC about that.

Zlopez: I looked at it, but it's only a template, not real library. It's only some basic functionality for future work.

anitya/lib/exceptions.py
```diff
@@ -115,7 +115,6 @@ class RateLimitException(AnityaException):
     def __init__(self, reset_time):
         self._reset_time = reset_time
 
-    @property
     def reset_time(self):
```
jeremycline: Ah, what I meant was just delete this method and change the name from ``self._reset_time`` to ``self.reset_time``

anitya/lib/backends/__init__.py
```diff
@@ -197,35 +225,40 @@ def call_url(self, url, insecure=False):
...
+            try:
+                resp = urllib2.urlopen(req)
+                content = resp.read().decode()
+            except URLError as e:
+                raise AnityaPluginException(
+                        'Could not call "%s" with error: %s' % (
+                            url, e.reason))
+            except UnicodeDecodeError:
+                raise AnityaPluginException(
+                        'FTP response is not unicode: %s' % url)
```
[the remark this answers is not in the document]
Zlopez: Sounds better than current error message.

anitya/lib/exceptions.py
jeremycline: I wouldn't really call it a bad habit (this is the only way to do things in Java), but it's not "Pythonic". Python uses ``_`` it indicate a variable, method, function, etc is private and might change at any time, but doesn't enforce it. If the interface you'd like to present to the public differs from the one you'd like to manage internally, it's a perfect place for the setter/getter pattern with properties. Otherwise, just exposing the variable itself is fine.

PR #586, opened 2018-08-22
Update mergify configuration

Zlopez:


cverna:
Looks good but I don't know if you want to have the merge strategy set to rebase. That way you keep a linear history of the commit in the repo.

See https://github.com/fedora-infra/bodhi/blob/develop/.mergify.yml#L15

Zlopez:
I thought about it. Till now we used merge commit as merge strategy, which is default in mergify.
@jeremycline 
What do you think about the merge strategy?

jeremycline:
Rebase is prettier, and if a bot's doing it I think that's fine. I don't think the prettiness is worth asking people to deal with a rebase so I used merge unless I was doing my own patch from the CLI. Rebase is especially annoying when there's a number of open PRs and you review them all, but mergify makes that easy.

Zlopez:
Ok, I will change the merge method to rebase.

PR #597, opened 2018-08-29
Add basic release process guide

jeremycline:
These are the steps I follow when making a release.

### DIFF
--- a/docs/contributing.rst
+++ b/docs/contributing.rst
@@ -160,5 +160,26 @@ This will also show you all incoming messages from `libraries.io's <https://libr
 SSE feed.
 
 
+Release Guide
+=============
+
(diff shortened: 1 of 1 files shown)


Zlopez:
There is already basic release process described in https://pagure.io/infra-docs/blob/master/f/docs/sysadmin-guide/sops/anitya.rst
Could you update this also there?

jeremycline:
So that stuff is for deploying new upstream releases, and this is for making new releases in upstream. Do you want those docs pointing at these docs, or something else?

Zlopez:
I just want if you could review the https://pagure.io/infra-docs/blob/master/f/docs/sysadmin-guide/sops/anitya.rst and say if this is correct.
I did some changes to the SOP.

Zlopez:
Thanks

Issue #587, opened 2018-08-24
Google style link is redirecting to wrong url


Zlopez:
In the Development Guide the link to [Google Style](http://www.sphinx-doc.org/en/latest/ext/example_google.html) redirecting to page where the user doesn't have permissions.

This should redirect to some public page like: http://www.sphinx-doc.org/en/1.7/ext/example_google.html

Zlopez:
I'm updating documentation right now, so I will take this one.

PR #601, opened 2018-08-31
Enhance documentation

Zlopez:
This PR is updating Anitya documentation and adds tooltips to editable fields.

This PR also solves https://github.com/release-monitoring/anitya/issues/587

### DIFF
--- a/anitya/templates/project_new.html
+++ b/anitya/templates/project_new.html
@@ -129,7 +129,8 @@
         "Matched groups will be concatenated by '.'. " + 
         "For example regex '(\d)_(\d)_(\d)' applied to '1_2_3' " +
         "will convert version to '1.2.3'.");
-    $("#insecure").prop('title', "This should be set only when Version URL is HTTP");
+    $("#insecure").prop('title', "This should only be set when the version URL " +
+        "is HTTPS and the host does not have a trusted certificate");
     $("#check_release").prop('title', "Check the release immediately");
     {%- if context == 'Add' -%}
     $("#distro").prop('title', "Distribution where the project is delivered");
--- a/anitya/templates/project_new.html
+++ b/anitya/templates/project_new.html
@@ -118,7 +118,28 @@
       {%- endautoescape -%}
     {%- endfor %}
 
+  function add_tooltips(){
+    $("#name").prop('title', "Project upstream name");
+    $("#homepage").prop('title', "Project homepage");
+    $("#version_url").prop('title', "URL where to look for new releases");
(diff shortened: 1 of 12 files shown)


anitya/templates/project_new.html
```diff
@@ -118,8 +118,29 @@ <h1>{{ context }} project</h1>
...
+    $("#homepage").prop('title', "Project homepage");
+    $("#version_url").prop('title', "URL where to look for new releases");
+    $("#version_prefix").prop('title', "Prefix filled here will be removed " +
+        "from retrieved version. For example version prefix " +
+        "'v' will convert 'v1.2.3' to '1.2.3'.");
+    $("#regex").prop('title', "Python regex to use for version retrieval. " + 
+        "Matched groups will be concatenated by '.'. " + 
+        "For example regex '(\d)_(\d)_(\d)' applied to '1_2_3' " +
+        "will convert version to '1.2.3'.");
+    $("#insecure").prop('title', "This should be set only when Version URL is HTTP");
```
jeremycline: This should be "This should only be set when the version URL is HTTPS and the host does not have a trusted certificate"

Zlopez:
When this PR is still open I added a new feature to documentation. The database schema is now generated using [sphinx-sadisplay](https://github.com/sphinx-contrib/sadisplay) extension.

jeremycline:
Generally it's best to have a separate PR for additional features. I built the docs locally and it doesn't look like the image in rendering inline (it's a link).

We used a different library for Bodhi, so ideally we can use the same one for both (this or the existing one which is packaged for Fedora)

Zlopez:
It is part of the documentation enhancement, so I think this is good place where to add it.

I will look at the Bodhi to see, how is this generated there.

Zlopez:
I changed the database schema to be generated same as in Bodhi. It looks better now.

Zlopez:
Weird, I tried to run the travis-ci locally using docker image - `travisci/ci-garnet:packer-1512502276-986baf0` and everything works, I don't know why the automatic build is failing.

jeremycline:
> It is part of the documentation enhancement, so I think this is good place where to add it.

A PR should be a narrow change set. Typically, they're one commit unless there's a set of related changes that all need to land at once. Once one has been submitted and initially reviewed, it's best not to add to it. When I come back for a second review, my expectation is the the only changes will be in response to my comments. When new things show up along with those changes, it makes it difficult to review.

Zlopez:
Sorry for this. I could remove the part with generation from this PR and move it to another PR, if this is more acceptable.

Zlopez:
Removed the last commit. I will create a new PR from it.

PR #603, opened 2018-09-06
Add database schema generation

Zlopez:
Moved to separate PR from https://github.com/release-monitoring/anitya/pull/601 as new feature.

### DIFF
--- a/ansible/roles/anitya-dev/tasks/main.yml
+++ b/ansible/roles/anitya-dev/tasks/main.yml
@@ -19,6 +19,7 @@
     - python3-devel
     - redhat-rpm-config
     - rpm-python
+    - python2-sqlalchemy_schemadisplay
 
 # Add various helpful configuration files
 - name: Install a custom bashrc
--- a/docs/generate_db_schema
+++ b/docs/generate_db_schema
@@ -1,0 +1,22 @@
+#!/usr/bin/env python
+"""
+Generate an image of the database schema
+"""
+from sqlalchemy_schemadisplay import create_schema_graph
+
+from anitya.db import meta
+
+
+def write_graph(filename):
+   graph = create_schema_graph(
+       metadata=meta.metadata,
+       show_datatypes=False,
+       show_indexes=False,
+       rankdir='LR',
+       concentrate=False
+   )
+   graph.write_png(filename)
+
+
+if __name__ == '__main__':
+    write_graph('images/database.png')
--- a/.gitignore
+++ b/.gitignore
@@ -2,6 +2,8 @@
 build/
 dist/
 docs/_build/
+docs/images/
+docs/docblocks/
 *egg-info/
 *eggs/
 *.cfg
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,8 +1,9 @@
-sudo: false
+sudo: true
 language: python
 install:
   - pip install -U setuptools pip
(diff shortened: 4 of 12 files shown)


docs/generate_db_schema
```diff
@@ -0,0 +1,22 @@
+#!/usr/bin/env python
```
bowlofeggs: Note that [Fedora's Python packaging guidelines](https://fedoraproject.org/wiki/Packaging:Python#Multiple_Python_Runtimes) will require this to be changed to an explicit Python version (and not to use env too), like ```/usr/bin/python3```. It is possible to use ```sed``` in the specfile to accomplish this, but I recommend using an explicit shebang here instead.

docs/generate_db_schema
```diff
@@ -0,0 +1,22 @@
...
+"""
+Generate an image of the database schema
+"""
+from sqlalchemy_schemadisplay import create_schema_graph
+
+from anitya.db import meta
+
+
+def write_graph(filename):
```
bowlofeggs: It'd be good to add a docblock here, including the type of the ```filename``` parameter.

ansible/roles/anitya-dev/tasks/main.yml
```diff
@@ -19,6 +19,7 @@
     - python3-devel
     - redhat-rpm-config
     - rpm-python
+    - python2-sqlalchemy_schemadisplay
```
bowlofeggs: Is it possible to use the ```python3``` package instead, or is Anitya not yet Python 3 ready? Python 2 is deprecated in Fedora 30 and some packages are getting dropped there (not sure about schemadisplay).

bowlofeggs:
heh, I forgot about Mergify. Well if you make another PR in light of my recommendation, let me know and I'll review ☺

docs/generate_db_schema
Zlopez: Anitya isn't yet packaged in Fedora, but thanks for pointing at this.

docs/generate_db_schema
Zlopez: I copied this from Bodhi, changed only what is imported. :-)

ansible/roles/anitya-dev/tasks/main.yml
Zlopez: I will look if there is `python3` package for this library.

Zlopez:
Yes, I will do. Thanks for review.

ansible/roles/anitya-dev/tasks/main.yml
Zlopez: Unfortunately there isn't `python3` package in Fedora 28.
However I see that the package maitainer is @jeremycline - Do you plan python3 package?

docs/generate_db_schema
Zlopez: Tried with `/usr/bin/python3` and this isn't working. Probably because of the `sqlalchemy_schemadisplay` is python2 import.
However it's working with `/usr/bin/env python3`.

ansible/roles/anitya-dev/tasks/main.yml
jeremycline: I can look into it. Not sure why I didn't do a Python 3 package to begin with... Maybe it didn't support Python 3? I guess I'll find out

PR #607, opened 2018-09-14
Small tweaks in generate_db_schema script

Zlopez:
Changes according to comments in https://github.com/release-monitoring/anitya/pull/603

### DIFF
--- a/docs/generate_db_schema
+++ b/docs/generate_db_schema
@@ -1,4 +1,4 @@
-#!/usr/bin/env python
+#!/usr/bin/env python3
 """
 Generate an image of the database schema
 """
@@ -8,14 +8,21 @@ from anitya.db import meta
 
 
 def write_graph(filename):
-   graph = create_schema_graph(
+    """
+    Creates graph from database classes using
+    `sqlalchemy_schemadisplay` library.
+
+    Args:
+        filename (str): Output file where image will be written
+    """
+    graph = create_schema_graph(
        metadata=meta.metadata,
        show_datatypes=False,
        show_indexes=False,
        rankdir='LR',
        concentrate=False
-   )
-   graph.write_png(filename)
+    )
+    graph.write_png(filename)
 
 
 if __name__ == '__main__':
--- a/docs/generate_db_schema
+++ b/docs/generate_db_schema
@@ -1,4 +1,4 @@
-#!/usr/bin/env python
+#!/usr/bin/env python3
 """
 Generate an image of the database schema
 """
--- a/docs/generate_db_schema
+++ b/docs/generate_db_schema
@@ -1,4 +1,4 @@
-#!/usr/bin/env python3
+#!/usr/bin/env python
 """
 Generate an image of the database schema
 """
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,7 +14,6 @@ env:
     - PYTHONWARNINGS=always::DeprecationWarning
   matrix:
     - TOXENV=lint
-    - TOXENV=docs
     - TOXENV=py27
 matrix:
   include:
@@ -22,3 +21,5 @@ matrix:
       env: TOXENV=py35
     - python: 3.6
       env: TOXENV=py36
+    - python: 3.6
+      env: TOXENV=docs


docs/generate_db_schema
```diff
@@ -1,4 +1,4 @@
-#!/usr/bin/env python
+#!/usr/bin/env python3
```
bowlofeggs: I recommend the fully explicit ```/usr/bin/python3``` here since that's what the Fedora packaging guidelines require.

Zlopez: Even with `#!/usr/bin/env python3` the test is failing, because sqlalchemy_schemadisplay is only python 2. The only option how this could run is to use `#!/usr/bin/env python` or `#!/usr/bin/python`

bowlofeggs:
On 9/15/18 11:37 AM, Zlopez wrote:
> Even with |#!/usr/bin/env python3| the test is failing, because
> sqlalchemy_schemadisplay is only python 2. The only option how this
> could run is to use |#!/usr/bin/env python| or |#!/usr/bin/python|

In that case, I recommend /usr/bin/python2 (note that the Fedora
guidelines also do not allow /usr/bin/python).

Zlopez:
When using explicitly `#!/usr/bin/python2` I'm getting the same issue with anitya modules, because they are python3.

I tried:
* `#!/usr/bin/python2` - failing import anitya.db
* `#!/usr/bin/python3` - failing import sqlalchemy_schemadisplay
* `#!/usr/bin/env python2` - failing import anitya.db
* `#!/usr/bin/env python3` - failing import sqlalchemy_schemadisplay (I don't have this issue locally)
* `#!/usr/bin/python` - failing import anitya.db
* `#!/usr/bin/env python` - everything works

So I don't see any other option than using  `#!/usr/bin/env python` till the sqlalchemy_schemadisplay will be ported to python3.

jeremycline:
Since the docs aren't built in Fedora, where is this failing?

Note, I built a python 3 package of sqlalchemy_schemadisplay in rawhide. The tests are definitely not Python 3 compatible, but I tested it and it seems to work.

Zlopez:
It's failing when you run `tox -e docs` and the `generate_db_schema` script is executed.

Or you could activate docs environment `source .tox/docs/bin/activate` and try `./generate_db_schema`.

Zlopez:
To be honest I don't understand why this is happening, when I look in the `docs` tox environment I can see that there is sqlalchemy_schemadisplay in `.tox/docs/lib64/python3.6/site-packages/`.
But if I try to run `./generate_db_schema` with shebang `#!/usr/bin/python3` I get:
```
  File "./generate_db_schema", line 5, in <module>
    from sqlalchemy_schemadisplay import create_schema_graph
ModuleNotFoundError: No module named 'sqlalchemy_schemadisplay'
```

jeremycline:
That's because using #!/usr/bin/python3 forces it to not use the venv-set interpreter

Zlopez:
Got it, but why this isn't working even with `#!/usr/bin/env python3`?

jeremycline:
Is the docs venv being set up with Python 2? If so, that'd point to /usr/bin/python3 as well. That'd be what I'd check first.

Zlopez:
See above, the site-packages is `python 3.6`.

Zlopez:
Tried `python --version` in docs venv to be sure and got `python 3.6.6`.

I'm still not sure, why the `sqlalchemy_schemadisplay` can't be find by python.

Zlopez:
Ok, I found out, that the python venv created by travis is `python-2.7.14`. This explains why the python3 doesn't work.

I also tried to run it locally with `#!/usr/bin/python3` and this is using following sys.path `['/home/zlopez/git/anitya/docs', '/usr/lib64/python36.zip', '/usr/lib64/python3.6', '/usr/lib64/python3.6/lib-dynload', '/usr/lib64/python3.6/site-packages', '/usr/lib/python3.6/site-packages']`. So it's not using venv libraries at all.

The correct shebang should be `#!/usr/bin/env python3`, but this needs to update Travis configuration.

@jcline Why we are even using Travis container with Ubuntu 14.04.5 LTS?

Zlopez:
I updated `.travis.xml` to use `python 3.6` for docs venv and changed shebang to `#!/usr/bin/env python3`.

Issue #598, opened 2018-08-29
Can't rename mapping for gstreamer


Zlopez:
When trying to rename package mapping for [gstreamer](https://release-monitoring.org/project/1263/) on Fedora to gstreamer1 I'm getting following error:
```
Aug 29 15:04:31 anitya-dev.example.com flask[6718]: Traceback (most recent call last):
Aug 29 15:04:31 anitya-dev.example.com flask[6718]:   File "/home/vagrant/.virtualenvs/anitya/lib/python2.7/site-packages/flask/app.py", line 2309, in __call__                                                   
Aug 29 15:04:31 anitya-dev.example.com flask[6718]:     return self.wsgi_app(environ, start_response)
Aug 29 15:04:31 anitya-dev.example.com flask[6718]:   File "/home/vagrant/.virtualenvs/anitya/lib/python2.7/site-packages/flask/app.py", line 2295, in wsgi_app                                                   
Aug 29 15:04:31 anitya-dev.example.com flask[6718]:     response = self.handle_exception(e)
Aug 29 15:04:31 anitya-dev.example.com flask[6718]:   File "/home/vagrant/.virtualenvs/anitya/lib/python2.7/site-packages/flask_restful/__init__.py", line 273, in error_router                                   
Aug 29 15:04:31 anitya-dev.example.com flask[6718]:     return original_handler(e)
Aug 29 15:04:31 anitya-dev.example.com flask[6718]:   File "/home/vagrant/.virtualenvs/anitya/lib/python2.7/site-packages/flask/app.py", line 1741, in handle_exception                                           
Aug 29 15:04:31 anitya-dev.example.com flask[6718]:     reraise(exc_type, exc_value, tb)
Aug 29 15:04:31 anitya-dev.example.com flask[6718]:   File "/home/vagrant/.virtualenvs/anitya/lib/python2.7/site-packages/flask/app.py", line 2292, in wsgi_app                                                   
Aug 29 15:04:31 anitya-dev.example.com flask[6718]:     response = self.full_dispatch_request()
Aug 29 15:04:31 anitya-dev.example.com flask[6718]:   File "/home/vagrant/.virtualenvs/anitya/lib/python2.7/site-packages/flask/app.py", line 1815, in full_dispatch_request                                      
Aug 29 15:04:31 anitya-dev.example.com flask[6718]:     rv = self.handle_user_exception(e)
[...]

PR #608, opened 2018-09-14
Add delete cascade on DB models

Zlopez:
This should prevent orphaned DB objects.
For example packages without project or distribution.

Fixes https://github.com/release-monitoring/anitya/issues/598

### DIFF
--- a/anitya/db/models.py
+++ b/anitya/db/models.py
@@ -21,7 +21,7 @@
 try:
     # The Python 3.6+ API
     from secrets import choice as random_choice
-except ImportError:
+except ImportError:  # pragma: no cover
     # Fall back to random with os.urandom
     import random
     random = random.SystemRandom()
@@ -230,7 +230,7 @@ class Packages(Base):
     __tablename__ = 'packages'
 
     id = sa.Column(sa.Integer, primary_key=True)
-    distro = sa.Column(
+    distro_name = sa.Column(
         sa.String(200),
         sa.ForeignKey(
             "distros.name",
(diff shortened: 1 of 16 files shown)


anitya/db/models.py
```diff
@@ -656,7 +658,9 @@ class ProjectVersion(Base):
     )
     version = sa.Column(sa.String(50), primary_key=True)
 
-    project = sa.orm.relation('Project', backref='versions_obj')
+    project = sa.orm.relation(
```
cverna: Maybe you could use the relationship API here like in [here](https://github.com/release-monitoring/anitya/pull/608/files#diff-426a0c336a142390e6230960f40efb78R352)

anitya/db/models.py
```diff
@@ -679,7 +683,9 @@ class ProjectFlag(Base):
     updated_on = sa.Column(sa.DateTime, server_default=sa.func.now(),
                            onupdate=sa.func.current_timestamp())
 
-    project = sa.orm.relation('Project', backref='flags')
+    project = sa.orm.relation(
```
cverna: Same as above

anitya/db/models.py
Zlopez: `relation` is the synonym to `relationship`. See here http://docs.sqlalchemy.org/en/latest/orm/relationship_api.html

cverna: So why not keep consistency and use the same name either `relation` or `relationship` I think it would make the code easier to read.

Zlopez: As you can see I used the same method that was already there, but you are right I didn't noticed this.

cverna:
I would suggest to try to keep PRs shorter, this one was not very fun to review since it was a mixed of model change and test coverage increase. Also maybe more than one commit would have helped here :see_no_evil:

Zlopez:
I was actually thought about separating this in two PRs, but in the end I decided to go with one PR.
Next time I will do test coverage raising in separate PR.

Zlopez:
Resolved conflict.

Zlopez:
I decided to split this PR in two https://github.com/release-monitoring/anitya/pull/637 to make it more readable for reviewers. The second one contains only tests and should be merged before this one to make travis pass.

I also added news fragments to this PR.

Zlopez:
Handpicked changes from #637 so the tests could pass.

Zlopez:
Fixed conflicts and squashed commits.
I will merge this manually.

Zlopez:
Fixed failing tests, forgot to start `tox` after conflict resolving.

Issue #606, opened 2018-09-13
Source map error: request failed with status 404 for various javascript packages


Zlopez:
This error is happening in the Anitya 0.13.0.
```
Source map error: request failed with status 404
Resource URL: https://stg.release-monitoring.org/static/js/jquery-1.10.2.min.js
Source Map URL: jquery-1.10.2.min.map[Learn More]
```

PR #610, opened 2018-09-17
Update javascript libraries

Zlopez:
Update of javascript libraries.

This fixes https://github.com/release-monitoring/anitya/issues/606

There is one known issue in Firefox - `Error: Syntax error, unrecognized expression: [data-toggle]="tooltip"`. This is caused by tooltip handling by bootstrap. Originally we used jquery-ui tooltips, but they were causing the issue above.


Zlopez:
Anitya isn't Fedora specific. So it shouldn't have any Fedora RPM requirements.

cverna:
> Anitya isn't Fedora specific. So it shouldn't have any Fedora RPM requirements.

Ha yes I forgot about that :+1:

Zlopez:
Resolved conflict and added news file. Will merge this manually.

PR #613, opened 2018-09-20
Fix for cron issues

Zlopez:
This PR should fix current cron issues on staging.

### DIFF
--- a/anitya/lib/backends/github.py
+++ b/anitya/lib/backends/github.py
@@ -88,8 +88,8 @@ class GithubBackend(BaseBackend):
             (owner, repo) = url.split('/')
         except ValueError:
             raise AnityaPluginException(
-                'Project %s was incorrectly set-up. \
-                Can\'t parse owner and repo' % project.name)
+                'Project %s was incorrectly set-up. '  % project.name +
+                'Can\'t parse owner and repo.')
 
         query = prepare_query(owner, repo)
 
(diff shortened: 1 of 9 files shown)


anitya/lib/backends/github.py
```diff
@@ -88,8 +88,8 @@ def get_versions(cls, project):
             (owner, repo) = url.split('/')
         except ValueError:
             raise AnityaPluginException(
-                'Project %s was incorrectly set-up. \
-                Can\'t parse owner and repo' % project.name)
+                'Project %s was incorrectly set-up. '  % project.name +
+                'Can\'t parse owner and repo.')
```
jeremycline: It's neater to do it as:

``"Project {} was incorrectly set-up. Can't parse owner and repo.".format(project.name)``

rather than having the format in the middle of a string

anitya/lib/backends/github.py
```diff
@@ -88,8 +88,8 @@ def get_versions(cls, project):
             (owner, repo) = url.split('/')
         except ValueError:
             raise AnityaPluginException(
-                'Project %s was incorrectly set-up. ' % project.name +
-                'Can\'t parse owner and repo.')
+                """Project {} was incorrectly set-up.
+                Can\'t parse owner and repo.""".format(project.name))
```
jeremycline: FYI you can do
```
raise AnityaPluginException(
    "..."
    "more text".format())

Zlopez: Didn't know that, it can't be done with `%`.

Issue #565, opened 2018-06-20
FAS login into staging error 500


abitrolly:
When logging with FAS to https://release-monitoring.app.os.stg.fedoraproject.org/ it fails with error 500 on https://release-monitoring.app.os.stg.fedoraproject.org/complete/fedora/

Zlopez:
I'm adding error message from Firefox:
```
The character encoding of the HTML document was not declared. The document will render with garbled text in some browser configurations if the document contains characters from outside the US-ASCII range. The character encoding of the page must be declared in the document or in the transfer protocol.
```

Zlopez:
@abitrolly 
I tested this on my local deployment and I don't have this issue.
I'm using latest commit, so maybe this is already fixed.

Zlopez:
There was some issue with database setup, it's fixed now.
This issue can be closed

abitrolly:
I am still getting 500 error trying to login with FAS to staging. Does it work for you?

Zlopez:
Yes, I'm actually logged in using FAS account.

Zlopez:
@abitrolly 
Do you still have the issue with logging in?

abitrolly:
@Zlopez yes, I still have the issue with logging in to https://release-monitoring.app.os.stg.fedoraproject.org/

Zlopez:
@abitrolly 
Could you try https://stg.release-monitoring.org/
Because this should be the correct staging url.

abitrolly:
Doesn't work too - https://stg.release-monitoring.org/complete/fedora/?janrain_nonce=2018-08 ... gives 500.

Zlopez:
This is weird, I logged in without issue.
Do you have this issue in different application?
Could you post here, what is printed in console?

abitrolly:
In JavaScript console? Nothing there.

abitrolly:
No issues with other apps.

Zlopez:
Ok, I was able to check the logs and found the issue.
I'm not sure why this is happening, but according to the logs, Anitya is trying to insert your user in the database even if your user already exists.
I will check the db schema, maybe this is only caused by bad constraint or unique key in user database.

abitrolly:
Probably something is not saved into DB for new users.

jeremycline:
At some point I believe there were permissions issues on the database where the anitya user didn't have write access to some tables (possibly the social_auth tables?). I wonder if things aren't covered in a single transaction and it got part-way through making an account.

[...]

PR #615, opened 2018-09-21
Add check for social_auth table on user creation

Zlopez:
Fixes https://github.com/release-monitoring/anitya/issues/565

### DIFF
--- a/anitya/app.py
+++ b/anitya/app.py
@@ -143,11 +143,17 @@ def integrity_error_handler(error):
     # for the user.
     if 'email' in error.params:
         Session.rollback()
-        other_user = models.User.query.filter_by(email=error.params['email']).one()
-        social_auth_user = other_user.social_auth.filter_by(user_id=other_user.id).one()
-        msg = ("Error: There's already an account associated with your email, "
-               "authenticate with {}.".format(social_auth_user.provider))
-        return msg, 400
+        if 'social_auth' in error.params:
+            msg = (
+                "Error: Authentication with authentication provider failed. "
+                "Please try again later...")
+            return msg, 500
+        else:
+            other_user = models.User.query.filter_by(email=error.params['email']).one()
+            social_auth_user = other_user.social_auth.filter_by(user_id=other_user.id).one()
+            msg = ("Error: There's already an account associated with your email, "
+                   "authenticate with {}.".format(social_auth_user.provider))
+            return msg, 400
 
     return 'The server encountered an unexpected error', 500
 
--- a/anitya/db/events.py
+++ b/anitya/db/events.py
@@ -17,10 +17,11 @@
 import logging
 
 from sqlalchemy import event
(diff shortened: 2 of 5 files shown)


anitya/tests/base.py
```diff
@@ -287,5 +288,23 @@ def create_flagged_project(session):
     return flag
 
 
+def create_user(session):
```
jeremycline: With tests I think explicit is preferable to DRY. Having a function to create a user is indirection that makes the test harder to read an understand at a glance. I don't think this saves a huge amount of typing and isn't worth that indirection.

Zlopez: I'm not sure if this is true, the user is created on many places in tests. When I added this feature, almost 50 tests started failing. So this saved me a lot typing.

anitya/db/events.py
```diff
@@ -57,3 +58,23 @@ def set_ecosystem(session, flush_context, instances):
...
+        session (sqlalchemy.orm.session.Session): The session that is about to be committed.
+        flush_context (sqlalchemy.orm.session.UOWTransaction): Unused.
+        instances (object): deprecated and unused
+
+    Raises:
+        ValueError: If the social_auth table is not associated with this user.
+    """
+    for new_obj in session.new:
+        if isinstance(new_obj, User):
+            if len(new_obj.social_auth.all()) == 0:
```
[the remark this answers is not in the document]
Zlopez: Good point

anitya/app.py
```diff
@@ -143,10 +143,16 @@ def integrity_error_handler(error):
...
         Session.rollback()
-        other_user = models.User.query.filter_by(email=error.params['email']).one()
-        social_auth_user = other_user.social_auth.filter_by(user_id=other_user.id).one()
-        msg = ("Error: There's already an account associated with your email, "
-               "authenticate with {}.".format(social_auth_user.provider))
-        return msg, 400
+        if 'social_auth' in error.params:
+            msg = (
+                "Error: Authentication with authentication provider failed. "
+                "Please try again later...")
```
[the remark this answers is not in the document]
Zlopez: This means, that the user is already created, but without social_auth table entry. This is why there is this error message. I was not sure what HTTP error should be used for this. It should work when tried later.

anitya/tests/base.py
Zlopez: You are probably right in this, it is better if this is done in the same place as tests, not in imported code.

Zlopez:
I did a rebase to split this PR to fix and test only PR (https://github.com/release-monitoring/anitya/pull/619).

Zlopez:
I did a rebase after merge of #619. Tests are passing now.

Zlopez:
Resolved conflict.

Zlopez:

Zlopez:
Because @jeremycline approved this one previously and only thing I did was to fix failing tests which are now going through. I will merge this manually.

PR #618, opened 2018-09-26
Add towncrier

Zlopez:


docs/contributing.rst
```diff
@@ -54,6 +54,30 @@ Python API documentation is automatically generated from the code using Sphinx's
...
+* ``author``: for contributor names
+* ``other``: for other changes
+
+And where the ``source`` part of the filename is:
+
+* ``42`` when the change is described in issue ``42``
+* ``PR42`` when the change has been implemented in pull request ``42``, and
+  there is no associated issue
+* ``Cabcdef`` when the change has been implemented in changeset ``abcdef``, and
+  there is no associated issue or pull request.
```
bowlofeggs: Since you use mergify and it rebases, I recommend against using commit hashes - the author won't be able to predict their hash, and so any hashes they use will lead to a 404 link.

docs/contributing.rst
```diff
@@ -54,6 +54,30 @@ Python API documentation is automatically generated from the code using Sphinx's
...
+* ``username`` for contributors (``author`` extention). It should be the
+  username part of their commits' email address.
```
bowlofeggs: s/commits'/commit's/

news/get-authors.py
```diff
@@ -0,0 +1,19 @@
+#!/usr/bin/env python3
+
```
bowlofeggs: I recommend adding a docblock that explains the purpose of this script and any args it accepts (if any).

bowlofeggs: I didn't see a general Copyright claim on Anitya - I also recommend adding a copyright claim on this file.

docs/contributing.rst
Zlopez: Yeah, you are right. I didn't noticed this.

news/get-authors.py
Zlopez: Same as above.

Issue #593, opened 2018-08-29
Add sorting by creating date for versions


Zlopez:
Currently the version scheme used by Anitya is sorting the versions by numbers, which is not good in some situation.
I think that it will be better to use date when the version was retrieved. This should solve the problem with showing other than newest version as latest.

jeremycline:
I think it's reasonable to sort by date discovered in the UI, although do keep in mind that some projects maintain multiple major releases at once (e.g. 1.2.2 could come out, then 1.1.4 could come out) which might look odd unless it's obviously sorted by release date.

Zlopez:
Yes, the date of discovering must be shown on the UI. Otherwise it doesn't make sense.

Zlopez:
In the PR #625 I changed the behavior a little. The user can now choose which version scheme he wants to use for the project.

PR #625, opened 2018-10-03
Add sorting by creation date

Zlopez:
This PR introduces a new version scheme, that is allowing sort by creation_date.
Allows change of version scheme when creating/editing project.
Slight changes how the versions are displayed.

Fixes https://github.com/release-monitoring/anitya/issues/593


anitya/lib/versions/date.py
```diff
@@ -0,0 +1,56 @@
...
+"""
+
+from .base import Version
+import functools
+
+
+@functools.total_ordering
+class DateVersion(Version):
+    """
+    This class implements an Date version plugin.
```
bowlofeggs: s/an Date/a Date/

anitya/lib/versions/date.py
```diff
@@ -0,0 +1,56 @@
...
+
+    It sorts versions by date created.
+    """
+
+    name = u'Date'
+
+    def __eq__(self, other):
+        """
+        Compare two versions for equality by date created.
+        """
```
bowlofeggs: You could add a Returns block since it returns a ```bool```:

```
Returns:
    bool: True if the ```created_on``` attributes are equal.
```

anitya/lib/versions/date.py
```diff
@@ -0,0 +1,56 @@
...
+        return self.created_on == other.created_on
+
+    def __lt__(self, other):
+        """
+        Compare two versions by date created.
+        """
```
bowlofeggs: Similar here since it also returns a ```bool```.

anitya/templates/project_new.html
```diff
@@ -253,6 +262,7 @@ <h1>{{ context }} project</h1>
   }
 
   $(document).ready(function() {
+    add_tooltips();
```
bowlofeggs: This change seems unrelated to sorting by creation date - should it be a separate pull request?

anitya/tests/lib/versions/test_date.py
```diff
@@ -0,0 +1,87 @@
+# -*- coding: utf-8 -*-
+#
+# Copyright © 2017  Red Hat, Inc.
```
bowlofeggs: s/2017  /2018 /

anitya/tests/lib/versions/test_date.py
```diff
@@ -0,0 +1,87 @@
...
+# should have received a copy of the GNU General Public License along
+# with this program; if not, write to the Free Software Foundation,
+# Inc., 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301, USA.
+#
+# Any Red Hat trademarks that are incorporated in the source
+# code or documentation are not subject to the GNU General Public
+# License and may only be used or replicated with the express permission
+# of Red Hat, Inc.
+import unittest
+
```
bowlofeggs: No empty newline here for PEP-8.

anitya/db/migrations/versions/24b6734e8565_creation_date_on_version.py
```diff
@@ -0,0 +1,30 @@
+"""Creation date on version
+
+Revision ID: 24b6734e8565
+Revises: 34b9bb5fa388
+Create Date: 2018-10-01 11:17:19.457383
+"""
```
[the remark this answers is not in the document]
Zlopez: This file is generated by alembic. I'm not sure if we should add copyright on this.

anitya/lib/versions/date.py
```diff
@@ -0,0 +1,56 @@
...
+# Any Red Hat trademarks that are incorporated in the source
+# code or documentation are not subject to the GNU General Public
+# License and may only be used or replicated with the express permission
+# of Red Hat, Inc.
+"""
+This adds support for comparing project versions by date created.
+"""
+
+from .base import Version
+import functools
```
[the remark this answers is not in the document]
Zlopez: I always forgetting about this.

anitya/lib/versions/date.py
Zlopez: You are right.

anitya/templates/project_new.html
Zlopez: I did this only for small optimalization, but you are right, I will remove it from this PR.

anitya/tests/lib/versions/test_date.py
Zlopez: Strange flake8 didn't threw any error.

Zlopez:
There was a conflict in `anitya.forms` which I resolved when working on the small tweaks.

bowlofeggs:
On Fri, 2018-10-05 at 15:20 +0000, Zlopez wrote:
> Strange flake8 didn't threw any error.

Yeah this is one of the PEP-8 things that I don't think flake8 checks
for. I think there is a checker for this particular thing, but I can't
remember what it was called. I think @beav showed it to me a few years
ago…

Zlopez:
squashed

Issue #600, opened 2018-08-30
Add rate limit handling


Zlopez:
In current state GitHub API is raising `RateLimit` exception, which is only logged by cronjob.

It will be nice to reschedule the projects check when rate limit is reset. This could be handled by saving all the projects with GitHub backend after exception is raised and run check again only for them.

abitrolly:
A better architecture  is a proper queue/microservice with rate limiting, which flexibly adapts itself to rate limits, and exposes logs and exceptions to public.

But first, I would like to [see stats](https://github.com/sdispater/poetry/tags).

Zlopez:
Yes, this is something I have in my TODO. To better handle the amount of requests. Right now, we are doing check for all the projects in one go, which is not really nice.
I want to divide this load to the whole day. So it will check only a few projects every hour (maybe something more complex later).
The exposition of the logs and exceptions should be done using fedmsg like you proposed in https://github.com/release-monitoring/anitya/issues/574
However, I'm not sure if there is any topic for this right now.

I don't understand the link you sent, this is pointing to tags sdispater/poetry project. How this has to do anything with stats?

Zlopez:
I change the title, because this is not only about GitHub.

Zlopez:
@yarons 
This is not a bad idea, but we have plenty of backends and it;s not good to freeze everything, because one backend reaches rate limit.
In PR #633 I have it implemented for every project separately. Each project is checked after specified time (currently 1 hour) and if rate limit is thrown it is rescheduled to time returned in rate limit exception.

yarons:
I meant freezing the specific backend, not all of them :laughing:.

Zlopez:
I will add this to my TODO list and look at it later. For now it will work on per project basis.

yarons:
So I should be expecting lots of failures ahead :)

Zlopez:
I didn't noticed any ratelimit error with the current number of projects. So I don't think we will hit this anytime soon.

yarons:
![image](https://user-images.githubusercontent.com/406826/48901901-87a76b00-ee5f-11e8-96d8-686504f0d080.png)
Hmmmm sure?

Zlopez:
Oh, last time I checked this, there wasn't any rate limit error.

Zlopez:
It now looks like we are hitting those, but the current PR should still help with this.

Zlopez:
[...]

PR #633, opened 2018-10-10
Add rate limit for backends

Zlopez:
Checking for new versions is done by `files/anitya_cron.py` script. This script was modified by this PR to do check based on the interval specified in backend. 
This should allow us to run cron job more frequently without fear of hitting rate limit. (NOTE: Rate limit is now implemented only for GitHub, but it could be used for any backend. The backend only needs to raise `RateLimitException`).

What this adds:
- two new columns to `projects` table (`last_check`, `next_check`)
- default check interval to `BaseBackend` (1 hour)
- test for this new feature
- news fragment for towncrier

How this feature works:
1. Cron script will read the current `time` on start
2. Cron script selects all projects, that have `next_check < time`
3. Cron script executes checking for new versions
4. After check is done sets `last_check` to current time and `next_check` to:
  a. If `RateLimitException` was raised then it uses datetime value from this exception
  b. Otherwise sets value to `check_interval` for used backend

This fixes issue https://github.com/release-monitoring/anitya/issues/600.


anitya/db/models.py
```diff
@@ -438,6 +441,30 @@ def __json__(self, detailed=False):
 
         return output
 
+    def set_check_times(self, session, last_check, next_check=None):
```
jeremycline: I know the existing models have a lot of these kinds of functions, but it's an overly complicated way to do this and they should all be removed eventually. You can just set the values on the object directly.

anitya/db/models.py
```diff
@@ -496,6 +523,24 @@ def all(cls, session, page=None, count=False):
         else:
             return query.all()
 
+    @classmethod
+    def ready_for_check(cls, session, time):
```
jeremycline: Like the above function, this doesn't belong on the model class (even though there's lot of functions like this). Instead, if it's used in multiple places, it should be put into a SQLAlchemy Query object that sub-classes https://github.com/release-monitoring/anitya/blob/7d115fca74a424e55e1b17743aae3e7039d59b6c/anitya/db/meta.py#L104

Since this is used in just one place, I'd just write the query out in the cron job.

Also, just a note (and I know this only here because existing code does it), but you don't need to pass the session around like this. SQLAlchemy creates sessions that are local to threads: https://github.com/release-monitoring/anitya/blob/7d115fca74a424e55e1b17743aae3e7039d59b6c/anitya/db/meta.py#L42 and you can get the current session by doing ``db.Session()``. The same session is also used when doing ``db.Project.query``.

requirements.txt
```diff
@@ -15,3 +15,4 @@ straight.plugin
 psycopg2
 pytoml
 wtforms
+python-dateutil
```
jeremycline: As mentioned above, you may find arrow easier to work with: https://arrow.readthedocs.io/en/latest

anitya/lib/backends/__init__.py
```diff
@@ -70,13 +71,16 @@ class BaseBackend(object):
             the project is a part of do not define a default version scheme.
             If this is not defined, :data:`anitya.lib.versions.GLOBAL_DEFAULT`
             is used.
+        check_interval (obj:`datetime.timedelta`): Interval which is used for periodic
```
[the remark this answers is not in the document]
Zlopez: Oh, I looked at sphinx example for google docs and saw it there, but if it isn't needed I will remove it.

anitya/lib/exceptions.py
```diff
@@ -110,10 +111,13 @@ class RateLimitException(AnityaException):
 
     Args:
         reset_time (str): Time when limit will be reseted (UTC time encoded in ISO-8601).
+        dt_reset_time (obj:`datetime.datetime`): Python datetime object representation of
+            reset_time argument.
```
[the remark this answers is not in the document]
Zlopez: I'm using UTC for everything, so I don't thing I need zone awareness.

requirements.txt
Zlopez: I only need it to actually parse the ISO format of the date, even if it is in UTC.

anitya/db/models.py
Zlopez: I just tried to make it more coherent with existing code. Didn't know, that this is bad practice.

Zlopez: I have one question regarding this. Where should I place methods, that are relevant to project like `check_project_release()`? Is this a good candidate for `models.db.Project` method?

jeremycline: > I just tried to make it more coherent with existing code. Didn't know, that this is bad practice.

Yeah, no worries, I never got around to cleaning up the models.

> I have one question regarding this. Where should I place methods, that are relevant to project like check_project_release()? Is this a good candidate for models.db.Project method?

That's likely a good candidate. It deals with a single project and performs something other than changing the database row. The general rule of thumb that I follow is things that methods that deal with a single row live on the project model class, things that deal with queries and multiple rows live on the Query object.

anitya/lib/exceptions.py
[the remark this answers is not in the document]
jeremycline: Hmmm. I see a call to ``datetime.now()`` which returns the local TZ (https://docs.python.org/3/library/datetime.html#datetime.datetime.now). So you're assuming here that the server is set to use UTC. That's the sane thing to do, of course, but it's an assumption that might not be true for anyone else running this and it's a subtle issue to track down. It'd be good to either document that it must run on a server using UTC, or deal with time zones. I'm in favor of dealing with time zones

anitya/db/models.py
```diff
@@ -339,6 +339,9 @@ class Project(Base):
     latest_version = sa.Column(sa.String(50))
     logs = sa.Column(sa.Text)
 
+    last_check = sa.Column(sa.DateTime, default=datetime.datetime.utcnow)
+    next_check = sa.Column(sa.DateTime, default=datetime.datetime.utcnow)
```
jeremycline: Oh, I forgot to mention this on my first review, but you're querying on this column so it should probably be indexed.

anitya/db/migrations/versions/9262ccab3573_add_check_times.py
```diff
@@ -0,0 +1,41 @@
...
+
+
+def upgrade():
+    """ Add next_check and last_check columns to the projects table. """
+    op.add_column(
+        'projects',
+        sa.Column(
+            'last_check',
+            sa.DateTime,
+            server_default=sa.func.current_timestamp()
```
jeremycline: According to the PSQL docs this comes with a time zone, as well. If the database server isn't running in UTC as well, you're going to get weird issues with this if you just assume it's UTC.

anitya/lib/exceptions.py
```diff
@@ -117,7 +117,7 @@ class RateLimitException(AnityaException):
 
     def __init__(self, reset_time):
         self.reset_time = reset_time
-        self.dt_reset_time = dateutil.parser.parse(reset_time)
+        self.dt_reset_time = arrow.get(reset_time).naive
```
jeremycline: I don't think it makes sense to throw away the timezone information here. I don't think we need to hold onto ``reset_time`` either. I'd recommend writing this as:

```
def __init__(self, reset_time):
    self.reset_time = arrow.get(reset_time)
```

This way you keep the timezone information, can accept anything arrow.get accepts, and there's not two attributes representing the same thing.

Zlopez: I tried this and it didn't work, because the database isn't timezone aware and doesn't know how to store this information (at least in vagrant environment).

Zlopez:
Squashed commits and fixed conflicts.

anitya/db/models.py
```diff
@@ -345,6 +345,9 @@ class Project(Base):
     latest_version = sa.Column(sa.String(50))
     logs = sa.Column(sa.Text)
 
+    last_check = sa.Column(sa.DateTime, default=datetime.datetime.utcnow, index=True)
+    next_check = sa.Column(sa.DateTime, default=datetime.datetime.utcnow, index=True)
```
jeremycline: I looked at the [SQLAlchemy docs](https://docs.sqlalchemy.org/en/latest/core/type_basics.html#sqlalchemy.types.TIMESTAMP) and there's a timestamp type that supports time zones. I'd recommend just using that.

Zlopez: I will use it than.

Zlopez: After playing with it for some time, I found out that this isn't working with PostgreSQL.
See https://stackoverflow.com/questions/414952/sqlalchemy-datetime-timezone
Not sure how we should handle the timezone awareness than.

jeremycline: That question isn't about the data type I linked so I'm not sure what you mean by "isn't working".

Zlopez: Maybe I wasn't clear enough, I tried to use that data type with `timezone=True` and the information is still saved to database without timezone information.
So every time I try to compare it with timezone aware datetime I get error.

jeremycline: Does "that data type" mean DateTime? If so that's not surprising, the documentation says "For timezone support, use at least the TIMESTAMP datatype, if not the dialect-specific datatype object." which is a subclass of DateTime.

Zlopez: No, I tried to use the TIMESTAMP datatype with `timezone=True`.

Zlopez: @jeremycline 
Any other suggestion about timezones? Otherwise it looks good as it is.

jeremycline: Apologies, I've not had much time to devote to this. Did you remember to drop the database and re-create it? In any case, since you're the one on the hook for problems if you're comfortable with the current approach, then I'll approve it.

Zlopez: It's possible I didn't remember to recreate database, I will try it again.

Zlopez: I recreated whole vagrant and I'm still getting this error:
```
TypeError: can't compare offset-naive and offset-aware datetimes
```
The database doesn't store offset, even when you have it defined like this:
```
    last_check = sa.Column(sa.TIMESTAMP(timezone=True),
                           default=arrow.utcnow().datetime, index=True)
    next_check = sa.Column(sa.TIMESTAMP(timezone=True),
                           default=arrow.utcnow().datetime, index=True)
```
Only solution to this, I see, is to store every datetime value to database in UTC.

Zlopez:

The coverage looks like it's decreasing, but this is caused by removal of some code that was covered before.
Not sure, why codecov has issue with this.

Zlopez:
Fix conflict

jeremycline:
There's still a conflict, I'll give it a look over once you've resolved it

Zlopez:
Fixed conflict.

Zlopez:
Rebased.
Everything should be ready for merging this PR.
Finally \o/

Zlopez:
I did a squash and fixed the issue with alembic.

Hopefully everything will be ok now.

Zlopez:
@jeremycline 
It should be ready for merge.

anitya/db/models.py
```diff
@@ -345,6 +346,11 @@ class Project(Base):
     latest_version = sa.Column(sa.String(50))
     logs = sa.Column(sa.Text)
 
+    last_check = sa.Column(sa.TIMESTAMP(timezone=True),
+                           default=arrow.utcnow().datetime, index=True)
+    next_check = sa.Column(sa.TIMESTAMP(timezone=True),
+                           default=arrow.utcnow().datetime, index=True)
```
jeremycline: This probably isn't what you want. The default here is a scalar, not a function. ``default`` accepts both, but what I expect you want to happen is for whatever the current time is when this project is added. This will be whatever time the code first runs (probably import time). You could do ``lambda: arrow.utcnow().datetime``.

anitya/lib/utilities.py
```diff
@@ -67,13 +68,27 @@ def check_project_release(project, session, test=False):
     versions_prefix = []
     max_version = ''
 
+    # don't change actual data during test run
+    if not test:
+        project.last_check = datetime.now()
```
jeremycline: ``arrow.utcnow().datetime`` I think.

anitya/lib/utilities.py
```diff
@@ -67,13 +68,27 @@ def check_project_release(project, session, test=False):
...
+        if not test:
+            project.logs = str(err)
+            session.add(project)
+            session.commit()
+        raise
+    except exceptions.RateLimitException as err:
+        _log.exception("AnityaError catched:")
+        if not test:
+            project.logs = str(err)
+            project.next_check = err.reset_time.to('utc').naive
```
jeremycline: You probably don't want the naive time here

files/anitya_cron.py
```diff
@@ -51,21 +53,19 @@ def projects_by_feed(session):
...
-    finally:
-        session.get_bind().dispose()
-        session.remove()
 
 
 def main(debug, feed):
     ''' Retrieve all the packages and for each of them update the release
     version.
     '''
+    time = datetime.now()
```
jeremycline: ``arrow.utcnow()``

anitya/db/models.py
Zlopez: Thanks for note, didn't noticed this.

anitya/lib/utilities.py
Zlopez: I found out, that I need this for python 2.7. Otherwise tests are starting to fail.

anitya/lib/utilities.py
Zlopez: I had to use `arrow.utcnow().naive`. Otherwise the python 2.7 tests are starting to fail.

anitya/lib/utilities.py
jeremycline: Why are you trying to support Python 2.7? I don't see any value in it, I'd say support 3.6+ or something.

Zlopez: You think it is time to drop 2.7 tests from tox and CI?

Zlopez: But I drop them with pleasure

Zlopez: But first I update this PR

Zlopez:
I dropped python 2.7 support. But it is still enabled in travis.

Zlopez:
The tests are passing, I think we are ready to merge this into master.

files/anitya_cron.py
```diff
@@ -93,7 +94,13 @@ def main(debug, feed):
         projects = list(projects_by_feed(session))
         session.commit()
     else:
-        projects = db.Project.all(session)
+        # Get all projects, that are ready for check
+        projects = session.query(
```
jeremycline: You can use ``db.Project.query...`` to get a query, rather than ``session.query(db.Project)``, FYI

anitya/db/migrations/versions/9262ccab3573_add_check_times.py
```diff
@@ -0,0 +1,42 @@
...
+# revision identifiers, used by Alembic.
+revision = '9262ccab3573'
+down_revision = 'b13662e5d288'
+
+
+def upgrade():
+    """ Add next_check and last_check columns to the projects table. """
+    op.add_column(
+        'projects',
+        sa.Column(
```
jeremycline: I don't see indexes being added or removed here -  did you regenerate your migration after altering the model to include an index?

Zlopez: I forgot about that

Zlopez:
I fixed few last things, found out, there were some issues with cron script, but it is working now.

And found out another new [issue](https://github.com/release-monitoring/anitya/issues/674) during cron test run, but I will fix this separately.

PR #635, opened 2018-10-12
Simplify logs

Zlopez:
This PR introduces simple status for every project instead of current robust log table, which is not really needed (all these log messages are sent through fedmsg anyway).
This PR is split to two commits.

First commit is dropping old logs table and removing every reference to it.

Second commit is introducing new flag on project which contains current update status. Current `Logs` section on project page is changed to `Status` section which is showing the `update_status` information:
* `None` - no version retrieved
screenshot from 2018-10-12 12-00-57
* `True` - version retrieved successfully
screenshot from 2018-10-12 12-01-13
* `False` - failure when obtaining version
screenshot from 2018-10-12 12-01-27

After merge of https://github.com/release-monitoring/anitya/pull/633 I will add a new column Last check time to above table.


anitya/db/models.py
```diff
@@ -338,6 +248,7 @@ class Project(Base):
 
     latest_version = sa.Column(sa.String(50))
     logs = sa.Column(sa.Text)
+    update_status = sa.Column(sa.Boolean, default=None)
```
jeremycline: I imagine this will be something that is queried on, so it should probably have an index. Also, a nit, but the name is a bit confusing as a boolean: "update_status = False" doesn't mean much to me. "update_successful" or similar would be clearer. Does it need to be nullable?

One thing you could do is make this column an integer and store the HTTP status code. If things like DNS or TLS validation fail you could store a constant you define. Then you don't need the logs column here either.

Zlopez: My idea was to use the `update_status` in this way:
- if `None` than update wasn't done yet - this should only happen on new project
- if `True` update was successful - this should be set if project was updated successfully, even if there is no new version retrieved
- if `False` update failed - this should be set if project check for new version failed for any reason

`project.logs` should contain error encountered when retrieving the version.

I wanted to name it `update` instead of `update_status`, but there is already method named `update`. But this could be easily changed to `update_successful`.

I don't think the HTTP status is good indicator. Most of the checks are failing before or after HTTP request is made. Like `Project is not correctly set-up` or `Can't parse new version succesfully`. These are the most frequent errors that raise `AnityaException` without any HTTP status code. I can probably map them to some specific status code and create a mapping for status codes to status messages.

What do you think about this?

jeremycline: Hmm. I guess there are more possible errors than would be terribly fun to enumerate, and ``logs`` isn't indexed there's not a huge return on investment.

The only thing I'd think about is whether or not you want this column and also a column to track the last time a project was checked (so when the cron job runs you sort by oldest) or if those could be combined in some way.

This is only slightly related, but it doesn't make sense to me for a project to be created and not checked (I always check the "check now" box). Maybe we should force a check before creating the project. This way we will always report to the user their settings don't work and we don't have to handle the "project is created, but not checked" case.

Oh, and whatever you decide on it needs a docblock entry.

Zlopez: The column containing last time the project was checked is already implemented here https://github.com/release-monitoring/anitya/pull/633.

But good idea about the automatic check of project. In this case, I just set the `update_status`, maybe I rename it to `update_successful`, to `False` until the check is actually successful.

Still it will be better to actually have some button `test` on create project site, that will try to check for new version, before commiting project to database and even better will be to do this also automatically in backend before commiting project.

Oh, I see I forgot to update docstring of Project class.

Zlopez: Main issue with automatic check right now is that you need the project to exist in the database before check. This is the reason, why you can't do test check when creating the project.

Zlopez:
I did a rebase with master and renamed `update_status` to `check_successful`.
Also I updated `Project` docstring.

Zlopez:
@jeremycline 
This is a good idea I will do it. About the tests, there isn't much code added and the diff hit 75% is because one line was changed in the code that isn't covered by any tests, but this is fixed here https://github.com/release-monitoring/anitya/pull/637

Zlopez:
I added back the `/logs` page. It's now looking like this:
screenshot from 2018-10-18 12-39-42

But it's now using `updated_on` field instead of `last_check`, because `last_check` is part of #633. So it will be probably better to merge #633 first and do small changes to this PR after.

Zlopez:
I fixed the conflicts, failing tests and added last_check information to project and logs page.
screenshot from 2018-12-07 19-43-59

I think it's ready for merge. \o/

anitya/tests/db/test_models.py
```diff
@@ -516,6 +549,7 @@ def test_init_distro(self):
         self.assertEqual(distros[0].name, 'Debian')
         self.assertEqual(distros[1].name, 'Fedora')
 
+    # TODO: Split to separate tests
```
jeremycline: I recommend filing a ticket in your issue tracker over leaving a # TODO comment.

anitya/db/migrations/versions/1324a36b9f69_add_updated_check_to_project.py
```diff
@@ -0,0 +1,42 @@
+"""Add updated check to project
```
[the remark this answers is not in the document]
Zlopez: Not any particular reason, I will merge them together.

anitya/db/models.py
```diff
@@ -549,6 +465,36 @@ def by_distro(cls, session, distro, page=None, count=False):
         else:
             return query.all()
 
+    @classmethod
+    def by_last_check(cls, session, page=None, count=False):
```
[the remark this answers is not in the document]
Zlopez: Ok, I will move it to query class. I need to do a refactoring of this later.

anitya/tests/db/test_models.py
Zlopez: I created a new issue for this https://github.com/release-monitoring/anitya/issues/680 and I will remove the TODO comment.

anitya/db/models.py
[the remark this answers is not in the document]
Zlopez: At the end I decided to remove it altogether. It's called only in one place and it's basically alias for `Project.query.order_by(Project.last_check.desc())`. The second reason for me to decide to remove it, is that I wasn't able to call `order_by` in separate source file without creating circle dependency.
In `anitya.db.models` I needed to import `anitya.db.queries` to replace `query` property of `Project` model and in the `anitya.db.queries` I needed to import `anitya.db.models` to be able to call `order_by(Project.last_check.desc())`.

Zlopez:
This pull request **fixes 2 alerts** when merging dbc6d16fac16eb3b891a48f32615b80be7ea7a63 into 1385598a0249062469da3e9e1dc000939e3408e2 - [view on LGTM.com](https://lgtm.com/projects/g/release-monitoring/anitya/rev/pr-cccc9f37f1e55453d6e03b482980ba7bd162781c)

**fixed alerts:**

* 2 for Conflicting HTML element attributes

---

*Comment posted by [LGTM.com](https://lgtm.com)*

Zlopez:
I did a rebase and squashed and reworded commits.

Zlopez:
This pull request **fixes 2 alerts** when merging 94fb978480dc881c7263000e1a269414ccec57ce into 44e7fb7ecd7bcc4284d100939080c20d6b00dfe1 - [view on LGTM.com](https://lgtm.com/projects/g/release-monitoring/anitya/rev/pr-e365db2714db779e09e3028d7dcc5a603da5fe3e)

**fixed alerts:**

* 2 for Conflicting HTML element attributes

---

*Comment posted by [LGTM.com](https://lgtm.com)*

Zlopez:
I'm merging this manually, because I addressed all the comments made by @jeremycline and this is the last PR blocking release of 0.14.0.

PR #637, opened 2018-10-16
Raise tests coverage

Zlopez:
This PR was originally part of the https://github.com/release-monitoring/anitya/pull/608, but I decided to separate the tests from it to make it more readable for the reviewers.

### DIFF
--- a/anitya/tests/db/test_models.py
+++ b/anitya/tests/db/test_models.py
@@ -26,6 +26,7 @@ anitya tests of the models.
 from uuid import uuid4, UUID
 import datetime
 import unittest
+import time
 
 from sqlalchemy.dialects import postgresql, sqlite
 from sqlalchemy.types import CHAR
@@ -39,6 +40,7 @@ from anitya.tests.base import (
     DatabaseTestCase, create_distro,
     create_project, create_package,
     create_flagged_project)
+from anitya.lib import utilities
 
 
 class ProjectTests(DatabaseTestCase):
@@ -178,6 +180,7 @@ class ProjectTests(DatabaseTestCase):
     def test_project_search(self):
         """ Test the Project.search function. """
         create_project(self.session)
+        create_package(self.session)
 
         projects = models.Project.search(self.session, '*', count=True)
         self.assertEqual(projects, 3)
@@ -187,6 +190,27 @@ class ProjectTests(DatabaseTestCase):
 
         projects = models.Project.search(self.session, '*', page='asd')
         self.assertEqual(len(projects), 3)
+
+    def test_project_search_no_pattern(self):
+        """
+        Assert that all projects are returned when
+        no pattern is provided.
+        """
+        create_project(self.session)
+
+        projects = models.Project.search(self.session, '')
+        self.assertEqual(len(projects), 3)
+
+    def test_project_search_by_distro(self):
+        """
+        Assert that only projects with mappings to specific distro
+        are returned when distro is provided.
+        """
+        create_project(self.session)
+        create_package(self.session)
+
+        projects = models.Project.search(self.session, '*', distro='Fedora')
+        self.assertEqual(len(projects), 2)
 
     def test_project_get_or_create(self):
         """ Test the Project.get_or_create function. """
@@ -199,6 +223,11 @@ class ProjectTests(DatabaseTestCase):
         self.assertEqual(project.homepage, 'https://test.org')
         self.assertEqual(project.backend, 'custom')
 
+    def test_project_get_or_create_exception(self):
+        """
+        Assert that exception is raised when incorrect
+        backend is provided.
+        """
         self.assertRaises(
             ValueError,
             models.Project.get_or_create,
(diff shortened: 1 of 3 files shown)


anitya/tests/db/test_models.py
```diff
@@ -155,6 +162,12 @@ def test_project_search(self):
         projects = models.Project.search(self.session, '*', page='asd')
         self.assertEqual(len(projects), 3)
 
+        projects = models.Project.search(self.session, '')
```
jeremycline: I think you can break both of these new tests out into separate methods and add doc-blocks to them describing what they're testing. Tests should have as few assertions in them as possible so that when they fail, it's very quickly obvious what went wrong (ideally from the name of the test and doc-block).

anitya/tests/db/test_models.py
```diff
@@ -175,9 +188,83 @@ def test_project_get_or_create(self):
             backend='foobar'
         )
 
+        create_project(self.session)
```
jeremycline: Same thing here as above, you can break it into its own function.

anitya/tests/db/test_models.py
```diff
@@ -175,9 +188,83 @@ def test_project_get_or_create(self):
...
+            self.session,
+            name='geany',
+            homepage='https://www.geany.org/')
+        post_projects = models.Project.search(self.session, '*', count=True)
+        self.assertEqual(project.name, 'geany')
+        self.assertEqual(project.homepage, 'https://www.geany.org/')
+        self.assertEqual(pre_projects, post_projects)
+
+    def test_project_updated(self):
+        """ Test the Project.updated method. """
```
jeremycline: This should be broken into multiple test cases

anitya/tests/db/test_models.py
```diff
@@ -188,6 +275,29 @@ def test_init_distro(self):
         self.assertEqual(distros[0].name, 'Debian')
         self.assertEqual(distros[1].name, 'Fedora')
 
+    def test_distro_search(self):
```
jeremycline: This should be broken into multiple test cases

anitya/tests/db/test_models.py
```diff
@@ -260,6 +360,79 @@ def test_packages__repr__(self):
...
 
 
+class ProjectFlagTestCase(DatabaseTestCase):
+    """ Tests for ProjectFlag model. """
+
+    def test_project_flag__repr__(self):
+        """ Test the ProjectFlag.__repr__ function. """
+        flag = create_flagged_project(self.session)
+
+        self.assertEqual(str(flag), '<ProjectFlag(geany, dgay@redhat.com, open)>')
```
jeremycline: The test name implies it's testing ``__repr__``, shouldn't this call ``repr(flag)``?

anitya/tests/db/test_models.py
```diff
@@ -260,6 +360,79 @@ def test_packages__repr__(self):
...
+        self.assertEqual(flag.__json__(), data)
+
+        data['reason'] = u'This is a duplicate.'
+        self.assertEqual(flag.__json__(detailed=True), data)
+
+    def test_project_flag_all(self):
+        """ Test the ProjectFlag.all function. """
+        create_flagged_project(self.session)
+
+        flags = models.ProjectFlag.all(self.session)
```
jeremycline: Ooof. I'm fine adding this test, but this function could be deleted and replaced with the normal SQLAlchemy ``models.ProjectFlag.query.all()``.

anitya/tests/db/test_models.py
```diff
@@ -260,6 +360,79 @@ def test_packages__repr__(self):
...
+        self.assertEqual(len(flags), 1)
+
+    def test_project_flag_search(self):
```
jeremycline: This should be broken into multiple test cases

anitya/tests/test_flask.py
```diff
@@ -497,14 +503,38 @@ def test_distro_projects_search(self):
         self.assertTrue(
             b'<h1>Search projects in Fedora</h1>' in output.data)
 
+        create_package(self.session)
+
+        output = self.app.get('/distro/Fedora/search/g')
```
jeremycline: These should be broken into multiple test cases

anitya/tests/test_flask.py
```diff
@@ -514,13 +544,18 @@ def test_projects_search(self):
...
 
-        self.assertEqual(output.data.count(b'<a href="/project/'), 1)
+        self.assertEqual(output.data.count(b'<a href="/project/1'), 1)
 
         output = self.app.get('/projects/search/?page=ab')
         self.assertEqual(output.status_code, 200)
         self.assertTrue(expected in output.data)
         self.assertEqual(output.data.count(b'<a href="/project/'), 3)
 
+        output = self.app.get('/projects/search/?exact=1')
```
jeremycline: Separate test :smiley_cat:

anitya/tests/test_flask.py
```diff
@@ -530,6 +565,75 @@ def test_projects_search(self):
             b'exact match, redirecting</li>'
         self.assertTrue(expected in output.data)
 
+    def test_logout(self):
+        """Assert the logout logouts user"""
+        with login_user(self.flask_app, self.user):
+            output = self.app.get('/logout')
+            self.assertEqual(output.status_code, 302)
+            self.assertEqual(
```
jeremycline: This asserts the logout endpoint redirects the user to /, but nothing about the state of the user's login.

anitya/tests/test_flask.py
```diff
@@ -530,6 +565,75 @@ def test_projects_search(self):
...
+                output.headers['Location'],
+                'http://localhost/'
+            )
+
+    def test_projects_search_by_name(self):
```
jeremycline: This should be multiple test cases

anitya/tests/test_flask.py
```diff
@@ -530,6 +565,75 @@ def test_projects_search(self):
...
+            ecosystem_name='pypi',
+        )
+        self.session.add(project)
+        self.session.commit()
+
+        output = self.app.get('/project/geany*/')
+        self.assertEqual(output.status_code, 200)
+        self.assertEqual(output.data.count(b'<a href="/project/'), 2)
+
+    def test_projects_updated(self):
```
jeremycline: Multiple test cases

anitya/tests/db/test_models.py
Zlopez: Yes, I should, but I didn't wanted to rewrite whole file. I used the same format that was already there.
But it is maybe good time to start with refactoring.

anitya/tests/db/test_models.py
Zlopez: I tried to create tests for methods that weren't covered. This was one of them.

anitya/tests/db/test_models.py
Zlopez: Probably, my bad.

anitya/tests/test_flask.py
Zlopez: Yeah, this should be here, I'm just need to look how to test this.

anitya/tests/db/test_models.py
jeremycline: Yep, I just wanted to let you know this is not a normal pattern :smile:

anitya/tests/db/test_models.py
jeremycline: Yeah, I don't think you need to break out the existing tests, but it's best not to make a bad situation worse, so I'd say any new assertions you're making should be in a new test case.

anitya/tests/db/test_models.py
```diff
@@ -175,9 +208,164 @@ def test_project_get_or_create(self):
...
+        self.assertEqual(len(projects), 3)
+        self.assertEqual(projects[0].logs, None)
+
+    def test_project_updated_newer_updated(self):
+        """
+        Assert that only not updated projects are returned when status is
+        set to 'newer_updated'.
+        """
+        create_project(self.session)
+        for project in self.session.query(models.Project).all():
```
jeremycline: FYI ``models.Project.query.all()`` should be equivalent here (db/meta.py sets up the query property). You don't need to change it, but all the existing code passing the session around is from code I never refactored and isn't necessary.

Zlopez:
The PR fixing the lint issue was merged, so I'm doing a rebase to fix also this PR.

Zlopez:
Fixed conflicts and rebased

Zlopez:

Issue #594, opened 2018-08-29
Add separate form for adding new distribution


Zlopez:
Currently Anitya has only text input field for distribution mapping.

It will be nice to have separate form for adding new distribution.
And when editing mapping on project you will only have selectbox.

- [x] Add new form to frontend for adding distribution - nice place for this will be distro mapping on project or distros page
- [x] Change text input field to selectbox for distribution mapping

Zlopez:
Merged into master. Closing.

PR #639, opened 2018-10-17
Update form for adding distribution

Zlopez:
This PR implements https://github.com/release-monitoring/anitya/issues/594.

What was changed:
* To add distribution through distros page user no longer needs to be admin
* Distribution in mapping page is now selected from dropdown, not automatically created if it doesn't exists
* Add confirmation dialog to `/distro/add` page if there are distributions with similar names (this is also showed on the page itself before submit)
* Polished `/distro/add` and `/project/<id>/map` pages to look a little better
* Add information where user can add new distribution directly to `/project/<id>/map`
* Fix failing test

This changes should prevent making distributions containing typos and duplicates of already existing distributions.


news/PR639.feature
```diff
@@ -0,0 +1 @@
+Update form for adding new distribution
```
bowlofeggs:
```suggestion
Update form for adding new distributions
```

anitya/forms.py
```diff
@@ -66,7 +66,10 @@ class FlagProjectForm(FlaskForm):
 
 
 class MappingForm(FlaskForm):
-    distro = StringField('Distribution', [validators.DataRequired()])
+    distro = SelectField(
+        'Distribution',
+        [validators.DataRequired()],
+        choices=[(item, item) for item in []])
```
bowlofeggs: What express it this way? Isn't this basically just saying ```choices=[]```?

anitya/tests/test_flask.py
```diff
@@ -919,3 +919,69 @@ def test_invalid_project_id(self):
...
+        with login_user(self.flask_app, self.user):
+            output = self.client.post('/distro/add', data={'csrf_token': 'abc', 'name': 'Fedora'})
+            self.assertEqual(200, output.status_code)
+            self.assertEqual(0, models.Distro.query.count())
+
+    def test_add_distro(self):
+        """Assert admins can add distributions."""
+        with login_user(self.flask_app, self.user):
+            output = self.client.get('/distro/add')
+            print(output.data)
```
bowlofeggs:
```suggestion
```

anitya/tests/test_flask.py
```diff
@@ -919,3 +919,69 @@ def test_invalid_project_id(self):
...
+            csrf_token = output.data.split(
+                b'name="csrf_token" type="hidden" value="')[1].split(b'">')[0]
+            data = {'name': 'Fedora', 'csrf_token': csrf_token}
+
+            output = self.client.post('/distro/add', data=data, follow_redirects=True)
+
+            # self.assertEqual(201, output.status_code)
```
bowlofeggs:
```suggestion
```

anitya/tests/test_flask.py
```diff
@@ -919,3 +919,69 @@ def test_invalid_project_id(self):
...
+    def test_add_distro(self):
+        """Assert admins can add distributions."""
+        with login_user(self.flask_app, self.user):
+            output = self.client.get('/distro/add')
+            print(output.data)
+            csrf_token = output.data.split(
+                b'name="csrf_token" type="hidden" value="')[1].split(b'">')[0]
+            data = {'name': 'Fedora', 'csrf_token': csrf_token}
+
+            output = self.client.post('/distro/add', data=data, follow_redirects=True)
```
bowlofeggs: Does ```self.client.post()``` assert that the return code is ```200``` by default?

anitya/tests/test_flask.py
```diff
@@ -919,3 +919,69 @@ def test_invalid_project_id(self):
...
+            output = self.client.get('/distro/add')
+            csrf_token = output.data.split(
+                b'name="csrf_token" type="hidden" value="')[1].split(b'">')[0]
+            data = {'name': 'Fedora', 'csrf_token': csrf_token}
+
+            create_output = self.client.post('/distro/add', data=data, follow_redirects=True)
+            dup_output = self.client.post('/distro/add', data=data, follow_redirects=True)
+
+            self.assertTrue(b'Distribution added' in create_output.data)
+            self.assertTrue(b'Could not add this distro' in dup_output.data)
```
bowlofeggs: It seems like it'd be good to assert that an HTTP error code was used here. Perhaps 409 (conflict)?

```suggestion
            self.assertTrue(b'Could not add this distro' in dup_output.data)
            self.assertEqual(409, output.status_code)
```

Zlopez: The user is redirected to `/distros` page, so the user gets 301 redirect.

Zlopez:
squash

Zlopez:
The PR fixing the lint issue was merged, so I'm doing a rebase to fix also this PR.

PR #647, opened 2018-10-25
Fix for failing lint test

Zlopez:
This PR should fix the failing lint test.

There was new version of flake that started checking things that are deprecated in newer versions of python. Unfortunately I still had old version in my tox env, so I didn't noticed this.


Zlopez:
Good advice, I usually write this to description of the PR, but it will be better to have it also in commit description.

PR #648, opened 2018-10-27
Adds additional recent python versions to testing

00willo:
Adds python 3.7 and the nightly 3.8 build (allowed to fail) for testing.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -21,5 +21,15 @@ matrix:
       env: TOXENV=py35
     - python: 3.6
       env: TOXENV=py36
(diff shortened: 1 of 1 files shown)


00willo:
@Zlopez I'm always a fan of running the latest being developed, even if just to know and keep on eye on where you stand, how much work you have coming in relation to tests or code etc. It is set to not fail the build, so it's only there for reference.

Zlopez:
@00willo 
You are right. If this isn't messing with tests it is only a reference.

PR #649, opened 2018-10-29
Add GDPR script

Zlopez:
Add SAR (Subject Access Requests) GDPR (General Data Protection Regulation)
script that could be used to gather user data from database.
This script currently takes data from users and
social_auth_usersocialauth table and saves them in json format.

### DIFF
--- a/anitya/db/models.py
+++ b/anitya/db/models.py
@@ -983,6 +983,31 @@ class User(Base):
         """
         return six.text_type(self.id)
 
+    def to_dict(self):
+        """
+        Creates json compatible dict from `User`.
+
+        Returns:
+            dict: `User` object transformed to dictionary.
+        """
+        social_auth = []
+        for soc_auth in self.social_auth.all():
+            social_auth.append(
+                dict(
+                    provider=soc_auth.provider,
+                    extra_data=soc_auth.extra_data,
+                    uid=soc_auth.uid,
+                )
+            )
+
+        return dict(
+            id=str(self.id),
+            email=self.email,
+            username=self.username,
+            active=self.active,
+            social_auth=social_auth,
+        )
+
 
 def _api_token_generator(charset=string.ascii_letters + string.digits, length=40):
     """
--- a/anitya/sar.py
+++ b/anitya/sar.py
@@ -1,0 +1,73 @@
+#!/usr/bin/env python3
+# -*- coding: utf-8 -*-
+#
+# Copyright 2018  Red Hat, Inc.
+#
+# This copyrighted material is made available to anyone wishing to use,
+# modify, copy, or redistribute it subject to the terms and conditions
+# of the GNU General Public License v.2, or (at your option) any later
+# version.  This program is distributed in the hope that it will be
+# useful, but WITHOUT ANY WARRANTY expressed or implied, including the
+# implied warranties of MERCHANTABILITY or FITNESS FOR A PARTICULAR
+# PURPOSE.  See the GNU General Public License for more details.  You
+# should have received a copy of the GNU General Public License along
+# with this program; if not, write to the Free Software Foundation,
+# Inc., 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301, USA.
+#
+# Any Red Hat trademarks that are incorporated in the source
+# code or documentation are not subject to the GNU General Public
+# License and may only be used or replicated with the express permission
+# of Red Hat, Inc.
+#
+"""
+This script is used for GDPR SAR (General Data Protection Regulation
+Subject Access Requests).
+It returns information about specific user saved by Anitya.
+It reads SAR_USERNAME and SAR_EMAIL environment variables as keys for
+getting data from database.
+
+ Authors:
+   Michal Konecny <mkonecny@redhat.com>
+"""
+
+import logging
+import os
+import json
+import sys
+
+from anitya.config import config
+from anitya import db
+
+_log = logging.getLogger('anitya')
+
+
+def main():
+    '''
+    Retrieve database entry for user.
+    '''
+    db.initialize(config)
+    _log.setLevel(logging.DEBUG)
+    sar_username = os.getenv('SAR_USERNAME')
+    sar_email = os.getenv('SAR_EMAIL')
+
+    users = []
+
+    if sar_email:
+        _log.debug('Find users by e-mail {}'.format(sar_email))
+        users = users + db.User.query.filter_by(email=sar_email).all()
+
+    if sar_username:
+        _log.debug('Find users by username {}'.format(sar_username))
+        users = users + db.User.query.filter_by(username=sar_username).all()
+
+    users_list = []
+    for user in users:
+        users_list.append(user.to_dict())
+
+    json.dump(users_list, sys.stdout)
+
+
+if __name__ == '__main__':
+    _log.info('SAR script start')
+    main()
+    _log.info('SAR script end')
--- a/anitya/tests/test_sar.py
+++ b/anitya/tests/test_sar.py
@@ -1,0 +1,187 @@
+# -*- coding: utf-8 -*-
+#
+# Copyright © 2018  Red Hat, Inc.
+#
+# This copyrighted material is made available to anyone wishing to use,
+# modify, copy, or redistribute it subject to the terms and conditions
+# of the GNU General Public License v.2, or (at your option) any later
+# version.  This program is distributed in the hope that it will be
+# useful, but WITHOUT ANY WARRANTY expressed or implied, including the
+# implied warranties of MERCHANTABILITY or FITNESS FOR A PARTICULAR
+# PURPOSE.  See the GNU General Public License for more details.  You
+# should have received a copy of the GNU General Public License along
+# with this program; if not, write to the Free Software Foundation,
+# Inc., 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301, USA.
+#
+# Any Red Hat trademarks that are incorporated in the source
+# code or documentation are not subject to the GNU General Public
+# License and may only be used or replicated with the express permission
+# of Red Hat, Inc.
+#
+
+'''
+anitya tests for GDPR SAR script.
+'''
+
+import pytest
+import mock
+import json
+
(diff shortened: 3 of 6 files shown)


anitya/db/models.py
```diff
@@ -948,6 +948,31 @@ def get_id(self):
         """
         return six.text_type(self.id)
 
+    def __json__(self, detailed=False):
```
bowlofeggs: It looks like ```detailed``` does not do anything (and it is also undocumented).

anitya/sar.py
```diff
@@ -0,0 +1,77 @@
...
+"""
+
+import logging
+import os
+import json
+
+from anitya.config import config
+from anitya import db
+
+LOG = logging.getLogger('anitya')
```
bowlofeggs: In Python, naming a variable in all caps conventionally means that it is a constant. The logger isn't a constant, so I recommend naming it ```_log``` (since it is also meant to be private to this module).

anitya/sar.py
```diff
@@ -0,0 +1,77 @@
...
+        LOG.debug('Find users by username {}'.format(SAR_USERNAME))
+        users = users + session.query(db.User).filter(db.User.username == SAR_USERNAME).all()
+
+    users_list = []
+    for user in users:
+        users_list.append(user.__json__())
+
+    with open(FILENAME, 'w') as fd:
+        LOG.info('Dump data to json {}'.format(users_list))
+        json.dump(users_list, fd)
```
bowlofeggs: Note that the SAR data must be written to stdout as documented here:

https://fedora-infra-docs.readthedocs.io/en/latest/sysadmin-guide/sops/gdpr_sar.html#script

anitya/tests/test_sar.py
```diff
@@ -0,0 +1,137 @@
...
+        )
+        self.session.add(user2)
+        self.session.commit()
+
+        sar.SAR_EMAIL = 'user@fedoraproject.org'
+        sar.FILENAME = os.path.join(self.tempDir, 'anitya.json')
+        sar.main()
+
+        testdata = open(sar.FILENAME).read()
+        print(testdata)
```
bowlofeggs: Stray print statement.

anitya/tests/test_sar.py
```diff
@@ -0,0 +1,137 @@
+# -*- coding: utf-8 -*-
+#
+# Copyright © 2014  Red Hat, Inc.
```
bowlofeggs: 2018

anitya/db/models.py
Zlopez: Copied the method from other class and didn't noticed this parameter. Will remove.

anitya/sar.py
Zlopez: Copied from `anitya_cron.py` script. Will change.

anitya/sar.py
Zlopez: I thought the output is `sar_output_file` environment variable. If it should only print on stdout it is much easier to test.

anitya/tests/test_sar.py
Zlopez: Will remove.

anitya/tests/test_sar.py
Zlopez: Copied from another file. Didn't noticed. Will change.

anitya/sar.py
```diff
@@ -0,0 +1,75 @@
...
+    '''
+    db.initialize(config)
+    session = db.Session()
+    _log.setLevel(logging.DEBUG)
+
+    users = []
+
+    if SAR_EMAIL:
+        _log.debug('Find users by e-mail {}'.format(SAR_EMAIL))
+        users = users + session.query(db.User).filter(db.User.email == SAR_EMAIL).all()
```
jeremycline: You can write this as ``db.User.query..filter(db.User.email == SAR_EMAIL).all()`` or even ``db.User.query.filter_by(email=SAR_EMAIL).all()``.

anitya/db/models.py
```diff
@@ -954,6 +954,31 @@ def get_id(self):
         """
         return six.text_type(self.id)
 
+    def __json__(self):
```
jeremycline: I know this is copied from other model, but I do not think the name of this method conveys what it does. It's not a "real" dunder method, nor is it actually returning JSON. ``as_dict`` or ``to_dict`` or something is clearer.

anitya/tests/test_sar.py
```diff
@@ -0,0 +1,130 @@
...
+        )
+        self.session.add(user2)
+        self.session.commit()
+
+        sar.SAR_EMAIL = 'user@fedoraproject.org'
+        sar.main()
+
+        out, err = self.capsys.readouterr()
+
+        self.assertTrue('"username": "' + user.username + '"' in out)
```
jeremycline: Using a format string here would make it easier to read:
```suggestion
        self.assertTrue('"username": "{}"'.format(user.username) in out)
```

anitya/sar.py
Zlopez: Didn't know you can write it like this. It is more readable. I will change it.

anitya/db/models.py
Zlopez: Yeah, you are right.

anitya/tests/test_sar.py
Zlopez: I will change this.

Zlopez:
@bowlofeggs is this looking good to you now?

anitya/tests/test_sar.py
```diff
@@ -0,0 +1,130 @@
...
+        )
+        self.session.add(user2)
+        self.session.commit()
+
+        sar.main()
+
+        out, err = self.capsys.readouterr()
+        print(out)
+
+        self.assertTrue('' in out)
```
bowlofeggs: This will be ```True``` for any ```str```:

```
>>> '' in 'hello'
True
```

I recommend asserting that the output is exactly the JSON string that is expected (would that be ```[]```?)

anitya/tests/test_sar.py
```diff
@@ -0,0 +1,130 @@
...
+
+        sar.SAR_USERNAME = 'user'
+        sar.main()
+
+        out, err = self.capsys.readouterr()
+
+        self.assertTrue('"username": "{}"'.format(user.username) in out)
+        self.assertTrue('"email": "{}"'.format(user.email) in out)
+        self.assertFalse('"username": "{}"'.format(user2.username) in out)
+        self.assertFalse('"email": "{}"'.format(user2.email) in out)
```
bowlofeggs: Rather than spot checking strings, why not load the output with ```json.loads()``` and asserting that it is exactly the expected structure, including all keys?

anitya/tests/test_sar.py
```diff
@@ -0,0 +1,130 @@
...
+        self.session.add(user2)
+        self.session.commit()
+
+        sar.SAR_EMAIL = 'user@fedoraproject.org'
+        sar.main()
+
+        out, err = self.capsys.readouterr()
+
+        self.assertTrue('"username": "{}"'.format(user.username) in out)
+        self.assertTrue('"email": "{}"'.format(user.email) in out)
```
bowlofeggs: Similarly here, it would be good to assert that the output is valid JSON and has all the expected keys.

anitya/sar.py
```diff
@@ -0,0 +1,74 @@
...
+import json
+import sys
+
+from anitya.config import config
+from anitya import db
+
+_log = logging.getLogger('anitya')
+
+SAR_USERNAME = os.getenv('SAR_USERNAME')
+SAR_EMAIL = os.getenv('SAR_EMAIL')
```
bowlofeggs: It would be easier to test this if these lines were part of main. Then you can use ```mock``` to set the environment variables and assert that these lines use ```os.getenv()``` correctly.

anitya/tests/test_sar.py
Zlopez: Thanks for noticing. I will change this to work with json.

anitya/tests/test_sar.py
Zlopez: Copied this from some other test, but you are right, it will be better to check against expected JSON. I will change this.

anitya/sar.py
Zlopez: Yes, you are right. Thanks for noticing.

Zlopez:
@bowlofeggs I addressed your changes and did a squash,

Zlopez:
Rebased with latest master

anitya/tests/test_sar.py
```diff
@@ -0,0 +1,156 @@
...
+        exp = [{
+            'id': str(user.id),
+            'username': user.username,
+            'email': user.email,
+            'active': user.active,
+            'social_auth': []
+        }]
+
+        with mock.patch('os.getenv') as mock_getenv:
+            mock_getenv.side_effect = mock_getenv_email
```
bowlofeggs: Why not use [```mock.patch.dict()```](https://docs.python.org/3/library/unittest.mock.html#unittest.mock.patch.dict) here instead? It's less code for you to maintain and is clearer for readers.

anitya/tests/test_sar.py
```diff
@@ -0,0 +1,156 @@
...
+        exp = [{
+            'id': str(user.id),
+            'username': user.username,
+            'email': user.email,
+            'active': user.active,
+            'social_auth': []
+            }
+        ]
+
+        sar.SAR_USERNAME = 'user'
```
bowlofeggs: This line doesn't seem needed anymore.

anitya/tests/test_sar.py
Zlopez: Didn't thought about this. You are right, it will be clearer.

anitya/tests/test_sar.py
Zlopez: Probably missed it.

Zlopez:
I fixed the conflict and addressed the last suggestions.
There are still failing tests, but these will be fixed by https://github.com/release-monitoring/anitya/pull/667

bowlofeggs:
Cool, let me know once tests pass.

Zlopez:
@bowlofeggs 
Tests are working again. \o/

PR #651, opened 2018-10-29
Add supported versions of python to setup script

Zlopez:


setup.py
```diff
@@ -62,6 +62,12 @@ def get_requirements(requirements_file='requirements.txt'):
     license='GPLv2+',
     download_url='https://fedorahosted.org/releases/a/n/anitya/',
     url='https://fedorahosted.org/anitya/',
+    classifiers=[
+        "Programming Language :: Python :: 2.7",
```
jeremycline: While Python 2.7 may still work, I wouldn't try to support it since we run 3 in production. I'd say supporting Python 3.6+ is reasonable (unless there's a nice API from 3.7 you want to use)

Zlopez: You are right, the Python 2.7 shouldn't be there.

PR #652, opened 2018-10-30
Change prefix handling

### LINKED ISSUE #644: GitLab backend - version prefix not working
For some reason if I add a version prefix to a GitLab source I'm getting the full name regardless.

This is the GitLab repo:
https://gitlab.inria.fr/flocq/flocq/tags

The project name is flock:
https://release-monitoring.org/project/7622/

Zlopez:
Fixes #644.

### DIFF
--- a/anitya/lib/utilities.py
+++ b/anitya/lib/utilities.py
@@ -64,12 +64,11 @@ def check_project_release(project, session, test=False):
             'No backend was found for "%s"' % project.backend)
 
     publish = False
-    versions = []
-    max_version = None
-    old_version = None
+    versions_prefix = []
+    max_version = ''
 
     try:
-        versions = backend.get_versions(project)
+        versions_prefix = backend.get_versions(project)
     except exceptions.AnityaPluginException as err:
         _log.exception("AnityaError catched:")
         project.logs = str(err)
@@ -77,28 +76,32 @@ def check_project_release(project, session, test=False):
         session.commit()
         raise
 
+    # Remove prefix
+    version_class = project.get_version_class()
+    versions = [version_class(
+        version=v, prefix=project.version_prefix) for v in versions_prefix]
+
     if test:
-        return versions
+        return [v.parse() for v in versions]
 
     # There is always at least one version retrieved,
     # otherwise this backend raises exception
     project.logs = 'Version retrieved correctly'
 
     p_versions = project.get_sorted_version_objects()
-    if p_versions:
-        old_version = p_versions[0].version
-
+    old_version = project.latest_version or ''
     for version in versions:
-        if version not in project.versions:
+        if version not in p_versions:
             project.versions_obj.append(
                 models.ProjectVersion(
                     project_id=project.id,
-                    version=version
+                    version=version.version
                 )
             )
 
     sorted_versions = project.get_sorted_version_objects()
-    max_version = sorted_versions[0].version
+    if sorted_versions:
+        max_version = sorted_versions[0].version
     if project.latest_version != max_version:
         project.latest_version = max_version
         publish = True
--- a/anitya/lib/backends/__init__.py
+++ b/anitya/lib/backends/__init__.py
@@ -214,7 +214,8 @@ class BaseBackend(object):
         '''
         vlist = self.get_versions(project)
         version_class = project.get_version_class()
-        sorted_versions = sorted([version_class(version=v) for v in vlist])
+        sorted_versions = sorted(
+            [version_class(version=v, prefix=project.version_prefix) for v in vlist])
         return [v.version for v in sorted_versions]
 
     @classmethod
@@ -319,10 +320,6 @@ def get_versions_by_regex_for_text(text, url, regex, project):
         if type(version) == tuple:
             version = ".".join([v for v in version if not v == ""])
 
-        # Strip the version_prefix early
-        if project.version_prefix is not None and \
-                version.startswith(project.version_prefix):
-            version = version[len(project.version_prefix):]
         upstream_versions[index] = version
 
         if " " in version:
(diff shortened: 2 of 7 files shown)


anitya/lib/backends/gitlab.py
```diff
@@ -101,7 +101,13 @@ def get_versions(cls, project):
 
         tags = []
         for tag in json:
-            tags.append(tag["name"])
+            name = tag["name"]
```
jeremycline: The version API (used by the ``Project.versions`` property) should handle both sorting and stripping prefixes. If the UI isn't displaying correctly-stripped versions, it should use that property to get the list, I think.

Zlopez: I don't think the version API is striping the prefix, because I saw this in few backends already.

jeremycline: See https://github.com/release-monitoring/anitya/blob/4ebc10295b985d78770ce068f92f2164802ef822/anitya/lib/versions/base.py#L45

Zlopez: In this case than something is wrong with this. I need to check if this works or what happened.

Zlopez: Ok, this issue is more complex than I thought. When we are doing check for new versions, we are calling get_versions directly on backend. This works on most of the backends, because they are doing the prefix handling.
I need to refactor this a little to make it working. Probably applying project prefix on retrieved versions before comparing them with existing versions.

Zlopez: I found out why this issue wasn't found before. Because every backend used regex to obtain new versions and function `get_version_by_regex` (see https://github.com/release-monitoring/anitya/blob/master/anitya/lib/backends/__init__.py#L284) is erasing the prefix. Github and Gitlab are using APIs instead of regexes.
I think the regex shouldn't be erased here, but rather in `check_project_release` or it shoudn't be removed at all and left this on the `parse` method in version plugin.

There are two options right now:
1) We could remove the prefix before saving the version to database - best place is `check_project_release`
2) We shouldn't remove the prefix at all - but this could cause issue, if the prefix is changed in the future
2) We

Zlopez:
I changed the current commit, because previous change was thrown away.
I decided to go with first option:
We could remove the prefix before saving the version to database - best place is check_project_release

jeremycline:
> We could remove the prefix before saving the version to database - best place is check_project_release

In my opinion, we shouldn't do this. This makes comparing the list of versions we have from the list of versions we find on a poll super simple. Dealing with prefixes is necessary for version sorting and tidy presentation to the user, but I think that's a separate concern.

Zlopez:
I'm not sure If I understand you. You mean, that we should let the version be the same how we obtained it?
But if somebody changes the prefix, we couldn't possibly compare them.
For example:
version-0.10 has prefix_version `version-`, but somebody in the upstream project will change the version prefix to something like v-0.12 and we are screwed, we change the version_prefix to `v-`, but every already retrieved versions will be uncomparable to new ones, because the project has only one prefix.

But when I'm thinking about it, we have the same issue when retrieving more than one version. We can possibly compare them if the prefix changes throughout the project.

Maybe the correct way to solve this is to allow multiple prefixes.

But still for the comparison of newly retrieved versions we need to first remove the prefix using version plugin asociated with project and than compare them with any existing versions, like I'm doing it here https://github.com/release-monitoring/anitya/blob/a7178fbbe1d35862291274d3bfad45c6024969a4/anitya/lib/utilities.py#L81.
But we could save it in the DB with prefix when we are done with comparison.

Zlopez:
Ok, I changed how the prefix is handled. When new release is obtained we temporary create a version object, so we can compare new versions with existing ones. But we are saving to DB the original version we obtained.

Example:
Project `foo` has version `version-0.1.0` and releases a new version `version-0.2.0`. Anitya obtains this new version and creates version object from it (RPM in this case) than calls `get_ordered_version_objects()` and obtains already saved versions. It finds out this version isn't saved in project and adds it to project `foo` as `version-0.2.0`.
The prefix is handled only by version object.

I hope this is correct from your point of view.

P.S.: I added the new [issue](https://github.com/release-monitoring/anitya/issues/655) for supporting multiple prefixes

jeremycline:
> But when I'm thinking about it, we have the same issue when retrieving more than one version. We can't possibly compare them if the prefix changes throughout the project.

Sure, but I also think this is optimizing for the wrong thing. People will do _crazy_ things with their versions and you won't be able to handle every case. What you should handle well is reasonable things, and not explode on crazy things. The point of release-monitoring is to send out notifications on new releases, the history is much less important (or completely useless).  It _used_ to be important when it only tried to announce the latest version by its arbitrary sorting algorithm, but we shouldn't do that anymore.

The reason, I think, the original plan was to not strip the version prefix was that people often need to fiddle with the prefix after they've written their regular expression.  The details escape me, and it's quite possible I'm mis-remembering.

The reason I spent any time thinking about version schemes and sorting was to make the downstream stuff (the-new-hotness) able to be smart about the pull requests it opened. If a project uses semantic versioning and the X version is bumped, it should only make a PR against Rawhide. If it's a Y or Z bump, it's likely it can go into stable releases. It could also map to module streams, but I have no idea what the current scheme is for those (if there even is one). If a project is doing something silly, just do a sane default (Rawhide only) and let the maintainer figure it out.

anitya/lib/utilities.py
```diff
@@ -64,41 +64,44 @@ def check_project_release(project, session, test=False):
...
+        versions_prefix = backend.get_versions(project)
     except exceptions.AnityaPluginException as err:
         _log.exception("AnityaError catched:")
         project.logs = str(err)
         session.add(project)
         session.commit()
         raise
 
+    # Remove prefix
+    version_class = project.get_version_class()
```
jeremycline: This code is getting copied and pasted around, but it should just be a property or method on the Project model. ``project.versions`` should really be this with a ``return [str(v) for v in versions]`` so it returns a list of strings with parsing applied.

Zlopez: Yeah, but this is applied before the versions are added to the project.

PR #653, opened 2018-11-05
Upgrade vagrant to Fedora 29

Zlopez:
Some things were changed:
* virtualenv-3 is now virtualenv
* sphinx version restriction was removed - 1.5.6 not working on python
3.7
* run alembic with heads instead of head, this will fix issue when
more migrations should be applied at once

### DIFF
--- a/ansible/roles/anitya-dev/tasks/main.yml
+++ b/ansible/roles/anitya-dev/tasks/main.yml
@@ -37,7 +37,7 @@
   become_user: "{{ ansible_env.SUDO_USER }}"
   shell: >
       source ~/.bashrc && mkvirtualenv anitya -a ~/devel/ &&
-      virtualenv-3 --system-site-packages ~/.virtualenvs/anitya/
+      virtualenv --python=/usr/bin/python3 --system-site-packages ~/.virtualenvs/anitya/
   args:
     creates: /home/{{ ansible_env.SUDO_USER }}/.virtualenvs/anitya
 
--- a/Vagrantfile.example
+++ b/Vagrantfile.example
@@ -4,7 +4,7 @@
 VAGRANTFILE_API_VERSION = "2"
 
 Vagrant.configure(VAGRANTFILE_API_VERSION) do |config|
- config.vm.box = "fedora/28-cloud-base"
+ config.vm.box = "fedora/29-cloud-base"
 
  # Forward traffic on the host to the development server on the guest
  config.vm.network "forwarded_port", guest: 5000, host: 5000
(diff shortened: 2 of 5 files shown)


ansible/roles/anitya-dev/tasks/db.yml
```diff
@@ -61,7 +61,7 @@
 
 - name: Apply database migrations
   become_user: "{{ ansible_env.SUDO_USER }}"
-  command: ~/.virtualenvs/anitya/bin/alembic upgrade head
+  command: ~/.virtualenvs/anitya/bin/alembic upgrade heads
```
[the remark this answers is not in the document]
Zlopez: I need to look how is this checked in Bodhi. This was the first solution I came up with.

bowlofeggs:
On Wed, 2018-11-07 at 01:24 -0800, Zlopez wrote:
> Couldn't find where this is checked in Bodhi

See https://github.com/fedora-infra/bodhi/blob/3.10.1/bodhi/tests/server/test_alembic.py#L32-L48

Zlopez:
This is the reason I didn't found it. I thought it is some check in pull request, but if this is only a test, I will add it to this PR.

bowlofeggs:
On Wed, 2018-11-07 at 19:22 +0000, Zlopez wrote:
> I thought it is some check in pull request, but if this is only a
> test, I will add it to this PR.

It would probably be more sensible to make it a PR check rather than a
unit test, but that's how it is in Bodhi for now. Now that Bodhi does
have separate checks, perhaps it would be good for me to split that out
into a separate type of test…

ansible/roles/anitya-dev/tasks/main.yml
```diff
@@ -37,7 +37,7 @@
   become_user: "{{ ansible_env.SUDO_USER }}"
   shell: >
       source ~/.bashrc && mkvirtualenv anitya -a ~/devel/ &&
-      virtualenv-3 --system-site-packages ~/.virtualenvs/anitya/
+      virtualenv --system-site-packages ~/.virtualenvs/anitya/
```
bowlofeggs: What is the purpose of this change?

bowlofeggs: It would be good to be explicit about using Python 3:

```suggestion
      virtualenv --python=/usr/bin/python3 --system-site-packages ~/.virtualenvs/anitya/
```

Zlopez:
I added the test with last commit, but we should first merge https://github.com/release-monitoring/anitya/pull/654 to fix the failing test.

ansible/roles/anitya-dev/tasks/main.yml
bowlofeggs: Since this was explicitly Python 3 before, it would be good to remain explicit:

```suggestion
      virtualenv --python=/usr/bin/python3 --system-site-packages ~/.virtualenvs/anitya/
```

Zlopez: Applied

PR #654, opened 2018-11-06
Fix: Multiple heads in alembic

Zlopez:
Fix for the current alembic issue when applying migrations.

I created a merge migration to fix this.


Zlopez:
Merged to fix #653

PR #656, opened 2018-11-09
Remove duplicate code for versions

Zlopez:


Zlopez:
Rebased and fixed the failing test. 

It turns out that this PR is solving also the issue with reversed Github version list. So I removed the reversion.

Zlopez:
Merging manually, already approved by @cverna

PR #658, opened 2018-11-15
Fix broken test

Zlopez:


Zlopez:
I will merge this immediately, because it makes every other PR fail.

PR #659, opened 2018-11-16
Add basic user management

Zlopez:
Changes:
* Adds new page `/users` which is similar to `/flags` page.
  This user management will help promote new admins or ban users.
screenshot from 2018-11-16 11-22-40
* Add `admin` flag to `User` model.
* User property `admin` is renamed to `is_admin`.
* Admin flag for preconfigured users is set on `is_admin` call.

### DIFF
--- a/anitya/admin.py
+++ b/anitya/admin.py
@@ -23,7 +23,7 @@ def is_admin(user=None):
     '''
     user = user or flask.g.user
     if user.is_authenticated:
-        return user.admin
+        return user.is_admin
 
 
 @ui_blueprint.route('/distro/<distro_name>/edit', methods=['GET', 'POST'])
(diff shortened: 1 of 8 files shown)


anitya/admin.py
```diff
@@ -464,3 +464,194 @@ def set_flag_state(flag_id, state):
     return flask.redirect(
         flask.url_for('anitya_ui.browse_flags')
     )
+
+
+@ui_blueprint.route('/users', methods=['GET'])
+@login_required
+def browse_users():
+
+    if not is_admin():
```
cverna: It would be nice to refactor this code in a decorator, similar to how `@login_required` works. But this should be done in another PR.

anitya/admin.py
```diff
@@ -464,3 +464,194 @@ def set_flag_state(flag_id, state):
...
+
+
+@ui_blueprint.route('/users/<user_id>/admin/<state>', methods=['POST'])
+@login_required
+def set_user_admin_state(user_id, state):
+
+    if not is_admin():
+        flask.abort(401)
+
+    if state.upper() == "TRUE":
```
cverna: It might be better to use a form here with some boolean fields , what do you think ?

anitya/admin.py
```diff
@@ -464,3 +464,194 @@ def set_flag_state(flag_id, state):
...
+
+
+@ui_blueprint.route('/users/<user_id>/active/<state>', methods=['POST'])
+@login_required
+def set_user_active_state(user_id, state):
+
+    if not is_admin():
+        flask.abort(401)
+
+    if state.upper() == "TRUE":
```
cverna: Same as above maybe using a form would make sense here

anitya/db/migrations/versions/b13662e5d288_add_admin_flag.py
```diff
@@ -0,0 +1,30 @@
...
+
+def upgrade():
+    """ Add 'admin' flag to users table. """
+    op.add_column(
+        'users',
+        sa.Column(
+            'admin',
+            sa.Boolean,
+            default=False)
+    )
```
cverna: This will remove the admin permission to the users that are currently admin. You might want to make sure that the current admins have the value set to `True` during the migration.

anitya/db/models.py
```diff
@@ -896,16 +898,21 @@ class User(Base):
...
-        Determine if this user is an administrator.
+        Determine if this user is an administrator. Set admin flag
+        if the user is preconfigured.
 
         Returns:
             bool: True if the user is an administrator.
         """
-        return six.text_type(self.id) in anitya_config.get('ANITYA_WEB_ADMINS', [])
+        if not self.admin:
+            if six.text_type(self.id) in anitya_config.get('ANITYA_WEB_ADMINS', []):
```
cverna: I think that it would be good to drop support for this config key. That way there would be only one place where you can manage admins.

anitya/templates/users.html
```diff
@@ -0,0 +1,141 @@
...
+        <td>{{ user.email }}</td>
+        <td>{{ user.is_admin }}</td>
+        <td>{{ user.is_active }}</td>
+        <td>
+            {% if user.is_admin %}
+            <form method="POST" action="{{ url_for('anitya_ui.set_user_admin_state',
+                user_id=user.id, state=False) }}">
+                <button type="submit" class="btn btn-danger btn-sm pull-right inline"
+                        id="user_btn">
+                    Revoke admin
```
cverna:
```suggestion
                    Remove admin permissions
```

anitya/templates/users.html
```diff
@@ -0,0 +1,141 @@
...
+                </button>
+                {% else %}
+                <form method="POST" action="{{ url_for('anitya_ui.set_user_admin_state',
+                user_id=user.id, state=True) }}">
+                    <button type="submit" class="btn btn-success btn-sm pull-right inline"
+                            id="user_btn">
+                        Promote admin
```
cverna:
```suggestion
                    Give admin permissions
```

anitya/templates/users.html
```diff
@@ -0,0 +1,141 @@
...
+                <button type="submit" class="btn btn-danger btn-sm pull-right inline"
+                        id="user_btn">
+                    Ban
+                </button>
+                {% else %}
+                <form method="POST" action="{{ url_for('anitya_ui.set_user_active_state',
+                user_id=user.id, state=True) }}">
+                    <button type="submit" class="btn btn-success btn-sm pull-right inline"
+                            id="user_btn">
+                        Remove ban
```
cverna:
```suggestion
                    Remove ban
```

anitya/tests/test_flask_admin.py
```diff
@@ -803,3 +805,563 @@ def test_set_flag(self):
...
+
+    def test_missing_state(self):
+        """Assert an missing state results in HTTP 404."""
+        with login_user(self.flask_app, self.admin):
+            output = self.client.post('/users/{}/admin/'.format(
+                six.text_type(self.user.id)
+            ))
+            self.assertEqual(404, output.status_code)
+
+    def test_missing(self):
```
cverna:
```suggestion
    def test_missing_user(self):
```

anitya/tests/test_flask_admin.py
```diff
@@ -803,3 +805,563 @@ def test_set_flag(self):
...
+
+    def test_missing_state(self):
+        """Assert an missing state results in HTTP 404."""
+        with login_user(self.flask_app, self.admin):
+            output = self.client.post('/users/{}/active/'.format(
+                six.text_type(self.user.id)
+            ))
+            self.assertEqual(404, output.status_code)
+
+    def test_missing(self):
```
cverna:
```suggestion
    def test_missing_user(self):
```

anitya/admin.py
Zlopez: This could be nice. I will add it to my TODO list.

anitya/admin.py
Zlopez: There is a form already on the page, but this change is done by button which sets the state value.

anitya/db/migrations/versions/b13662e5d288_add_admin_flag.py
Zlopez: Current admins are defined in the configuration file, so this shouldn't change anything.

anitya/db/models.py
Zlopez: The admins set in the configuration file shouldn't be stripped of this privilege. This is failsafe against some admin going rampage and revoking rights for everyone else and yourself. Also the configuration file could be changed in ansible, so you can easily remove or add admin even if you don't have the admin rights in the applications itself.

anitya/templates/users.html
Zlopez: I need to check, how this will actually look on the page first.

anitya/admin.py
cverna: Ack

anitya/db/models.py
cverna: Ok, I was just concerned that with time it will  not be obvious where the admins are managed (ie config or in the db).

anitya/admin.py
Zlopez: Same as above

Zlopez:
@cverna 
Addressed your suggestions and rebased.

PR #663, opened 2018-11-20
Show version url on project page

Zlopez:
* Introduce new method to Project model
* Introduce new method to backend plugins

### DIFF
--- a/anitya/lib/backends/bitbucket.py
+++ b/anitya/lib/backends/bitbucket.py
@@ -45,6 +45,36 @@ class BitBucketBackend(BaseBackend):
         return cls.get_ordered_versions(project)[-1]
 
     @classmethod
+    def get_version_url(cls, project):
+        ''' Method called to retrieve the url used to check for new version
+        of the project provided, project that relies on the backend of this plugin.
+
+        Attributes:
+            project (:obj:`anitya.db.models.Project`): Project object whose backend
+                corresponds to the current plugin.
+
+        Returns:
+            str: url used for version checking
+        '''
+        url_template = 'https://bitbucket.org/%(version_url)s/' \
+                       'downloads?tab=tags'
+        url = ''
+        if project.version_url:
+            url = project.version_url.replace(
+                'https://bitbucket.org/', '')
+        elif project.homepage.startswith('https://bitbucket.org'):
+            url = project.homepage.replace(
+                'https://bitbucket.org/', '')
+
+        if url.endswith('/'):
+            url = url[:-1]
+
+        if url:
+            url = url_template % {'version_url': url}
+
+        return url
+
+    @classmethod
     def get_versions(cls, project):
         ''' Method called to retrieve all the versions (that can be found)
         of the projects provided, project that relies on the backend of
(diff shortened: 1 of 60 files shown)


Zlopez:
Fix conflict

Zlopez:
I noticed, that this is also fixing issue with exception raised in gitlab backend, which caused the whole cron job to fail. I will add a new news file for towncryer before merging.

anitya/lib/backends/bitbucket.py
```diff
@@ -44,6 +44,36 @@ def get_version(cls, project):
...
+                       'downloads?tab=tags'
+        url = ''
+        if project.version_url:
+            url = project.version_url.replace(
+                'https://bitbucket.org/', '')
+        elif project.homepage.startswith('https://bitbucket.org'):
+            url = project.homepage.replace(
+                'https://bitbucket.org/', '')
+
+        if url.endswith('/'):
```
cverna: you can use `url.strip("/")` here instead of the if logic

anitya/lib/backends/cpan.py
```diff
@@ -48,6 +48,23 @@ def get_version(cls, project):
...
+        of the project provided, project that relies on the backend of this plugin.
+
+        Attributes:
+            project (:obj:`anitya.db.models.Project`): Project object whose backend
+                corresponds to the current plugin.
+
+        Returns:
+            str: url used for version checking
+        '''
+        url = 'https://metacpan.org/release/%(name)s/' % {
```
cverna: I think it would be nice to have some consistency in the use of the `%` format operator and `format()` method. Also if anitya is Python 3 only and runs on Python 3.6 f-strings are nice.

anitya/lib/backends/custom.py
```diff
@@ -49,6 +49,22 @@ def get_version(cls, project):
...
+        Attributes:
+            project (:obj:`anitya.db.models.Project`): Project object whose backend
+                corresponds to the current plugin.
+
+        Returns:
+            str: url used for version checking
+        '''
+        url = project.version_url
+
+        return url
```
cverna:
```suggestion
        return project.version_url
```

anitya/lib/backends/custom.py
```diff
@@ -49,6 +49,22 @@ def get_version(cls, project):
...
+        of the project provided, project that relies on the backend of this plugin.
+
+        Attributes:
+            project (:obj:`anitya.db.models.Project`): Project object whose backend
+                corresponds to the current plugin.
+
+        Returns:
+            str: url used for version checking
+        '''
+        url = project.version_url
```
cverna:
```suggestion
```

anitya/lib/backends/github.py
```diff
@@ -59,6 +59,32 @@ def get_version(cls, project):
...
+        Returns:
+            str: url used for version checking
+        '''
+        url = ''
+        if project.version_url:
+            url = project.version_url
+        elif project.homepage.startswith('https://github.com'):
+            url = project.homepage.replace('https://github.com/', '')
+
+        if url.endswith('/'):
```
cverna: maybe use `url = url.strip("/")` and remove the `if` statement

PR #664, opened 2018-11-21
Fix: Latest version contains prefix

### LINKED ISSUE #662: Latest version on Project UI is shown with prefix
After introduction of new prefix handling and retrieval of multiple versions at once. I noticed, that the latest_version attribute on project is filled with original retrieved version with prefix instead of prefix free version.

Zlopez:
Fixes #662 


Zlopez:
Fix conflict

PR #670, opened 2018-12-03
Refresh page after full check

Zlopez:

### DIFF
--- a/anitya/templates/project.html
+++ b/anitya/templates/project.html
@@ -281,6 +281,7 @@
               output = res.error;
             }
             $('#version_info').html("Latest versions: " + output);
+            window.location.reload(false);
           }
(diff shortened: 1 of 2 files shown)


anitya/templates/project.html
```diff
@@ -281,6 +281,7 @@ <h3 class="panel-title">
               output = res.error;
             }
             $('#version_info').html("Latest versions: " + output);
+            window.location.reload(false);
```
cverna: Just out curiosity why is this needed ?

Zlopez: This is a button on the page. If you click on it, it will do a regular check for new versions. This could change project in the database, but the page will still have old information. It's confusing for the user that used this button.

Issue #674, opened 2018-12-07
Crash when version is too long


Zlopez:
Today I checked the output of cron job in anitya and it crashed on this project:
https://github.com/jd/daiquiri/tags

There are tags, that are too long and probably not even version.
To prevent crashing in the future, I will check the length of version string and will not save it if it's longer then column length in database.

PR #677, opened 2018-12-10
Fix crash when version is too long

Zlopez:
Fixes https://github.com/release-monitoring/anitya/issues/674

### DIFF
--- a/anitya/lib/utilities.py
+++ b/anitya/lib/utilities.py
@@ -103,14 +103,16 @@ def check_project_release(project, session, test=False):
 
     p_versions = project.get_sorted_version_objects()
     old_version = project.latest_version or ''
+    version_column_len = models.ProjectVersion.version.property.columns[0].type.length
     for version in versions:
         if version not in p_versions:
-            project.versions_obj.append(
-                models.ProjectVersion(
-                    project_id=project.id,
-                    version=version.version
+            if len(version.version) < version_column_len:
+                project.versions_obj.append(
+                    models.ProjectVersion(
+                        project_id=project.id,
+                        version=version.version
+                    )
                 )
-            )
 
     sorted_versions = project.get_sorted_version_objects()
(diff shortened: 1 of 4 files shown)


anitya/lib/utilities.py
```diff
@@ -103,14 +103,16 @@ def check_project_release(project, session, test=False):
...
     p_versions = project.get_sorted_version_objects()
     old_version = project.latest_version or ''
+    version_column_len = models.ProjectVersion.version.property.columns[0].type.length
     for version in versions:
         if version not in p_versions:
-            project.versions_obj.append(
-                models.ProjectVersion(
-                    project_id=project.id,
-                    version=version.version
+            if len(version.version) < version_column_len:
```
cverna: Should we log something when the version is too long ?

anitya/lib/utilities.py
```diff
@@ -103,14 +103,16 @@ def check_project_release(project, session, test=False):
...
+                project.versions_obj.append(
+                    models.ProjectVersion(
+                        project_id=project.id,
+                        version=version.version
```
cverna: The current column size is 50 char any reason for not making this bigger ? maybe 100 ?

anitya/lib/utilities.py
Zlopez: Yeah, we should. I will add a info level log message.

anitya/lib/utilities.py
Zlopez: It's enough for most projects, I only saw one project where this limit is not enough and this looks like mistake not a correct version. See https://github.com/jd/daiquiri/tags

Zlopez: If this limit will be an issue in future we can update it, but right now it is enough.

cverna: ok :+1:

PR #679, opened 2018-12-12
Fix vagrant up

jeremycline:
virtualenv is not in the vagrant box. Be explicit and use virtualenv-3.

Signed-off-by: Jeremy Cline <jcline@redhat.com>


Zlopez:
Which version of the fedora are you using in vagrant? Because virtualenv-3 is not available in Fedora 29. This is why I removed it. Otherwise I'm getting this:
```
/bin/sh: virtualenv-3: command not found
```

PR #681, opened 2018-12-13
Revert "Fix vagrant up"

Zlopez:
Reverts release-monitoring/anitya#679


Zlopez:
I want to revert this change, because I can't do `vagrant provision` without getting error.

Issue #691, opened 2019-01-07
pagure backend doesn't strip prefix


abitrolly:
https://release-monitoring.org/project/12208/

![image](https://user-images.githubusercontent.com/8781107/50789765-2f70df00-12ce-11e9-8ed7-4c6aa7c9b545.png)

It might be that it breaks notifications about new releases.

Zlopez:
This should be handled differently in current master. The prefix is now handled by `Version` object and notification is sent without prefix.
Previously the prefix was stripped before the new version was stored in database, which lead to duplication. Now the prefix is stored to database and it is only stripped when shown on the frontend or send out.

abitrolly:
What do you propose in this case?

Zlopez:
This is how it looks in current master
![screenshot from 2019-01-08 09-30-40](https://user-images.githubusercontent.com/6943409/50818696-185edb00-1328-11e9-864e-751e1c3426ac.png)

Zlopez:
There is one issue however, I can't recognize the wrong versions anymore (ones without prefix). I will probably add raw version column for admins, because only admins have ability to remove the versions.

Zlopez:
This is how admin will see it after the change will be merged. I think this is enough to recognize the wrong versions.
![screenshot from 2019-01-08 09-39-43](https://user-images.githubusercontent.com/6943409/50819109-6aecc700-1329-11e9-848a-0565b22373d3.png)

abitrolly:
Removing history is not good. There is some value in the data about when software was released and seen by monitoring.

The error is likely caused by inability to detect latest version from raw `pagure` version.

![image](https://user-images.githubusercontent.com/8781107/50827672-304a5500-1350-11e9-91eb-e6224bd0e818.png)

abitrolly:
There are many projects from pagure with this bug.

Zlopez:
@abitrolly This is not about removing history, but about removing duplicates. You will still have the history intact.

Zlopez:
This behavior is changed in 0.14.0. There is no odd version anymore. You either get new version or not.

Zlopez:
I checked the project in latest version and it's showing correctly. Closing.

abitrolly:
It is good that history is preserved.

PR #696, opened 2019-01-08
Show raw version on project page

Zlopez:
This will allow admins to recognize versions with and without prefix
when removing versions from project.

screenshot from 2019-01-08 09-39-43


Zlopez:
This PR resolves the issue found in https://github.com/release-monitoring/anitya/issues/691

PR #705, opened 2019-01-15
Fix: Comparing by dates created version duplicates

### LINKED ISSUE #702: Cron job is crashing because of project anyterm
Project anyterm (https://release-monitoring.org/project/8330/) is causing crash of cron job when checking for latest version when it is already retrieved.

This is in the cron job log:
```
sqlalchemy.orm.exc.FlushError: New instance <ProjectVersion at 0x7f27714f3e10> with identity key (<class 'anitya.db.models.ProjectVersion'>, (8330, '1.2.3'), None) conflicts with persistent instance <ProjectVersion at 0x7f2771107f60>
```

This looks like it can't recognize that the version was already retrieved.
[anitya-1547478000-nhgqd.log](https://github.com/release-monitoring/anitya/files/2755778/anitya-1547478000-nhgqd.log)

Zlopez:
Date version scheme compared for equality only by dates, this caused
creation of duplicities, which led to crash when commiting changes.
This commit changes the way how the date version scheme is comparing for
equality. Now it is calling `__eq__` method in Version class. Versions are
now considered equal if they are same after prefix is stripped.

Fixes #702 

### DIFF
--- a/anitya/lib/versions/base.py
+++ b/anitya/lib/versions/base.py
@@ -74,8 +74,11 @@ class Version(object):
             version = self.version[len(self.prefix):]
 
         # Many projects prefix their tags with 'v', so strip it if it's present
-        if v_prefix.match(version):
-            version = version[1:]
+        try:
+            if v_prefix.match(version):
+                version = version[1:]
+        except TypeError:
+            raise InvalidVersion(version)
 
         return version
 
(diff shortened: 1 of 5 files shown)


jeremycline:
When we made the versions code, comparing by date found was never the plan and this seems fraught with problems. It's intended for versioning schemes used by projects like semantic versioning (https://semver.org/), calendar (https://calver.org/), etc. In my opinion, the date version should be reverted.

Zlopez:
It was created to solve some of the issues, that were already there, but you are probably right. However I need to create migration to change any project, that is using date versioning scheme back to rpm.

Issue #714, opened 2019-01-30
Javascript error on add project page


Zlopez:
# Issue
Javascript is not working on https://release-monitoring.org/project/new, because of the syntax error in javascript.
```
SyntaxError: unexpected token: identifier
```
This is related to this [line](https://github.com/release-monitoring/anitya/blob/master/anitya/templates/project_new.html#L150), which is missing the `;`. Also it looks like the jinja is removing `\n`, so the next line is actually appended to previous line.

# Solution
Add missing `;`
Add empty line before jinja `if` statement

yarons:
Can we close this? @Zlopez

nqb:
Any workaround to add a project before 0.15 released on release-monitoring.org ?

Zlopez:
Everything should work, but the fields aren't hide or shown when backend is changed.

So you can add new projects, but you'll see more fields than we actually use.

In the meantime you can look at the [staging](https://stg.release-monitoring.org/) instance of Anitya and see what fields are currently used there. The staging is running on 0.15.0 already. Look at the id of the fields.

PR #717, opened 2019-01-30
Added semicolon

yarons:
Solved #714 
I don't know anything about the code but it looks like a rather easy task so why not actually :smile:


Zlopez:
Thanks for PR \o/
Could you also do few more things:
* Add semicolon to other lines where it's missing
* Add empty line before jinja if statement
* Add news file to `news` folder - https://anitya.readthedocs.io/en/latest/contributing.html#release-notes

Zlopez:
Yes, you have it right, but I recommend to use the name of the issue as content.
The file should be `714.bug` with content:
`Javascript error on add project page`
So we know what was actually fixed in the changelog.

Author files are created automatically from commit authors.

The example in the documentation will be nice.

Zlopez:
I will merge this after #720 which is fixing the tests.

PR #719, opened 2019-01-31
Added example of usage

yarons:
As proposed by @Zlopez 

Thanks!


Zlopez:
I noticed you approved my change, so now you have two black lines above and bellow the example. :-D Could you fix this.

Zlopez:
I will merge this after #720 which is fixing the tests.

PR #743, opened 2019-02-21
Remove ansible deprecation warning

Zlopez:

### DIFF
--- a/ansible/roles/anitya-dev/tasks/main.yml
+++ b/ansible/roles/anitya-dev/tasks/main.yml
@@ -81,12 +81,11 @@
 - name: Install Anitya into the virtualenv
   become_user: "{{ ansible_env.SUDO_USER }}"
   pip:
-    name: "{{ item }}"
+    name:
+      - "file:///home/{{ ansible_env.SUDO_USER }}/devel"
+      - "file:///home/{{ ansible_env.SUDO_USER }}/devel/anitya_schema"
     extra_args: "-e"
     virtualenv: /home/{{ ansible_env.SUDO_USER }}/.virtualenvs/anitya/
-  with_items:
-    - "file:///home/{{ ansible_env.SUDO_USER }}/devel"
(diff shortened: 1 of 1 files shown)


ansible/roles/anitya-dev/tasks/main.yml
```diff
@@ -81,12 +81,12 @@
 - name: Install Anitya into the virtualenv
   become_user: "{{ ansible_env.SUDO_USER }}"
   pip:
-    name: "{{ item }}"
+    name: [
+            "file:///home/{{ ansible_env.SUDO_USER }}/devel",
+            "file:///home/{{ ansible_env.SUDO_USER }}/devel/anitya_schema"
+          ]
```
bowlofeggs: Why use JSON syntax here instead of the way it was expressed previously?

Zlopez: I used the same syntax that is used for dnf module in this ansible script.
The way it was expressed before is throwing deprecation warning and will not work from ansible 2.8 and above.

bowlofeggs:
The deprecation warning is not about YAML syntax vs. JSON syntax - it's about giving the dnf module a list in the name key instead of using the ```with_items``` construction.

ansible/roles/anitya-dev/tasks/main.yml
bowlofeggs: i.e., this will also fix the deprecation warning, and won't be a mixture of YAML and JSON:

```
name:
  - "file:///home/{{ ansible_env.SUDO_USER }}/devel",
  - "file:///home/{{ ansible_env.SUDO_USER }}/devel/anitya_schema"
```

Zlopez: Oh, I understand now. I used the same syntax as for dnf module in same ansible script.

It's confusing that the YAML supports also JSON format.

I will fix this.

Zlopez:
Fixed and rebased.

PR #759, opened 2019-03-28
Fix failing tests

Zlopez:

### DIFF
--- a/anitya/tests/test_flask.py
+++ b/anitya/tests/test_flask.py
@@ -683,7 +683,7 @@ class FlaskTest(DatabaseTestCase):
         create_project(self.session)
 
         output = self.app.get("/project/geany")
-        self.assertEqual(output.status_code, 308)
+        self.assertIn(output.status_code, [301, 308])
         self.assertEqual(output.headers["Location"], "http://localhost/project/geany/")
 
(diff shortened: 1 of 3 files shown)


anitya/tests/test_flask.py
```diff
@@ -683,7 +683,7 @@ def test_projects_search_by_name(self):
         create_project(self.session)
 
         output = self.app.get("/project/geany")
-        self.assertEqual(output.status_code, 301)
+        self.assertEqual(output.status_code, 308)
```
pypingou: This seems to be a change in a recent version of either flask or werkzeug, I wonder if you shouldn't do something like:
``self.assertTrue(output.status_code in [301, 308])``

Zlopez: I updated this to `assertIn`

I found some documentation about the error 308 and it should be more reliable than 301 (doesn't allow change from `POST` to `GET`). See https://httpstatuses.com/308

pypingou: I've just checked, it's indeed a change in werkzeug: https://werkzeug.palletsprojects.com/en/0.15.x/changes/#version-0-15-0 (4th bullet point)

PR #761, opened 2019-03-29
Fix message not send on package API request

### LINKED ISSUE #760: anitya.project.map.new not send when adding new mapping through APIv2
There is missing message publish call when adding new package mapping using APIv2. This causing issue when project was added, new version found, but the mapping is added later.

Zlopez:
Fixes #760 

### DIFF
--- a/anitya/tests/test_flask_api_v2.py
+++ b/anitya/tests/test_flask_api_v2.py
@@ -1,6 +1,6 @@
 # -*- coding: utf-8 -*-
 #
-# Copyright © 2017  Red Hat, Inc.
+# Copyright © 2019  Red Hat, Inc.
 #
 # This copyrighted material is made available to anyone wishing to use,
 # modify, copy, or redistribute it subject to the terms and conditions
@@ -23,6 +23,7 @@
 Tests for the Flask-RESTful based v2 API
 """
 from __future__ import unicode_literals
+from unittest import mock
 
 import json
 from social_flask_sqlalchemy import models as social_models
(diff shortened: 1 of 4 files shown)


anitya/tests/test_flask_api_v2.py
```diff
@@ -481,6 +483,22 @@ def test_valid_request(self):
...
+            "project": dict(project.__json__()),
+            "distro": {"name": "Fedora"},
+            "message": {
+                "agent": "user@fedoraproject.org",
+                "project": "requests",
+                "distro": "Fedora",
+                "new": "python-requests",
+            },
+        }
+        mock_publish.assert_called_with(topic="project.map.new", msg=msg)
```
cverna: you use fedora-messaging `testing.mock_sends` for that if you prefer. See example here --> https://pagure.io/waiverdb/blob/master/f/tests/test_events.py#_31

Zlopez: Thanks, I will change it.

Zlopez: I tried to implement the `testing.mock_sends`, but I'm getting this in tox:
```
TypeError: test_valid_request() missing 1 required positional argument: 'mock_publish'
```

The method looks like this now:
```python
    def test_valid_request(self, mock_publish):
        """Assert packages can be created."""
        request_data = {
            "project_ecosystem": "pypi",
            "project_name": "requests",
            "package_name": "python-requests",
            "distribution": "Fedora",
        }
        project = models.Project.query.filter_by(
            name="requests", ecosystem_name="pypi"
        ).one()

        msg = {
            "project": dict(project.__json__()),
            "distro": {"name": "Fedora"},
            "message": {
                "agent": "user@fedoraproject.org",
                "project": "requests",
                "distro": "Fedora",
                "new": "python-requests",
            },
        }

        expected_msg = api.Message(
            topic="anitya.project.map.new",
            body=msg
        )

        with testing.mock_sends(expected_msg):
            output = self.app.post(
                "/api/v2/packages/", headers=self.auth, data=request_data
            )

        self.assertEqual(output.status_code, 201)
        data = _read_json(output)
        self.assertEqual({"distribution": "Fedora", "name": "python-requests"}, data)
```

Zlopez: I will leave it like it is now, because of the change in `project` object that is done inside the post request.

PR #787, opened 2019-06-07
Add option to check releases instead tags on GitHub

### LINKED ISSUE #733: Allow fetching releases on Github backend
Add checkbox to project, that will allow checking for releases instead of tags.

This will solve the issue with projects, that are using tags for other things than only releases.

### LINKED ISSUE #734: Handling projects which tag things that aren't releases
So the Slic3r Prusa Edition project (https://release-monitoring.org/project/13388/) has a somewhat bizarre set of tags.  It is a fork of another project but uses a different versioning scheme.  When it pulls from its "upstream" it inherits the tags, which all are numerically much larger.

In addition, there are all sorts of tags made to the repo that don't correspond to real releases.  The actual releases all seem to correspond to tags prefixed with "version_".

You can certainly supply a version prefix to anitya, but this has the effect of simply removing the prefix if it exists.  If it doesn't exist in a tag, that tag is unchanged.  What would make more sense to me is if anitya simply ignored all tags that don't match the version prefix (as well as removing that prefix from the version as normal).  But I suspect but changing the behavior of this would break a number of existing projects.

Another option is to somehow teach anitya how to get actual releases out of github rather than looking at all tags.  This would need to be an option because plenty of projects only make tags and don't designate releases.

Zlopez:
Add new `releases_only` check on project and new call for GitHub API v4
which is checking releases instead of tags.

Fixes #733

Fixes #734

### DIFF
--- a/anitya/db/migrations/versions/5e209766aead_add_release_check_to_project.py
+++ b/anitya/db/migrations/versions/5e209766aead_add_release_check_to_project.py
@@ -1,0 +1,33 @@
+"""Add release check to project
+
+Revision ID: 5e209766aead
+Revises: e34988f3e2f4
+Create Date: 2019-06-06 14:53:06.236820
+"""
+
+from alembic import op
+import sqlalchemy as sa
+
+
+# revision identifiers, used by Alembic.
+revision = "5e209766aead"
+down_revision = "e34988f3e2f4"
+
+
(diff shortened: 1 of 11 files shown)


Zlopez:
The documentation build should be fine after PR #789 is merged

PR #797, opened 2019-07-29
Restore missing closing """ in sample configuration

stormi:
A previous commit has removed the closing """ from the default_regex
string. This commits adds it back.


Zlopez:
Could you add news file to `news` folder?

See https://anitya.readthedocs.io/en/latest/contributing.html#release-notes

stormi:
Is it good this way?

PR #818, opened 2019-08-26
Update black formatting

vfazio:
Use the prerelease version of black

Enforce py36 standards since python2 support was dropped


Zlopez:
I'm not sure if it's good to use prerelease version of black for formatting our code.

However enforcing python 3.6 sounds like a good idea.

Zlopez:
Ok, in this case I only recommend to add news file and will approve it.

PR #846, opened 2019-10-14
Use tag name as version instead of release name

### LINKED ISSUE #845: "Internal Server Error" while adding new project
I am trying to add https://github.com/watchexec/watchexec 

image

This submission fails with internal server error.

Zlopez:
Because GitHub allows name of the release to be anything. It will be
better to use tags associated with the releases instead of using the
release names.

Fixes #845

### DIFF
--- a/anitya/lib/backends/github.py
+++ b/anitya/lib/backends/github.py
@@ -161,6 +161,7 @@ class GithubBackend(BaseBackend):
             )
 
         versions = parse_json(json, project)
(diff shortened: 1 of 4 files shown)


abitrolly:
Does that remove this option?

image

Zlopez:
> Does that remove this option?
> 
> image

No, this only uses name of the tag from which release was created instead of name of the release. I tested it on watchexec and this worked.

PR #852, opened 2019-11-13
Fix for building docs in travis

Zlopez:


Zlopez:
It looks like this was only temporary issue.

PR #858, opened 2019-12-12
Bump Vagrant box to Fedora 31

nphilipp:
This PR just bumps the `config.vm.box` value. The Vagrant box provisions well on top of F31 without changes to the Ansible files. I've skimmed them and only found one thing where I'm not sure if I should change it, namely in `ansible/roles/anitya-dev/files/bashrc`, there's this line...:

````
source /usr/bin/virtualenvwrapper-3.sh
````

...which is a symlink to `virtualenvwrapper.sh`. If you want, I can change that one, too, just to be prepared if the symlink ever gets dropped.

### DIFF
--- a/news/PR858.dev
+++ b/news/PR858.dev
@@ -1,0 +1,1 @@
+Bump Vagrant box to Fedora 31
--- a/Vagrantfile
+++ b/Vagrantfile
@@ -4,7 +4,7 @@
 VAGRANTFILE_API_VERSION = "2"
 
 Vagrant.configure(VAGRANTFILE_API_VERSION) do |config|
- config.vm.box = "fedora/30-cloud-base"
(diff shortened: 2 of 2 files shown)

### CODE REVIEW

**Comment by nphilipp** [comment]
> Could you just add news file for this change?

Done. Is the format I used okay?

File: news/PR858.dev [hunk 1]
```diff
@@ -0,0 +1 @@
+bump Vagrant box to be based off Fedora 31
```
Zlopez:
```suggestion
Bump Vagrant box to Fedora 31
```

Zlopez: Just a small suggestion, this will be in changelog later, so it's better if it's start with capital letter.

nphilipp: Done.

PR #859, opened 2019-12-12
Change ecosystem when homepage or backend is changed

### LINKED ISSUE #764: If URL is changed, update ecosystem value as well
Example
https://release-monitoring.org/project/13720/

I modified the .de into .com for the blogspot URL, but ecosystem stayed in .de.
This is problematic because the APIv2 search parameters is using ecosystem to match projects.

Zlopez:
Till now ecosystem was only set once, when project was created. With
this commit ecosystem is changed when homepage or backend change is
emitted. The ecosystem value is set as follows:
 - Ecosystem is set using backend if backend is associated with ecosystem
 - Ecosystem is set to homepage

Fixes #764

### DIFF
--- a/anitya/db/__init__.py
+++ b/anitya/db/__init__.py
@@ -32,4 +32,7 @@ from .models import (  # noqa: F401
     User,
     ApiToken,
 )
-from .events import set_ecosystem  # noqa: F401
+
+# You need to import the events to register them with application
+# If they are not imported, there wouldn't be triggered
+from . import events  # noqa: F401
--- a/anitya/db/events.py
+++ b/anitya/db/events.py
@@ -19,50 +19,68 @@ import logging
 from sqlalchemy import event
 
 from anitya.lib import plugins
-from .meta import Session
 from .models import Project
 
 
 _log = logging.getLogger(__name__)
 
 
-@event.listens_for(Session, "before_flush")
-def set_ecosystem(session, flush_context, instances):
+def _set_ecosystem(project, backend, homepage):
     """
-    An SQLAlchemy event listener that sets the ecosystem for a project if it's null.
+    Set ecosystem to correct value. Priority is as follows:
+        1. Ecosystem is set to backend if backend is associated with ecosystem
+        2. Ecosystem is set to homepage
 
     Args:
-        session (sqlalchemy.orm.session.Session): The session that is about to be committed.
-        flush_context (sqlalchemy.orm.session.UOWTransaction): Unused.
-        instances (object): deprecated and unused
-
-    Raises:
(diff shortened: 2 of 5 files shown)


anitya/db/__init__.py
```diff
@@ -32,4 +32,4 @@
     User,
     ApiToken,
 )
-from .events import set_ecosystem  # noqa: F401
+from .events import set_ecosystem_backend, set_ecosystem_homepage  # noqa: F401
```
nphilipp: Is there even a reason to import `set_ecosystem_*` here? AIUI, neither function is used anywhere else in the code, they just work by being decorated with `event.listens_for()`.

anitya/db/events.py
```diff
@@ -19,50 +19,68 @@
...
-        ValueError: If the ecosystem_name isn't valid.
+    Args:
+        target (sqlalchemy.orm.state.InstanceStace): Instance of the object where
+                change is happening.
+        value (str): The new value of backend.
+        old (str): The old value of backend.
+        initiator (sqlalchemy.orm.attributes.Event): The event object that is initiating this
+                transition.
+    """
+    if not value == old:
```
nphilipp: Why not `if value != old:`? :wink:

anitya/db/events.py
```diff
@@ -19,50 +19,68 @@
...
-                # Validate the field
-                valid_names = [e.name for e in plugins.ECOSYSTEM_PLUGINS.get_plugins()]
-                valid_names.append(new_obj.homepage)
-                if new_obj.ecosystem_name not in valid_names:
-                    raise ValueError(
-                        'Invalid ecosystem_name "{}", must be one of {}'.format(
-                            new_obj.ecosystem_name, valid_names
-                        )
-                    )
+    if not value == old:
```
nphilipp: Same here: `if value != old:`

anitya/db/__init__.py
Zlopez: Probably not, it will be better to remove it.

anitya/db/events.py
Zlopez: I'm probably just confused from java and I thought this is not working in Python. :-D

anitya/db/__init__.py
Zlopez: After trying to remove it, I could say that this is needed. Without it the events are not registered and just skipped.

Zlopez: I added a comment to this line, so it's clear, why it's needed.

nphilipp: AIUI, merely importing the module should be sufficient:

````
# The events module needs to be imported so the handlers are registered.
from . import events  # noqa: F401
````

PR #863, opened 2019-12-17
GitHub: Store and use latest known version cursors

nphilipp:
Along with that, change how VCR records remote API calls:

- decode compressed responses so the stored "cassettes" are easier to be read by humans
- replace authorization headers with placeholders to avoid exposing credentials


anitya/lib/backends/github.py
```diff
@@ -132,36 +132,65 @@ def get_versions(cls, project):
...
+            # without one.
+            cursor_attempts = (project.latest_known_cursor, None)
         else:
-            raise AnityaPluginException(
-                '%s: Server responded with status "%s": "%s"'
-                % (project.name, resp.status_code, resp.reason)
-            )
+            cursor_attempts = (None,)
+
+        for cursor in cursor_attempts:
```
Zlopez: We are missing test, where the cursor_attempts is empty.

anitya/tests/lib/backends/test_github.py
```diff
@@ -232,6 +232,19 @@ def test_get_versions(self):
         obs = backend.GithubBackend.get_ordered_versions(project)
         self.assertEqual(obs, exp)
 
+        project.latest_known_cursor = "invalid cursor"
```
Zlopez: Could you move this to separate test? I want to break this to multiple tests anyway in the future.

anitya/db/models.py
```diff
@@ -335,10 +336,15 @@ def create_version_objects(self, versions):
         versions = sorted(
             [
                 version_class(
-                    version=version,
+                    version=version if isinstance(version, str) else version["version"],
```
Zlopez: Why this change?

anitya/lib/backends/github.py
nphilipp: `cursor_attempts` can't be empty, it is either `(project.latest_known_cursor, None)` or `(None,)`, per lines 139 & 141. It implements attempting retrieving versions after the latest known cursor at first, and if that fails (or no cursor is known), it retrieves the whole list.

anitya/tests/lib/backends/test_github.py
nphilipp: Then let me break it up now. I was bothered by this test being so big anyway. :wink:

anitya/db/models.py
nphilipp: Oops, I seem to have missed to commit the code which actually passes around cursors with the versions here. Hang on.

nphilipp:
The latest push tidies up the GitHub backend tests, mainly:

- make test projects accessible by mnemonics rather than numerical ids
- have one method per test

Issue #829, opened 2019-09-05
Add counter for number of failed check to project


Zlopez:
# Description
Currently we only watching error output from last check, which is enough, but it will be nice to know how long this is happening. So I want to introduce following enhancements:
1. Add counter to project, which will track number of failed checks and will be reset to 0, when project is successfully checked
2. Add option to search for project based on the number of failures (this will probably need db index)
3. (Optionally) Remove projects, which are failing for long time (could be configurable amount of failures) and didn't retrieved any version yet

## Requirements
We need to change how no new version is announced. We currently announcing is as `AnityaException`, which is considered as error.
We should make difference between `No version found` and `No new version found`, it could look similar, but:
* `No version found` could mean that there is no version on the page we are checking, regex is no longer matching anything (most of the backends are using regex for version matching), this could mean that page no longer exists or it was edited.
* `No new version found` means that we are still see some version, but no new was found. This shouldn't be an error, this just means, that project didn't released any new version. This should be much easier to recognize with #778 merged.

## How this will make Anitya better
Anitya currently contains plenty of projects, which can't be checked because the site no longer exists, project is incorrectly set up, the page has different content etc. This enhancement will help with identifying and possibly removing those projects.

PR #872, opened 2019-12-19
Add automation for removing projects

Zlopez:
New configurable field is introduced `error_threshold`. When project
reaches this threshold and didn't retrieved any version, it is
automatically deleted.

This will remove projects, that are incorrectly set up.

Implements optional part of #829


Zlopez:
This PR needs #864 merged, because it requires `error_counter` field on project.

PR #874, opened 2019-12-20
Link to commit of latest version if known

nphilipp:
This only links the commit from the latest version field, where we can make it explicit that it's a commit, not a release page or anything:

Screenshot from 2019-12-20 23-18-55


nphilipp:
Hmm, there are two failures flagged by Travis CI:

* #1008.6 "Python: 3.7": anitya.tests.test_check_service::CheckerTests::test_run_timeout() failing which it does every now and again, without rhyme or reason
* #1008.7 "Python: nightly": ERROR: InterpreterNotFound: python3.8

I'll force push again to see if the #1008.6 issue improves, there's nothing I can do about the latter.

Zlopez:
@nphilipp 3.8 could be safely ignored
However the `test_run_timeout` test is failing sometimes and I can't figure out why exactly is this happening.

PR #877, opened 2020-01-01
Do not duplicate project search form

abitrolly:
Currently search form for projects present both in page header and in results table header. This PR leaves search form in table header only when searching inside distros.

### DIFF
--- a/anitya/templates/search.html
+++ b/anitya/templates/search.html
@@ -39,17 +39,14 @@
 
   <div class="col-sm-6">
     {% if distroname %}
-    <form action="{{ url_for('anitya_ui.distro_projects_search', distroname=distroname)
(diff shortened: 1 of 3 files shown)


tox.ini
```diff
@@ -45,7 +45,7 @@ commands =
 install_command = pip install --pre {opts} {packages}
 deps = black
 commands =
-    python -m black --check {posargs:.}
+    python -m black --check --diff {posargs:.}
```
Zlopez: I don't think this is a good idea. It is better to knew that format is failing somewhere else.

If there is something failing it should be fixed in another PR.

abitrolly:
Squashed, moved black to another PR and added new fragment.

Issue #881, opened 2020-01-08
Libraries.io consumer is creating junk PyPi projects


yarons:
I'm not sure what happened but it seems like many new PyPi project were added, looking at all the projects page I'm seeing that there are many projects with the same version number and links that doesn't work.

What's up? Maybe it's a malicious act that requires a rollback?

Zlopez:
This could be caused by libreries.io consumer, which is listening to every project from PyPi and Rubygems. But this should be solved by https://github.com/release-monitoring/anitya/issues/829 in the future.

yarons:
But, what are we doing to remove the entries now? Will this fix remove these entries from the current DB?

Zlopez:
Yes, it will remove every project, that couldn't be checked for new version after few days.

yarons:
Hmmm but what about projects with a temporary website issue?

Zlopez:
If the project ever retrieved any version it will not be deleted. This automation will only delete projects without any version, so only the ones that have incorrect setup.

yarons:
Cool, sounds better.

Zlopez:
This will be part of next release, which is now waiting for last PR to be merged. So hopefully this will be out soon.

Zlopez:
The projects created by libraries.io consumer had at least one version. So this new automatic will unfortunately not remove these junk projects. I need to look at the consumer and see why is this happening.

Zlopez:
After looking at the libraries.io consumer script, I can tell that this project must have existed on PyPi for some time and probably was only created and then removed from PyPi. So I will try to clean the projects created by this situation, but this looks like issue on PyPi side and I'm not sure if we actually could prevent this.
I could probably expand the automatic later to also remove the project with only one retrieved version, but this could potentially remove good project.

yarons:
Cool, all I could notice is that the name contsins 2 characters and that the names are increasing by 1 and the only version is 0.0.1.

Zlopez:
I did something better in #882, I started to use mapping to distro as another condition. So with this change it will delete projects, that reach error counter and have no mapping. If they have mapping, but don't have any version, they are still deleted.

This should remove many junk projects, that didn't received any version and only throwing error, or doesn't receive versions anymore and didn't have any mapping to downstream. This should clear Anitya db a little.

Zlopez:
For other projects that are failing and doesn't fall in any of these condition, I want to add option to archive projects. See #865 for more information.

Zlopez:
[...]

PR #882, opened 2020-01-09
Use mapping as another condition for deletion

Zlopez:
Recently there was a flood of junk projects from PyPi received through
libraries.io consumers. These projects doesn't exist anymore on PyPi,
but at least one version was retrieved, so current solution of delete
candidate recognition doesn't recognize this project as candidate
for deletion, but if the project doesn't have mapping it could be
in most cases safely deleted.

Still if there is mapping and no versionis retrieved it will be deleted.
The prime condition is still the error threshold.


Zlopez:
This is additional enhancement for #829 and should also fix #881

PR #883, opened 2020-01-09
Fix search for projects by distro

### LINKED ISSUE #876: Distro search is broken
https://release-monitoring.org/distro/Alpine/search/?pattern=acpi gives 500 Internal Error.

The fix is the same as https://github.com/release-monitoring/anitya/pull/757/files but I am curious why the `distro` relationship doesn't work here?

https://github.com/release-monitoring/anitya/blob/299701f9865f3833cd4eaa4413effa32a5b0996a/anitya/db/models.py#L156-L158

Zlopez:
Use distro_name instead of distro relationship on package, because the
later caused error 500.
Limit project search by mapping to distro if available. Otherwise we
will end up with projects from another distros.

Fixes #876


abitrolly:
@Zlopez did you find out why `distro` relationship didn't work? Maybe relationship is unnecessary?

PR #895, opened 2020-03-22
GitHub backend: do not raise error if no new upstream version

### LINKED ISSUE #892: No upstream version found doesn't mean there is error in project check
# Description
Currently if we don't retrieve any version we raise an exception, but this could be correct if there is no new version (at least for GitHub, which is using cursor now). There should be difference between no new version found and no version found at all (project is incorrectly setup or the upstream doesn't exist).

## Requirements
* If project could be checked, but no version could be retrieved it shouldn't throw any error. (This is currently happening only in GitHub backend)

It will prevent raising false alarms about no longer updated projects.

olivierlemasle:
For GitHub backend, Anitya throws an error and fails with "No upstream version found" if there's no new version. However, it should throw an error only if it cannot fetch versions, or if there's no version at all, but not when there's no new version because of the cursor.

Fix #892

### DIFF
--- a/anitya/lib/backends/github.py
+++ b/anitya/lib/backends/github.py
@@ -195,7 +195,7 @@ class GithubBackend(BaseBackend):
             project.latest_version_cursor = None
             versions = cls._retrieve_versions(owner, repo, project)
 
-        if len(versions) == 0:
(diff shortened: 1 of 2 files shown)


olivierlemasle:
@Zlopez Ok, done (and commits squashed)

PR #914, opened 2020-05-18
Fix broken links to fedmsg.com

abitrolly:
See https://github.com/fedora-infra/fedmsg/issues/524

I left link to web archive in few places, because it contains better introduction for newcomers than this text from current RTD site.

> fedmsg (Federated Message Bus) is a library built on ZeroMQ using the PyZMQ Python bindings. fedmsg aims to make it easy to connect services together using ZeroMQ publishers and subscribers.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -32,7 +32,7 @@ Anitya
 Anitya is a release monitoring project. It provides a user-friendly interface
 to add, edit, or browse projects. A cron job can be configured to regularly
 scan for new releases of projects. When Anitya discovers a new release for a
-project, it publishes a ZeroMQ message via `fedmsg <http://fedmsg.com>`_.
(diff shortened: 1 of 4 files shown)


abitrolly:
I don't know why tests failed.

Zlopez:
This test sometimes fail, not sure why, but it does.

Thanks for the PR, I want to remove every fedmsg related stuff in next release.

abitrolly:
@Zlopez do you mean replace with [`fedora-messaging`](https://github.com/fedora-infra/fedora-messaging)? Because knowing that you can subscribe to your own updates feed is useful. Worthy to be documented for sure.

Zlopez:
@abitrolly The replacement is already in place, but there is still legacy fedmsg mode you can switch to. I will try to document it, the documentation itself would be worthy of updating.

Issue #753, opened 2019-03-08
Separate stable releases from unstable ones


cheese1:
It would be great to be able to **exclude** specific patterns.
Sometime one does not want to get notified of alpha,beta,pre- and test-releases, release-candidates....

it worked in such a way with cnucnu and i always miss that feature.
i for my part almost never package unstable software and i think most packagers do so.

Zlopez:
I actually thought about this feature recently. I just need to think how to implement this. Currently I have few ideas in the head:

1. Let this be handled by the version scheme - it should mark them as pre-release versions
2. Let this on user - add new field, where user could specify, which versions should be ignored, every version containing any string in this field will be ignored.
3. Combine the two options above, let something be handled by version scheme and rest by user.

Either case will help with false notifications about new versions and will made versions on the project more readable.

cheese1:
i think a way to accomplish that would be:
* have a list of "stable" versions  (the full list minus alpha, beta,...)
* have a second list of all versions (we already have that)
* have a predefined regex to define the "minus"-list, it should be possible to redefine that regex

new tickets (for fedora) should be triggered for the "stable" list, in my opinion.
the output via api should have a new line version_stable and a new array versions_stable, eg:

{
  ...
            "updated_on": 1482495004.0,
            "version": "2.12-alpha",
            "version_stable": "2.10",
            "version_url": "zero-install",
            "versions": [
                "2.12-alpha",
                "2.11-beta",
                "2.10",
                "2.9.1",
                "2.9",
                "2.8",
                "2.7"
            ],
            "versions_stable": [
                "2.10",
                "2.9.1",
                "2.9",
                "2.8",
                "2.7"
            ]
        }

Zlopez:
This sounds like a good idea. I will look at it more closely, when I start working on this issue.

yarda:
I have the same problem with Tcl package in Fedora (https://bugzilla.redhat.com/show_bug.cgi?id=1488695). I just want bugzilla reports about stable releases, not alphas, but it seems:
* I cannot change the regex in anitya, becase there are already mappings for more downstreams and I don't want to break them (maybe they are interested in getting notifications about alphas)
* Adding new project like e.g. tcl-stable with different regex could help, but I consider it as a dirty hack which should be avoided

So at the moment there is Fedora bugzilla about development version of the package, but I am missing notifications about stable updates (this is what recently happened in (https://bugzilla.redhat.com/show_bug.cgi?id=1488695).

[...]

Issue #623, opened 2018-09-27
Allow removing more versions at once


Zlopez:
It will be easier for administrators if they could remove multiple project's versions at once.

This will allow much faster solving of flagged projects.

cheese:
[OT] How to apply administratorship? I flagged multiple projects but none has been worked on.

Zlopez:
@cheese I get through the flagged projects on friday, it depends what issue you had.
If you wanted to remove some versions, I can do this, but right now it doesn't matter. The check function is now retrieving every version.

I want to add some mechanic that will prevent this in the future, so the project will not retrieve versions that you don't want.

See https://github.com/release-monitoring/anitya/issues/753

Zlopez:
If you want to be admin in Anitya, just give me your e-mail you are using in Anitya and I can give you administrator rights.

cheese:
I use FAS to login. My FAS id is cheeselee. The email is cheeselee@fedoraproject.org or robinlee.sysu@gmail.com.

Zlopez:
@cheese I gave you the admin rights on release-monitoring.org

If you have any question ask on #fedora-apps freenode channel or here.

cheese:
OK. Thanks. I try to fix issues I flagged to get started.

PR #944, opened 2020-06-29
Add filter field for versions

### LINKED ISSUE #898: Add filter field for versions
# Description
We should add filter field to versions, so we could prevent retrieving some of the versions. This field should prevent the versions containing matching string to be retrieved. 

For example if filter is set to `prefix`, it will skip versions like `prefix-5.5`, `3.2.1-prefix` and so on.

## Requirements
* Add field to `project_new` template with tooltip
* Add method to `BaseBackend` that will filter the retrieved versions based on the project filter
* Add column `filter` to `Project`, this column will be same as prefix
* Call the new method in every backend before returning the versions
* (Optional - only for admin) Add button that applies filter to existing project and removes every version that matches the filter

This will allow users to filter versions containing specific strings and lowers the number of flaged projects with versions for deletion.

Zlopez:
* Add filter filed to project html template
* Add method for filtering to BaseBackend
* Call the above method on every backend
* Add column filter to Project
* Add migration for new column

Closes #898


Zlopez:
After looking at #753 I'm starting to think about getting rid of this PR and using the parts of it. The filter field could be used to identify unstable versions and the apply filter button for admins could be used to implement #623 

Not retrieving some versions from upstream is not a good way to go.

Zlopez:
This PR helped to implement #753 and #623 and it's no longer needed

Zlopez:
After I thought about this for a while it could still be helpful. There are projects, that are tagging things that are not releases at all and this will help to filter that.

PR #967, opened 2020-10-01
Remove social_auth info from SAR script

### LINKED ISSUE #954: sar.py fails with AttributeError: 'User' object has no attribute 'social_auth'
Script for recovering GDPR related data fails with the error in title.

Here is the full traceback:
```
  File "/usr/local/bin/sar.py", line 72, in <module>
    main()
  File "/usr/local/bin/sar.py", line 65, in main
    users_list.append(user.to_dict())
  File "/usr/local/lib/python3.8/site-packages/anitya/db/models.py", line 947, in to_dict
    for soc_auth in self.social_auth.all():
AttributeError: 'User' object has no attribute 'social_auth'
```

Zlopez:
To fix #954 I decided to remove social auth info from SAR script. Only
useful information provided was the list of providers the user was
authenticated with.

The info from the SAR script now contains id of the user, mail and if
the account is active or not.

Fixes #954.

### DIFF
--- a/anitya/db/models.py
+++ b/anitya/db/models.py
@@ -995,22 +995,12 @@ class User(Base):
         Returns:
             dict: `User` object transformed to dictionary.
         """
-        social_auth = []
-        for soc_auth in self.social_auth.all():
-            social_auth.append(
-                dict(
-                    provider=soc_auth.provider,
-                    extra_data=soc_auth.extra_data,
-                    uid=soc_auth.uid,
-                )
-            )
 
         return dict(
             id=str(self.id),
             email=self.email,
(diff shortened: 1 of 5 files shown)


pypingou:
> remove social auth info from SAR script. Only useful information provided was the list of providers the user was authenticated with.

Well, if that's information you have/record about the user, then we should include it

Zlopez:
@pypingou I didn't found any clear way to actually get the info from social_auth tables, there should be `social_auth` relation, but this doesn't look like it's available, thus the error.

The only way I can think of is dirty way to call the SQL statements directly to select from these tables.

PR #1053, opened 2021-03-10
 Link Fedora packages to their source (#557)

### LINKED ISSUE #557: Link Fedora package_name to package info page
image

Would be good to go directly to package page for specified distribution.

abitrolly:
Because there are no package pages yet,
and because it is easier to track when the
package version will appear there.

Closes #557.

### DIFF
--- a/anitya/templates/project.html
+++ b/anitya/templates/project.html
@@ -186,7 +186,15 @@
                 {% for package in project.packages %}
                 <tr>
                   <td>{{ package.distro_name }}</td>
-                  <td>{{ package.package_name }}</td>
+                  <td>
+                    {% if package.distro_name == 'Fedora' %}
+                      <a href="https://src.fedoraproject.org/rpms/{{ package.package_name }}">
+                        {{ package.package_name }}
+                      </a>
+                    {% else %}
+                      {{ package.package_name }}
+                    {% endif %}
+                  </td>
                   <td>
                     <a href="{{
                       url_for('anitya_ui.edit_project_mapping',
--- a/news/557.feature
+++ b/news/557.feature
@@ -1,0 +1,1 @@
+Link Fedora packages to their source


anitya/templates/project.html
```diff
@@ -186,7 +186,15 @@ <h3 class="panel-title">Mappings</h3>
                 {% for package in project.packages %}
                 <tr>
                   <td>{{ package.distro_name }}</td>
-                  <td>{{ package.package_name }}</td>
+                  <td>
+                    {% if package.distro_name == 'Fedora' %}
+                      <a href="https://bugzilla.redhat.com/buglist.cgi?component={{ package.package_name }}&product=Fedora">
```
Zlopez, line 191: Wouldn't it be better to point to src.fedoraproject.org? It has all the related links and it also allows you to change notification settings for Anitya.

Zlopez (changes requested):
This is excellent feature! I only have one thing that is missing, could you add a news file? It's used for changelog generation. See https://anitya.readthedocs.io/en/latest/contributing.html#release-notes

news/feature.557
```diff
@@ -0,0 +1 @@
+Link Fedora packages to Bugzilla
```
Zlopez, line 1: The name of the file should be reversed. It should be `557.feature`.

anitya/templates/project.html
abitrolly: Search on that side gave me a link to https://src.fedoraproject.org/rpms/kompose and there is just a source code.

I would be interested to immediately see the current stable Fedora version, the version of package in that stable, the latest version, if there is a new version pending, and how to start version update process, what is the process.

abitrolly: Ok. I didn't expect pagure to contain the package info. Guess makes sense to link to that.

abitrolly:
Let me redo this.

Zlopez (approved):

abitrolly:
I now understand why I didn't notice the version box at https://src.fedoraproject.org/rpms/kompose It takes more than 5 seconds to load it.

abitrolly:
Opened https://pagure.io/pagure/issue/5132

Zlopez:
> I now understand why I didn't notice the version box at https://src.fedoraproject.org/rpms/kompose It takes more than 5 seconds to load it.

Yeah, it's dynamically loaded, so it takes time to show it.

Zlopez:
I will try to release this on Monday.

Issue #1064, opened 2021-03-17
Linking to distro package


glensc:
I see that `adcli` has a link to Fedora distribution:
- https://release-monitoring.org/project/21/

![image](https://user-images.githubusercontent.com/199095/111442121-b3298000-8710-11eb-9313-4b40cae9cde3.png)

how to add such link to other distributions?

Zlopez:
This was added by this [PR](https://github.com/fedora-infra/anitya/pull/1053) created by @abitrolly, feel free to look how you can do it for other distribution.

glensc:
Oh, such wow hack, no problem adding another one:
- https://github.com/fedora-infra/anitya/pull/1065

But I think the distro mapping should come from config rather hardcoded conditions to template.
Also, that template is huge, I'm sure the templating engine used supports "include" other templates to make templates smaller and more specific function.

I'll leave those concerns to maintainers.

Zlopez:
We could make it configurable, but it's not really a priority right now.

That template is that huge till the time I started maintaining that project, there wasn't really that much changed, but splitting it to multiple sub-templates sounds like a good way to make it better maintainable.

Zlopez:
I added #1067 and #1066 to track these enhancements.

Zlopez:
I will close this one in favor of #1066

PR #1065, opened 2021-03-17
Add PLD Linux package link to project page

glensc:
Similarly to https://github.com/fedora-infra/anitya/pull/1053

### DIFF
--- a/anitya/templates/project.html
+++ b/anitya/templates/project.html
@@ -191,6 +191,10 @@
                       <a href="https://src.fedoraproject.org/rpms/{{ package.package_name }}">
                         {{ package.package_name }}
                       </a>
+                    {% elif package.distro_name == 'PLD-Linux' %}
+                      <a href="https://github.com/pld-linux/{{ package.package_name }}">
+                        {{ package.package_name }}
+                      </a>
                     {% else %}
                       {{ package.package_name }}
                     {% endif %}
--- a/news/PR1065.feature
+++ b/news/PR1065.feature
@@ -1,0 +1,1 @@
+Add PLD Linux package link to project page.


anitya/templates/project.html
```diff
@@ -191,6 +191,10 @@ <h3 class="panel-title">Mappings</h3>
                       <a href="https://src.fedoraproject.org/rpms/{{ package.package_name }}">
                         {{ package.package_name }}
                       </a>
+                    {% elseif package.distro_name == 'PLD-Linux' %}
```
glensc, line 194: I'm not sure what template engine is used here, so this may not work.

Zlopez: We are using jinja2 as template engine. I will try this on my local instance of Anitya to make sure, that it's working.

Zlopez (changes requested):
Could you also add a news file. See https://anitya.readthedocs.io/en/latest/contributing.html#release-notes or look at the #1053

anitya/templates/project.html
glensc: Note to myself, using jinja2:
- https://github.com/fedora-infra/anitya/blob/215adea038141bb59a4dd825e336b09d2c96f26e/requirements.txt#L8

glensc:
> Could you also add a news file. See [anitya.readthedocs.io/en/latest/contributing.html#release-notes](https://anitya.readthedocs.io/en/latest/contributing.html#release-notes) or look at the #1053

Added. However, I think the change is rather trivial not worth the changelog entry:
- https://keepachangelog.com/en/1.0.0/ (but this project doesn't indicate it's been followed)

Also, I think the effort for changelog creation, is too much for developers. Especially, if want to make a simple change from GitHub Web. Traversing files, finding where to add them, and whats the syntax. Bleh. You should look into better release workflows that can obtain the changelog/release notes from pull request titles and labels.

I personally have been keeping an eye on https://github.com/laminas/automatic-releases

My 2¢ (not attached)

anitya/templates/project.html
glensc: @Zlopez no need to test locally, Travis CI caught the invalid syntax fine:
- https://travis-ci.org/github/fedora-infra/anitya/jobs/763250533#L1331

Zlopez:
I only want the news file for changes that would be interesting for users of the application. If this is just some small change in documentation or fix for some typo in code I wouldn't care about it.

news/1065.feature
```diff
@@ -0,0 +1 @@
+Add PLD Linux package link to project page. #1065, #1064
```
Zlopez, line 1: Remove the link from the news file, those aren't needed.
```suggestion
Add PLD Linux package link to project page.
```

glensc: How else would you have a link to both of them in changelog?

Zlopez:
I was able to test it on the local instance of Anitya and the link is working fine, it redirected me to the source repo of the project.

Zlopez (approved):

anitya/templates/project.html
Zlopez: I still want to see that the link is really working.

news/1065.feature
Zlopez: No need to link both of them, just the issue itself, or PR if there is no issue. I see that you are using #1065 instead of #1064 in the name. This wouldn't work because the link will point to issue with id 1065, which doesn't exists.

I would recommend to link the rename it to 1064.feature

Zlopez (changes requested):
Change name of the news file to point to correct issue.

news/1065.feature
glensc: > This wouldn't work because the link will point to issue with id 1065, which doesn't exists.

yes, it will work tecnically, GitHub has a unique id namespace for issues and pull requests, and they both redirect to each other:
- https://github.com/fedora-infra/anitya/issues/1065
- https://github.com/fedora-infra/anitya/pull/1064

I would rather rename it to `PR1065.feature` then, as while it #1065 covers #1064, it doesn't solve it, i.e it's a rather open question for documentation or something.

however, I can't do file rename from GitHub web, so need to check out this locally to make changes, so this will take time.

Zlopez: Thanks for doing this

Zlopez (approved):

glensc:
Thanks, @Zlopez this will help to verify matched package validity more easily!

Zlopez:
No problem, I will merge the current pull requests and release a new version, so it will be in release-monitoring.org soon.

Issue #1068, opened 2021-03-17
Fix google code backend


Zlopez:
This backend is still pointing to old Google code page instead of the archived one. It will be nice to fix this.

Zlopez:
After some playing with the archived google.code I decided to remove the backend from Anitya. The versions on archived pages are shown as some javascript code and you can't actually access them by parsing the page by regex.

I found 28 projects that are still using it:
```
 eqntott
 scitools
 ratproxy
 innotop
 pyrit
 qtgpsc
 pyodbc
 uim
 gphotoframe
 pidgin-musictracker
 batti
 sir
 googlecl
 pidgin-privacy-please
 tcpjunk
 opencc
 msn-pecan
 spindown
 rst2pdf
 pyicqt
 hamcrest
 gnome-mplayer
 xmonad-log-applet
 libgcal
 jarjar
 usbsoftrock
 cagraph
 skipfish
```

I will edit them before removing the backend from Anitya.

The backend remove introduces backward incompatible change, so it probably means that we need to release Anitya 2.0.0.

PR #1069, opened 2021-03-17
Change the default regex to match unstable versions

### LINKED ISSUE #1063: Make the default regex pull in rc/alpha,etc?
https://release-monitoring.org/project/1257/

Zlopez:
The current regex only matched versions which didn't contained `-`
skipping every unstable release. This commit is updating the default
regex to also retrieve the unstable releases.

Closes #1063


Zlopez:
To make the tests pass, we first need to fix #1068

PR #1106, opened 2021-05-19
Use trusty dist for travis

Zlopez:


Zlopez:
This is actually Ubuntu 14.04, so it's not usable for us.

PR #1114, opened 2021-06-02
Added mypy POC and attempted to fix issues 

AdamSaleh:
Added mypy and attempted to fix issues  resulting from additional type-annotations. Removed py36.

Things I am not sure about:

**The calendar comparisons:** https://github.com/fedora-infra/anitya/compare/master...AdamSaleh:anitya_mypy_poc?expand=1#diff-6d66ea3411d110aa6961c53b9437d5adcecd56119de31c6a7530df1aa15314c3R491

I might go over-board with the isInstance checks, so some things are not comparable that were before. Or I have uncovered bug in the implementation? :-) If it's the former, I would like to figure out a way how to do this with mypy.

**The defaults (that are not None):** https://github.com/fedora-infra/anitya/compare/master...AdamSaleh:anitya_mypy_poc?expand=1#diff-7c42726a003d98fb9eab5f791a56c51d0422dde1d49020ec105961e0eadbdea3R86

Some should be easy, i.e. empty-list can be used in most places instead of None, because it already is False-y.
Some I am not sure about, like default REGEX, e.t.c.

**Typed dictionary**: it is really cool for mypy to shout at you when you have typo in your dictionary keys. But there already is support for dictionaries that don't contain all of the keys, and we might want to use that instead of lots of {key : None} initialization

### DIFF
--- a/anitya/lib/backends/__init__.py
+++ b/anitya/lib/backends/__init__.py
@@ -26,6 +26,7 @@ from datetime import timedelta
 
 # sre_constants contains re exceptions
 import sre_constants
+from typing import List
 import urllib.request as urllib
 from urllib.error import URLError
 
@@ -35,7 +36,8 @@ import arrow
 
 from anitya.config import config as anitya_config
 from anitya.lib.exceptions import AnityaPluginException
-from anitya.lib.versions import RpmVersion
+from anitya.lib.versions import GLOBAL_DEFAULT, RpmVersion
+
 import six
 
 REGEX = anitya_config["DEFAULT_REGEX"]
@@ -79,11 +81,11 @@ class BaseBackend(object):
             checking for new versions. This could be overriden by backend plugin.
     """
 
-    name = None
-    examples = None
-    default_regex = None
-    more_info = None
-    default_version_scheme = None
+    name: str
+    examples: List[str]
+    default_regex: str
+    more_info: str
+    default_version_scheme = GLOBAL_DEFAULT
     check_interval = timedelta(hours=1)
 
     @classmethod
--- a/anitya/lib/versions/calver.py
+++ b/anitya/lib/versions/calver.py
@@ -26,7 +26,7 @@ See `Calendar versioning`_.
 """
 import functools
 import re
-from typing import Tuple
+from typing import Optional, Tuple, TypedDict
 
 from .base import Version
 
@@ -53,6 +53,16 @@ def split_by_match(regex: str, string: str) -> Tuple[str, str]:
     return "", string
 
 
+class SplitResult(TypedDict, total=True):
+    year: Optional[str]
+    month: Optional[str]
+    day: Optional[str]
+    minor: Optional[str]
+    micro: Optional[str]
+    modifier: Optional[str]
+    rc_number: Optional[str]
+
+
 @functools.total_ordering
 class CalendarVersion(Version):
     """
@@ -70,7 +80,7 @@ class CalendarVersion(Version):
 
     name = "Calendar"
 
-    def split(self) -> dict:
+    def split(self) -> SplitResult:
         """
         Does a simple lexical analysis of the version against pattern.
 
@@ -82,14 +92,15 @@ class CalendarVersion(Version):
                 This could happen if the pattern is not correct or if
                 the version can't be parsed by the pattern.
         """
-        result_dict = {
-            "year": None,
-            "month": None,
+
+        result_dict: SplitResult = {
             "day": None,
-            "minor": None,
             "micro": None,
+            "minor": None,
             "modifier": None,
+            "month": None,
             "rc_number": None,
+            "year": None,
         }
         # Check if the pattern is correct
         if not self.pattern:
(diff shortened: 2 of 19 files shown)


.travis.yml
```diff
@@ -26,23 +26,18 @@ jobs:
...
         - bash -c 'shopt -s globstar nullglob; shellcheck **/*.sh'
         - tox
-    - python: 3.6
+    - python: 3.8
       env: TOXENV=docs
 
-    - stage: test
-      python: 3.6
-      env: TOXENV=py36
-    - python: 3.7
```
Zlopez, line 44: Why removing python 3.7?

anitya/lib/backends/__init__.py
```diff
@@ -79,11 +81,11 @@ class BaseBackend(object):
...
     """
 
-    name = None
-    examples = None
-    default_regex = None
-    more_info = None
-    default_version_scheme = None
+    name : str
+    examples : List[str]
+    default_regex = REGEX
```
Zlopez, line 86: This should be Optional, because the REGEX is only used for custom backend. I would go with str instead.

anitya/lib/versions/calver.py
```diff
@@ -53,6 +53,16 @@ def split_by_match(regex: str, string: str) -> Tuple[str, str]:
     return "", string
 
 
+class SplitResult(TypedDict, total=True):
```
Zlopez, line 56: I'm not a fun of having multiple classes in one .py file, but this looks good. I didn't knew about the `TypedDict` till now.

anitya/lib/versions/calver.py
```diff
@@ -258,12 +273,16 @@ def prerelease(self) -> bool:
             return True
 
         for pre_release_filter in self.pre_release_filters:
-            if pre_release_filter and pre_release_filter in self.version:
+            if (
+                pre_release_filter
+                and self.version
```
Zlopez, line 278: Thanks for catching this.

anitya/lib/versions/calver.py
```diff
@@ -300,15 +321,15 @@ def __lt__(self, other: Version) -> bool:
         Returns:
             (bool) True if lower than, False otherwise.
         """
-        try:
+        if isinstance(self, CalendarVersion):
```
Zlopez, line 324: This looks much better. :+1:

anitya/lib/versions/calver.py
```diff
@@ -300,15 +321,15 @@ def __lt__(self, other: Version) -> bool:
...
             version_dict_self = self.split()
-        except ValueError:
+        else:
             # The version can't be split, so we consider it lesser
             return True
 
-        try:
+        if isinstance(other, CalendarVersion):
```
Zlopez, line 330: Same as above.

anitya/lib/versions/rpm.py
```diff
@@ -48,10 +48,7 @@
 
     warnings.warn("Failed to import 'rpm', emulating RPM label comparisons")
 
-    try:
-        from itertools import zip_longest
-    except ImportError:
-        from itertools import izip_longest as zip_longest
+    from itertools import zip_longest
```
Zlopez, line 51: This was probably needed by old version of python, I'm glad this could be removed.

anitya/mail_logging.py
```diff
@@ -34,7 +34,9 @@
 
 psutil = None
 try:
-    import psutil
+    import psutil as _psutil
```
Zlopez, line 37: Why to do it like this?

anitya/tests/lib/versions/test_calver.py
```diff
@@ -488,8 +488,9 @@ def test_lt_both_prerelease_one_unversioned(self):
 
     def test_lt_nonsense(self):
         """Assert CalendarVersion supports < comparison."""
+        # unclear if it should really be valid -junk modifier?
         old_version = calver.CalendarVersion(
-            version="2019.04.23junk", pattern="YYYY.0M.DD-MODIFIER"
```
Zlopez, line 492: This was correct, look at the name of the test. I'm trying to test if nonsense version would be lesser than the correct version.

tox.ini
```diff
@@ -1,5 +1,5 @@
 [tox]
-envlist = py36,py37,py38,diff-cover,lint,format,bandit,docs
+envlist = py38,diff-cover,lint,format,bandit,docs
```
Zlopez, line 2: Could you add mypy to test run?

.travis.yml
Zlopez: Could you add mypy to .travis.yml?

Zlopez (changes requested):
I have a few things in the comments, but I have a few other things here:
* Could you remove supported python versions from `setup.py`, if you are removing them from tests?
* Could you add news file for the change? See https://anitya.readthedocs.io/en/latest/contributing.html#release-notes (I would say that this will be dev change)
* Also look at the formatting, it's failing in Travis CI.

anitya/mail_logging.py
AdamSaleh: Because the linter complained about psutil being shadowed by the import.

anitya/tests/lib/versions/test_calver.py
AdamSaleh: Fixed it back to the original.

tox.ini
AdamSaleh: Yes.

anitya/lib/versions/rpm.py
AdamSaleh: Great!

.travis.yml
AdamSaleh: Because https://www.python.org/dev/peps/pep-0589/ - Typed Dict has been introduced in 3.8

anitya/mail_logging.py
Zlopez: Ok, in this case fine

.travis.yml
Zlopez: Ok, could you reflect this change in setup.py, it's used for creating metadata for PyPi packages about supported python versions.

news/PR1114.dev
```diff
@@ -0,0 +1,2 @@
+Introduced static-type checking through inclusion of mypy in tox.
```
Zlopez, line 1: I'm not sure if towncrier can process multiline news file. Let's keep this only with mypy update.

Zlopez:
@AdamSaleh Because the F34 is by default running on python 3.9, it would be nice to have this in as soon as possible. I deployed the new version of Anitya on staging today running F34 and it's has plenty of issues and this PR should solve them.

news/PR1114.dev
Zlopez: Could you try to run `towncrier --draft` to test this?

Zlopez: I tried it and it doesn't seems to be an issue.

Zlopez:
I continued on this PR, but there is one thing missing `mypy.cfg`. I found out, that the `*.cfg` was in `.gitignore`, which probably caused this issue.
@AdamSaleh Could you update the PR with `mypy.cfg`? I already fixed everything else, even added mypy job to `.zuul.yaml`. So if `tox -e mypy` was already working, it should be just a question of adding the `mypy.cfg`

Zlopez:
I added https://github.com/fedora-infra/anitya/pull/1153 for Python 3.9 support, so no need to fix this here. I will fix any conflict that will arise from this.

Zlopez:
I rebased the PR to latest master. So the tests are now run against Python 3.9.

Zlopez:
It looks like you rewrote my changes :-/ and I did the same by doing git pull from this PR

Zlopez:
I was able to reset the branch to before `git pull` state and did a merge. Let's see if this fix the issue

Zlopez:
It looks like all the remaining issues are fixed.

PR #1121, opened 2021-06-14
Download dump only when import_production_database is set

abitrolly:


abitrolly:
I am doing disk cleanup and this patch was sitting unmerged there.

Zlopez (approved):
Nice catch

PR #1122, opened 2021-06-16
Fix mention of Freenode

mscherer:


docs/admin-user-guide.rst
```diff
@@ -160,7 +160,7 @@ Basics
 * try to not remove anything that could be of value
 * when removing something, try to create as little impact for others as possible
 * check `flags`_ once in a while
-* if in doubt, ask in #fedora-apps on freenode IRC
+* if in doubt, ask in #fedora-apps on libera IRC
```
Zlopez, line 163: It will be probably better to use libera.chat, because libera is showing multiple unrelated search result. And even better would be to have a link to https://libera.chat/

docs/user-guide.rst
```diff
@@ -491,7 +491,7 @@ Reporting Issues
 
 You can report problems on our `issue tracker`_. The `source code`_ is also
 available on GitHub. The development team hangs out in ``#fedora-apps`` on the
-freenode network. Please do stop by and say hello.
+libera network. Please do stop by and say hello.
```
Zlopez, line 494: It will be probably better to use libera.chat, because libera is showing multiple unrelated search result. And even better would be to have a link to https://libera.chat/

Zlopez (commented):
Same comment as on the-new-hotness

Zlopez (approved):

Zlopez:
Thanks for the contribution :-)

PR #1125, opened 2021-06-22
Add Zuul configuration

### LINKED ISSUE #1124: Migrate to Zuul CI
# Description
Travis-CI is no longer a good option for Anitya, because of the limitation and now when travis-ci.org instance is no longer working we need to find a new CI. Because some other projects are using Zuul and I'm familiar with it, I decided to move over to it.

## Requirements
* Create a Zuul configuration file
* Add Anitya to Zuul CI
* Add checks for Zuul to mergify
* Update documentation
* (Optional) Solve the codecov issue

The Anitya will again have a CI running.

Zlopez:
Closes #1124 


.zuul.yml
```diff
@@ -0,0 +1,50 @@
+#- job:
```
morucci, line 1: let's rename the file to .zuul.yaml

Zlopez: It looks like this helped

.zuul.yaml
```diff
@@ -0,0 +1,160 @@
...
+- project:
+    check:
+      jobs:
+        - tox-lint
+        - tox-format
+        - tox-python37
+        - tox-python38
+# Anitya wasn't tested with python 3.9 yet and there are few failing tests now
+# TODO: Fix tests for python 3.9
+#        - tox-python39
```
AdamSaleh, line 157: I think I have some fixes in my stale pr for mypy setup, so once this is merged I can look into that.

AdamSaleh (approved):
Looks good. Like the use of pods, should be faster than full vm's.

Zlopez:
I want to setup mergify and update docs before merging this

.zuul.yaml
Zlopez: Thanks, this will be nice

I was a little nervous without working CI

Zlopez:
@softwarefactory-project-zuul help

Issue #1127, opened 2021-06-28
Update vagrant environment to newer Fedora


Zlopez:
The vagrant environment now use Fedora 32, we should switch to newer Fedora, because Fedora 32 is already EOL.

PR #1128, opened 2021-06-29
Update vagrant environment to Fedora 34

LenkaSeg:
Should solve issue #1127


requirements.txt
```diff
@@ -2,18 +2,19 @@ alembic>=1.3.1, <2.0.0
...
-pytoml>=0.1.18, <1.0.0
 semver>=2.8.1, <3.0.0
-social-auth-app-flask>=1.0.0, <2.0.0
 social-auth-app-flask-sqlalchemy>=1.0.1, <2.0.0
+social-auth-app-flask>=1.0.0, <2.0.0
+sqlalchemy>=1.3.17, <2.0.0
 sseclient>=0.0.26, <1.0.0
 straight.plugin>=1.5.0, <2.0.0
-sqlalchemy>=1.3.17, <2.0.0
+toml==0.10.2
```
Zlopez, line 17: Don't use exact versions, this could fail if version is not available in Fedora yet
```suggestion
toml>=0.10.2, <1.0.0
```

requirements.txt
```diff
@@ -2,18 +2,19 @@ alembic>=1.3.1, <2.0.0
...
 webargs>=8.0.0, <9.0.0
+werkzeug==0.16.1
```
Zlopez, line 19: Same as above
```suggestion
werkzeug>=0.16.1, <1.0.0
```

requirements.txt
```diff
@@ -2,18 +2,19 @@ alembic>=1.3.1, <2.0.0
...
+flask>=1.1.4, <3.0.0
 jinja2>=2.11.2, <4.0.0
 ordered-set>=3.1, <5.0.0
 python-dateutil>=2.8.0, <3.0.0
-pytoml>=0.1.18, <1.0.0
 semver>=2.8.1, <3.0.0
-social-auth-app-flask>=1.0.0, <2.0.0
 social-auth-app-flask-sqlalchemy>=1.0.1, <2.0.0
+social-auth-app-flask>=1.0.0, <2.0.0
+sqlalchemy>=1.3.17, <2.0.0
```
Zlopez, line 14: Why the changes in ordering?

Zlopez (changes requested):
Two small changes.

Could you also add to commit message, why the requirements were changed? Something like "pytoml is now toml"  etc.

requirements.txt
LenkaSeg: I ran `:sort` to sort stuff alphabetically, so it is quicker to find the keywords. When I add a new item to requirements, I usually write it in the top and then run `sort`.

abitrolly: But `flask` should be placed before `flask-login`, so that `:sort:` is doing wrong.

Zlopez: @abitrolly Yes, I would assume the same. that `flask` should before `flask-login`

LenkaSeg: I think in case there would be whitespaces after the name, the sort would work correctly.

LenkaSeg: Should I add whitespaces so the sort works or reorder it manually and keep it without spaces?

requirements.txt
Zlopez: The werkzeug is not direct dependency it could be removed altogether

requirements.txt
Zlopez: I never used `sort` on requirements file, but it sounds like a good idea. I wonder what will happen if we run `black` on requirements.txt

Zlopez: It can't be parsed by `black` unfortunately

Zlopez: @LenkaSeg Let's use the manual ordering for this PR and if you want, you can create a separate PR for changing the requirements file to be sortable by `sort`

LenkaSeg: Ok, returning it back.

Zlopez (dismissed):
This looks good to me :+1:

Zlopez:
One more thing, could you just update the commit message to describe the changes?

LenkaSeg:
I updated the commit message => pytoml is now toml.

PR #1144, opened 2021-07-25
Fix multiple small (but sometimes serious) issues

oturpe:
Master branch has multiple issues small (but sometimes very serious).

## Vagrant development environment setup failed during 'vagrant up'
Errors during from task _Build the Anitya documentation_:

> user-guide.rst:210:Inline strong start-string without end-string.
> user-guide.rst:3:Duplicate explicit target name: \"sourceforge.net\"

Like the errors say, the entry about _Sourceforge (git)_ was corrupted
in two different ways:
1. Strong string was started with `**`, but never terminated.
   The fix is obvious.
2. The same document had two links using format '\`text <url>\`_',
   which is not allowed in reStructured Text.
   Fixed by using '__' in the second link, making it "anonymous".

## Fix corrupted javascript
Caused the page it was on malfunction in serious ways.

##  Display error correctly when Test Check fails
When Test Check failed in the Edit Project page,
resulting error was displayed as '[Object object]'.
This was a result of concatenating an object to string.
Fixed by using JSON.stringify().

### DIFF
--- a/anitya/templates/project_new.html
+++ b/anitya/templates/project_new.html
@@ -263,7 +263,7 @@
         "'https://packagist.org/packages/phpunit/php-code-coverage' " +
         "this field should contain 'phpunit/php-code-coverage'.");
     };
-    if (str == 'Sourceforge') || (str == 'Sourceforge (git)){
+    if ((str == 'Sourceforge') || (str == 'Sourceforge (git)')){
       $("label[for='version_url']").text('Sourceforge name');
       $('#version_url').prop('title', "Use this field if the name used " +
         "on the rss page of the project differs from the name of the project." );
--- a/anitya/templates/project_new.html
+++ b/anitya/templates/project_new.html
@@ -379,7 +379,7 @@
           $(btn).show(); $(btn + "-spinner").hide();
           alert(
               'Unable to retrieve the latest version from upstream!\n'
-              + 'ERROR: ' + res.responseJSON );
+              + 'ERROR: ' + JSON.stringify(res.responseJSON) );
         }
     });
   };
--- a/docs/user-guide.rst
+++ b/docs/user-guide.rst
@@ -207,8 +207,8 @@ The backends available are:
   You need to provide **Sourceforge name** if the name on RSS feed is different then the
   project name on Sourceforge.
 
-* **Sourceforge (git) for projects hosted on
-  `sourceforge.net <https://sourceforge.net>`_
+* **Sourceforge (git)** for projects hosted on
+  `sourceforge.net <https://sourceforge.net>`__
 
   Sourceforge (git) retrieves the git tags in case a project hosted on sourceforge has them.
   When a url to get the tags from is not provided, the sourceforge homepage of a project
--- a/news/PR1144.bug
+++ b/news/PR1144.bug
@@ -1,0 +1,1 @@
+Fix documentation and javascript issues


docs/user-guide.rst
```diff
@@ -207,8 +207,8 @@ The backends available are:
   You need to provide **Sourceforge name** if the name on RSS feed is different then the
   project name on Sourceforge.
 
-* **Sourceforge (git) for projects hosted on
-  `sourceforge.net <https://sourceforge.net>`_
+* **Sourceforge (git)** for projects hosted on
```
Zlopez, line 210: I'm surprised this wasn't caught by the tests. It was recently added in https://github.com/fedora-infra/anitya/pull/1134.

anitya/templates/project_new.html
```diff
@@ -377,7 +377,7 @@ <h1>{{ context }} project</h1>
           $(btn).show(); $(btn + "-spinner").hide();
           alert(
               'Unable to retrieve the latest version from upstream!\n'
-              + 'ERROR: ' + res.responseJSON );
+              + 'ERROR: ' + JSON.stringify(res.responseJSON) );
```
Zlopez, line 380: I'm not that familiar with javascript, so this will be probably my error. Thanks for fixing this.

Zlopez (changes requested):
The changes are looking good, could you just add a news file to this PR?

Name it `PR1144.bug` and add it to `news` folder`. The text could be something like `Fix documentation and javascript issue`

docs/user-guide.rst
oturpe: I was too. I looked at it again, Zuul configuration is wrong,
job `anitya-tox-docs` is actually running Bandit.
Probably just a copy paste error and easy to fix?

Otherwise, Vagrant uses the Makefile in the `docs` folder
while Zuul runs some commands embedded inside `tox.ini`.
That could cause varying results, too.

```
- job:
    name: anitya-tox-docs
    ...
    parent: tox
    run: ci/test-tox.yaml
    vars:
      tox_env: bandit
    ...
```

anitya/templates/project_new.html
oturpe: You are not the only one :laughing:
This is a very common mistake in error handlers.
The default string resulting from adding an object to string is the useless `[Object object]`
and it is even easier to accidentally write that kind of code
because e.g. `console.log(object)` actually prints something useful.

docs/user-guide.rst
Zlopez: This really looks like a copy paste error on my side. I will fix it. I will also unify how the docs are built in vagrant and in tox.

oturpe:
> The changes are looking good, could you just add a news file to this PR?
> 
> Name it `PR1144.bug` and add it to `news` folder`. The text could be something like `Fix documentation and javascript issue`

Uh, sorry. I was not entirely sure if there is required or encouraged for every pull request. Fixed.

Zlopez:
> > The changes are looking good, could you just add a news file to this PR?
> > Name it `PR1144.bug` and add it to `news` folder`. The text could be something like `Fix documentation and javascript issue`
> 
> Uh, sorry. I was not entirely sure if there is required or encouraged for every pull request. Fixed.

I don't create them, if this is only about fixing tests or some small changes in dev environment, but it's good to have them if this is anything that affects users of release-monitoring.org

Zlopez:
Let me first fix the documentation build so I can see that it really fails, then I will merge this PR.

anitya/templates/project_new.html
abitrolly: That's why people are using some JavaScript replacements on serious projects?

Zlopez: It would be nice to get rid of Javascript entirely, but I'm not frontend developer, so this is not my cup of tea.

docs/user-guide.rst
Zlopez: I looked at the makefile and the docs tox target and they are almost identical, the tox one is even running `./generate_db_schema`. So I would just fix the zuul file.

Zlopez:
The Zuul PR is #1147

Zlopez:
The PR #1147 failed, so let's merge this one and see if it will pass.

Zlopez (approved):

LenkaSeg:
@oturpe Thanks for the fixes! I missed it totally.

PR #1146, opened 2021-07-25
Use version filter during Edit Project Test Check

### LINKED ISSUE #1143: Version Filter not applied on Test Check
While Version Filter is correctly applied when a project is submitted,
it is not applied for the _Test check_ in the add/edit screen.
This makes is unnecessarily difficult to assert that the filter is configured correctly.

## To reproduce
Using the Munt project as an example,
since it uses a monorepo where the different component releases
are all mixed up in GitHub releases of a single repo.

### Step 1: No filtering
Enter the following items
* **Project name**: mt32emu-qt
* **Homepage**: http://munt.sourceforge.net/
* **Backend**: GitHub
* **Version scheme**: RPM
* **GitHub owner/projec**t: munt/munt

Click _Test check_

Observe many results of format `<prefix>_<number_with_underscores>`

### Step 2: Add version prefix
Add the following data:
* **Version prefix**: mt32emu_qt_

Click _Test check_

Observe same results as before,
with entries matching the given prefix
converted to plain `<number_with_underscores>`.

### Step 3: Add version filter
Add the following data which should filter additional entries:
* **Version filter**: munt_,mt32emu_win32drv_,mt32emu_smf2wav_,libmt32emu_

Click _Test check_

Observe that the results did not change.

If you proceed to _Submit_, the filter is applied correctly in the actual results.
Only the _Test check_ is misbehaving.

oturpe:
The version filter input is on the page, api accepts it,
but it was not used.
Fixed by simply passing it in.
Now Test Check accurately shows the versions
resulting from  configuration on the page.

Fixes #1143

Signed-off-by: Otto Urpelainen <oturpe@iki.fi>

### DIFF
--- a/anitya/templates/project_new.html
+++ b/anitya/templates/project_new.html
@@ -342,6 +342,7 @@
     var _version_url = $('#version_url').val();
     var _version_prefix = $('#version_prefix').val();
     var _pre_release_filter = $('#pre_release_filter').val();
+    var _version_filter = $('#version_filter').val();
     var _regex = $('#regex').val();
     var _releases_only = $('#releases_only').is(':checked');
     var _insecure = $('#insecure').is(':checked');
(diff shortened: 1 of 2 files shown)


Zlopez (approved):
Thanks for this fix, I didn't knew this issue is caused by something trivial like not sending the variable.

PR #1147, opened 2021-07-26
Use correct tox_env for docs job

Zlopez:

### DIFF
--- a/.zuul.yaml
+++ b/.zuul.yaml
@@ -103,7 +103,7 @@
     parent: tox
     run: ci/test-tox.yaml
     vars:
-      tox_env: bandit
+      tox_env: docs
       dependencies:
         - graphviz
         - python3-sphinxcontrib-httpdomain


Zlopez:
@oturpe Hm, it seems there is still an issue in the docs build. Something about autoflask. I will try to fix this.

Zlopez:
I can reproduce the issue in local vagrant machine, let's see if I will be able to fix it.

Zlopez:
I created https://github.com/sphinx-contrib/httpdomain/issues/53, because I don't see any obvious issue in the documentation.

oturpe:
I did not get that error yesterday when I ran `vagrant up` to work on #1143,
and I did not get it today when I recreated the Vagrant machine to reproduce this.
I do see it if I run `tox -e docs`, however.
Sphinx is not familiar to me, I have no idea what might cause that.

Zlopez:
I know that the issue is related to `sphinx-contrib.httpdomain` library, but I'm not sure what exatcly causing it. And why this is not thrown when setting up vagrant. Maybe the python version is different in vagrant machine and tox env.

oturpe:
By trial and error: it is the Sphinx version. 4.0.3 still works well, 4.1.1 not.
The Vagrant machine gets its Sphinx from Fedora 34 repositories that are currently at 3.4.3,
tox uses `test_requirements.txt` where 4.1.1 is specified.
Simply downgrading test_requirements.txt to 4.1.0 fixes the issue.

I suppose you thought you can rely on CI to check dependabot's suggestions
but for this particular dependency,
the copy paste error in the CI scripts meant nothing was checked.

I could make a PR, but probably it is easier if you just fix this in this one?

Zlopez:
I thought it could be some dependency that was updated during the usage of wrong tox_env in CI. @oturpe If you can create the PR, it would be nice. :+1:

oturpe:
> I thought it could be some dependency that was updated during the usage of wrong tox_env in CI. @oturpe If you can create the PR, it would be nice. +1

#1149

Zlopez:
Let's see if the test will pass now.

PR #1149, opened 2021-07-26
Downgrade Sphinx to compatible version 4.0.3

### LINKED ISSUE #1148: Documentation generation fails due to "Problem in http domain"
When documentation is generated though Tox `docs` task,
the following error happens:

```
$ tox -r -e docs
...
reading sources... [ 90%] release-notes
reading sources... [100%] user-guide
Warning, treated as error:
<autoflask>:1:Problem in http domain: field is supposed to use role 'obj', but that role is not in the domain.
ERROR: InvocationError for command /home/otto/src/Ohjelmistot/anitya/.tox/docs/bin/sphinx-build -W -b html -d /home/otto/src/Ohjelmistot/anitya/.tox/docs/tmp/doctrees . _build/html (exited with code 2)
__________________________________________________ summary ___________________________________________________
ERROR:   docs: commands failed
```

oturpe:
On Sphinx 4.1.1, Tox `docs` task fails due to following kind of error:
    
```
autoflask>:1:Problem in http domain: field is supposed to use role 'obj', but that role is not in the domain.
```
    
This does not happen on Sphinx 4.0.3, so downgrading to that.
    
Fixes #1148 

Also `.gitignore` improvement.

### DIFF
--- a/news/1148.bug
+++ b/news/1148.bug
@@ -1,0 +1,1 @@
+Downgrade Sphinx to compatible version 4.0.3
--- a/test_requirements.txt
+++ b/test_requirements.txt
@@ -14,6 +14,7 @@ requests==2.26.0
 requests-oauthlib==1.3.0
 
 # Required to test building the docs
-sphinx==4.1.1
+# sphinxcontrib-httpdomain errors on Sphinx 4.1.x 
+sphinx==4.0.3
 sphinxcontrib-httpdomain==1.7.0
 sqlalchemy_schemadisplay==1.3


.gitignore
```diff
@@ -3,6 +3,7 @@ build/
 dist/
 docs/_build/
 docs/docblocks/
+docs/images/
```
Zlopez, line 6: We don't want to ignore this folder, otherwise the database diagram is missing in documentation on readthedocs.

Zlopez (changes requested):
Remove the docs/images/database.png, this file should be uploaded only if there is any change in database.

Also don't ignore the folder that contains it, otherwise the images will be missing in the documentation. We could probably prevent that by adding the ./generate_database_image script to Makefile for docs building.

Zlopez (approved):

abitrolly:
> Remove the docs/images/database.png, this file should be uploaded only if there is any change in database.

The best way probably would be to generate the image in reproducible way.

oturpe:
> Remove the docs/images/database.png, this file should be uploaded only if there is any change in database.
> 
> Also don't ignore the folder that contains it, otherwise the images will be missing in the documentation. We could probably prevent that by adding the ./generate_database_image script to Makefile for docs building.

Ok, I dropped the commit related to these.
It is just that when you run `tox -e docs`,
that file changes and Git starts showing it as modified.
I took this a case of missing `.gitignore`,
now I see that things were intentionally set up this way.

Btw. you could both commit the file **and** have it in `.gitignore`.
That works, but because it is not a common pattern,
it might just cause more confusion than
an autogenerated file showing up in `git status` as modified does.

abitrolly:
> It is just that when you run tox -e docs, that file changes and Git starts showing it as modified.

It appears that on my F34 the `docs/images/database.png` is same even if I regenerate it. How do you run it? Can you attach your version here?

Zlopez:
The output is same, but I think that the metadata are different, so this is why git thinks that this is always a new version.

abitrolly:
Where the metadata is stored?

Zlopez:
Directly in PNG, just check the the file with `exiftool`
```
$ exiftool database.png 
ExifTool Version Number         : 12.26
File Name                       : database.png
Directory                       : .
File Size                       : 31 KiB
File Modification Date/Time     : 2021:02:21 13:33:25+01:00
File Access Date/Time           : 2021:07:25 09:48:01+02:00
File Inode Change Date/Time     : 2021:05:06 12:30:19+02:00
File Permissions                : -rw-r--r--
File Type                       : PNG
File Type Extension             : png
MIME Type                       : image/png
Image Width                     : 453
Image Height                    : 719
Bit Depth                       : 8
Color Type                      : RGB with Alpha
Compression                     : Deflate/Inflate
Filter                          : Adaptive
Interlace                       : Noninterlaced
Background Color                : 255 255 255
Image Size                      : 453x719
Megapixels                      : 0.326
```

abitrolly:
@Zlopez no, these fields are not stored in .png
```
File Modification Date/Time     : 2021:02:21 13:33:25+01:00
File Access Date/Time           : 2021:07:25 09:48:01+02:00
File Inode Change Date/Time     : 2021:05:06 12:30:19+02:00
```
I regenerated the image from scratch and `git` doesn't detect any changes. There must be something else here. Perhaps a problem with database.

database

Zlopez:
Hm, in this case I'm not sure why git is thinking that the image is sometime new even if there wasn't any change in the database.

PR #1175, opened 2021-09-03
Add Python (PEP 440) versioning scheme

encukou:
Fixes: https://github.com/fedora-infra/anitya/issues/1015

I hope I did everything right :)


Zlopez (commented):
Just noticed one small issue. Otherwise this looks great. Thanks for the contribution. :+1:

Zlopez:
Mypy issue was solved by https://github.com/fedora-infra/anitya/pull/1178, but you should look at the failure of diff-cover test.

requirements.txt
```diff
@@ -8,6 +8,7 @@ flask-login>=0.4.1, <1.0.0
 flask-wtf>=0.14.2, <1.0.0
 jinja2>=2.11.2, <4.0.0
 ordered-set>=3.1, <5.0.0
+packaging>=20.0  # (calendar-based version)
```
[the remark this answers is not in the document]
encukou: It's a PEP-440 version like all Python dependencies. But PEP-440 defines comparisons for both calendar and semver-like versioning schemes.
I put the comment here to note that an upper limit like `<21.0` wouldn't make sense. I'll adjust the  comment to make that clearer.

encukou:
Thanks!
I expanded the tests to cover the missed code (and more)

Zlopez (approved):

anitya/tests/lib/versions/test_python.py
```diff
@@ -0,0 +1,279 @@
...
+
+    def test_lt_rc_vs_a(self):
+        """Assert rc prereleases are greater than pre prereleases of the same version."""
+        old_version = python.PythonVersion(version="1.0.0a2")
+        new_version = python.PythonVersion(version="1.0.0rc1")
+        self.assertTrue(old_version < new_version)
+        self.assertFalse(new_version < old_version)
+
+    def test_lt_alpha_vs_beta(self):
+        """Assert alpha prereleases are greater than beta prereleases of the same version."""
```
hroncok, line 156:
```suggestion
        """Assert rc prereleases are greater than alpha prereleases of the same version."""
```

anitya/tests/lib/versions/test_python.py
```diff
@@ -0,0 +1,279 @@
...
+        self.assertEqual(str(version), "1.1rc1")
+        self.assertEqual(version, python.PythonVersion(version="1.1rc1"))
+
+    def test_normalization_integer(self):
+        """Assert that numeric parts are normalized."""
+        version = python.PythonVersion(version="0900.00+foo0100")
+        self.assertEqual(str(version), "900.0+foo0100")
+
+    def test_normalization_prerelease_spelling(self):
+        """Assert that alternate prerelease speling is accepted."""
```
hroncok, line 251:
```suggestion
        """Assert that alternate prerelease spelling is accepted."""
```

anitya/tests/lib/versions/test_python.py
```diff
@@ -0,0 +1,279 @@
...
+            ("beta", "b"),
+            ("c", "rc"),
+            ("pre", "rc"),
+            ("preview", "rc"),
+        ):
+            version = python.PythonVersion(version="1.0" + alt)
+            self.assertEqual(str(version), "1.0" + normal + "0")
+
+    def test_normalization_postrelease_spelling(self):
+        """Assert that alternate postrelease speling is accepted."""
```
hroncok, line 263:
```suggestion
        """Assert that alternate postrelease spelling is accepted."""
```

hroncok:
When can we expect this on https://release-monitoring.org/ ?

Zlopez:
@hroncok I'm currently waiting till [this PR](https://github.com/fedora-infra/anitya/pull/1220) will be finished and merged. Current version of Anitya in master branch doesn't allow users to log in because of the issue in `social_auth` library. We need to migrate to another library before we can release a new version.

PR #1182, opened 2021-09-19
Better error message when GitHub token is missing

oturpe:
In order to access GitHub during development,
GitHub access token needs to be manually set
in the configuration file.
This is unavoidable, since developer's own token is needed.
On the other hand,
the error message about missing token was not very clear
because it was just 401 Unauthenticated from GitHub.
A special error message for the case of unset token is added.
This tells developers clearly what is wrong in their setup
and is not worse than 401 in a production setup.


Zlopez (approved):
This looks great, thanks you

Issue #1139, opened 2021-07-20
Migrate from social_auth to something else


Zlopez:
The [social_auth-flask](https://github.com/python-social-auth/social-app-flask) is unmaintained for some time, we should look for the alternative that is ideally packaged in Fedora.

Zlopez:
According to #1145 this is now blocking the authentication process even in vagrant dev env. I'm giving this issue a high priority and we should solve it as soon as possible.

LenkaSeg:
What about Authlib as a replacement

https://github.com/lepture/authlib

https://python.libhunt.com/authlib-alternatives

It's
- with openid
- packaged in Fedora: https://src.fedoraproject.org/rpms/python-authlib

LenkaSeg:
Related PR #1220

PR #1220, opened 2021-11-12
Migrate social_auth to authlib

LenkaSeg:
Allow authentication with Fedora, GitHub and Google.
This PR aims to get rid of the social_auth library and use authlib instead.

Related issue #1139

### DIFF
--- a/anitya/auth.py
+++ b/anitya/auth.py
@@ -1,0 +1,59 @@
+import flask
+import flask_login
+
+from anitya.db import User, Session
+
+
+def create_oauth_blueprint(oauth):
+    oauth_blueprint = flask.Blueprint('oauth', __name__)
+
+    @oauth_blueprint.route("/oauthlogin/<name>")
+    def oauthlogin(name):
+        client = oauth.create_client(name)
+        if client is None:
+            flask.abort(400)
+        redirect_uri = flask.url_for('.auth', name=name, _external=True)
+        return client.authorize_redirect(redirect_uri)
+
+
+    @oauth_blueprint.route("/auth/<name>")
+    def auth(name):
+        client = oauth.create_client(name)
+        if client is None:
+            flask.abort(400)
+        id_token = flask.request.values.get('id_token')
+        if flask.request.values.get('code'):
+            token = client.authorize_access_token()
+            if id_token:
+                token['id_token'] = id_token
+        elif id_token:
+            token = {'id_token': id_token}
+
+        # for Google
+        if 'id_token' in token:
+            user_info = client.parse_id_token(token)
+        # for Github
+        else:
+            client.response = client.get('user', token=token)
+            user_info = client.response.json()
+            if user_info['email'] is None:
+                resp = client.get('user/emails', token=token)
+                resp.raise_for_status()
+                data = resp.json()
+                user_info['email'] = next(email['email'] for email in data if email['primary'])
+
+        # Check if the user exists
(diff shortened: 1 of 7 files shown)


LenkaSeg:
What is working so far: authentication with GitHub and Google.
What is needed: Fedora, create a new social table in models, do something with the next, so it does not redirect to main page, tests, documentation.

requirements.txt
```diff
@@ -11,7 +11,6 @@ ordered-set>=3.1, <5.0.0
 packaging>=20.0  # (no upper limit, this is a calendar-based version)
 python-dateutil>=2.8.0, <3.0.0
 semver>=2.8.1, <3.0.0
-social-auth-app-flask>=1.0.0, <2.0.0
```
Zlopez, line 14: It would be nice to add the authlib to requirements.txt as well.

ansible/roles/anitya-dev/tasks/main.yml
```diff
@@ -42,6 +42,7 @@
            fedora-messaging,
            python3-alembic,
            python3-arrow,
+           python-authlib,
```
Zlopez, line 45: It's nice that the authlib is available in Fedora, one less package to download from PyPi.

anitya/config.py
```diff
@@ -66,32 +66,27 @@
...
     r"\.(?:tar|t[bglx]z|tbz2|zip)",
-    # Token for GitHub API
-    GITHUB_ACCESS_TOKEN=None,
     SSE_FEED="http://firehose.libraries.io/events",
     CRON_POOL=10,  # Number of workers for check service
     CHECK_TIMEOUT=600,  # Timeout for check service
     # When this number of failed checks is reached,
     # project will be automatically removed, if no version was retrieved yet
     CHECK_ERROR_THRESHOLD=100,
+    # Enabled authentication backends
```
Zlopez, line 79: This should be also done in the example config and config for vagrant and container setup.

anitya/auth.py
```diff
@@ -0,0 +1,59 @@
+import flask
```
Zlopez, line 1: This file deserves some comments. :-)

LenkaSeg: I saw that the social_auth stuff was registered as flask blueprint in the `app.py`, so I was trying to make it similar.
The rest is from authlib examples. The `def oauthlogin` is usually named `login` in the authlib documentation. I renamed it because there already is login function in ui.py.
Then I created the tokens in github and google for the anitya app and since both parties send a different user_info, so it is processed differently. 
I will add docstrings and comments.

romulasry:
This would be nice for https://rpmfusion.org/ next. Good job.

Zlopez:
Because there is now initiative working on Flask OIDC to remove the deprecated parts, we should probably use it instead of using authlib directly.

romulasry:
https://github.com/puiterwijk/flask-oidc

Zlopez:
So here is the list of things I need to finish after initial work with this PR:

- [ ] Update documentation
- [ ] Update default config file
- [ ] Add Fedora oauth support
- [x] Remove social_auth tables
- [ ] Create alembic migration for social_auth tables

I found out that the `social_auth` tables are no longer needed as the e-mail is enough for authlib.

After merging this I will probably release a new major version, as the removal of `social_auth` tables is backwards compatibility breaking change.

Zlopez:
I would need to review all the changes, but I think we are ready to go.

PR #1272, opened 2022-01-19
Only include unyanked crate versions

carlwgeorge:
crates.io allows maintainers to "yank" a crate.  When this happens, the crates backend should not include the yanked version.

https://doc.rust-lang.org/cargo/commands/cargo-yank.html

### DIFF
--- a/anitya/lib/backends/crates.py
+++ b/anitya/lib/backends/crates.py
@@ -214,7 +214,8 @@ class CratesBackend(BaseBackend):
                 was in an unexpected format.
         """
         filtered_versions = cls.filter_versions(
-            [v["num"] for v in cls._get_versions(project)], project.version_filter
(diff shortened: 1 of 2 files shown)


Zlopez:
Thanks for the PR! Could you check the failing test and add [news file](https://anitya.readthedocs.io/en/latest/contributing.html#release-notes) for this change?

Zlopez (approved):

decathorpe:
Hi @carlwgeorge ! Sorry for taking so long to reply, your email did not get lost :)

Thanks for filing this PR, looks good from my side.

I just have one question: Once a version has been added to the database, it won't get removed again, if it becomes "yanked" at some later point, will it?

carlwgeorge:
No, once anitya sees a version, it's saved even if it changes to yanked status later.  But such a version can be flagged for removal by the admins.

PR #1417, opened 2022-07-25
Use Homepage from PyPi package

JakobDev:
With this PR anitya uses the Homepage that the author of the PyPi package has set rather than the PyPI page.

Note:
I don't know how to test this, therefore it has not been tested yet.

### DIFF
--- a/anitya/lib/backends/pypi.py
+++ b/anitya/lib/backends/pypi.py
@@ -150,5 +150,15 @@ class PypiBackend(BaseBackend):
         for entry in items:
             title = entry["title"]["value"]
             name, version = title.rsplit(None, 1)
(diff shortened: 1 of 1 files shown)


Zlopez (commented):
This method isn't actually called anywhere in Anitya. It was in past but not in the current version. The current release of Anitya is using [libraries.io](https://github.com/fedora-infra/anitya/blob/master/anitya/librariesio_consumer.py) as the source to look for new projects to add.

The code you wrote is good, but I'm not sure if we have a place in Anitya project, where this could be currently used. The backends in current state are only handling versions and not much more, although it would be nice if we could normalize the homepage based on the backend.

Issue #1013, opened 2021-01-27
GitHub backend's "Check releases instead of tags" does not work


michel-slm:
Tested on https://release-monitoring.org/project/4199/

If I edit the project, set the backend to GitHub, and check this box, the version check URL still uses tags rather than releases

PR #1434, opened 2022-08-14
fix get_version_url in github backend

erolkskn:
Resolves: #1013 

Also while i was fixing bug that mentioned in #1013 , i noticed that "/tags" value is always replaced by "" https://github.com/fedora-infra/anitya/blob/c94a0b4088fe4542c7cf7d483bd1beffd4b2154c/anitya/lib/backends/github.py#L158. This can lead to problems if given project's owner or repo name is "tags".  In order to prevent this problem, i've replaced `url.replace("/tags", "")` with `url.rstrip("/tags")`. Also added tests for it.

### DIFF
--- a/anitya/lib/backends/github.py
+++ b/anitya/lib/backends/github.py
@@ -88,7 +88,9 @@ class GithubBackend(BaseBackend):
             url = url[:-1]
 
         if url:
-            url = "https://github.com/{}/tags".format(url)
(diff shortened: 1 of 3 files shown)


Zlopez (approved):
This is an awesome addition, thanks for the PR

erolkskn:
> This is an awesome addition, thanks for the PR

Thank you for your comment :)

Issue #1463, opened 2022-09-15
expose "additional" support for setting project properties only available in `POST /api/v2/versions/` in `POST /api/v2/projects/`


decathorpe:
It looks like it is currently impossible to create a project *and* set, for example, a `pre_release_filter` with one POST request, since this argument is not accepted when *creating* a project, only very counter-intuitively by *editing* it by sending a POST request to the `api/v2/versions/` endpoint.

The following properties are missing from the `POST /api/v2/projects/` endpoint:

- `version_scheme`
- `version_pattern`
- `pre_release_filter`
- `version_filter`
- `releases_only`

I don't know why these are missing from the "create a new project" endpoint, since they are present in both the Web UI for creating a new project, *and* in the "edit an existing project" functionality that's hidden behind the `api/v2/versions/` endpoint.

I'd like to create projects via the API, but I need to set `version_scheme="Semantic"` and `version_filter="alpha;beta;rc;pre"` when creating them, otherwise I might need release-monitoring.org administrator help to clean up some versions that are recognised incorrectly, and it looks like I can only do that with the Web UI right now.

---

Side note: I dug through the source code of those endpoints already to figure this out, and it looks like adding these settings to the `POST /api/v2/projects/` endpoint should be as simple as adding the new fields to the accepted arguments and to the `utilities.create_project()` call, which is the same in both endpoints. If you want, I can submit a PR for this. :)

Zlopez:
The reason for those properties missing from `POST /api/v2/projects/` endpoint is simple. These were introduced after the endpoint was created and it was never updated.
The `api/v2/versions/` was created much later, and this is why it contains these changes.

You are right adding those should be simple enough, just allowing new optional fields to the endpoint.

PR #1465, opened 2022-09-18
add missing optional parameters to API v2 project creation endpoint

decathorpe:
Fixes: #1463

### DIFF
--- a/anitya/api_v2.py
+++ b/anitya/api_v2.py
@@ -449,7 +449,7 @@ class ProjectsResource(MethodView):
         :reqjson string backend: The project backend (github, folder, etc.).
         :reqjson string version_url: The URL to fetch when determining the
                                      project version (defaults to null).
-        :reqjson string version_scheme: The project version scheme.
+        :reqjson string version_scheme: The project version scheme. If missing, it's set to "RPM".
         :reqjson string version_pattern: The version pattern for calendar version scheme.
         :reqjson string version_prefix: The project version prefix, if any. For
                                         example, some projects prefix with "v".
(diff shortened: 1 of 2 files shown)


anitya/api_v2.py
```diff
@@ -471,7 +475,11 @@ def post(self):
             "homepage": fields.Str(required=True),
             "backend": fields.Str(required=True),
             "version_url": fields.Str(missing=None),
+            "version_scheme": fields.Str(missing=None),
```
Zlopez, line 478:
```suggestion
            "version_scheme": fields.Str(missing="RPM"),
```
`version_scheme` should be always set, it's not required field in the API, but it's set to RPM in `create_project` function if not provided. Let's set it to "RPM" if it's not provided.

anitya/api_v2.py
```diff
@@ -449,8 +449,12 @@ def post(self):
         :reqjson string backend: The project backend (github, folder, etc.).
         :reqjson string version_url: The URL to fetch when determining the
                                      project version (defaults to null).
+        :reqjson string version_scheme: The project version scheme.
```
Zlopez, line 452:
```suggestion
        :reqjson string version_scheme: The project version scheme. If missing, it's set to "RPM".
```

Zlopez (changes requested):
One thing that should be changed, otherwise looks fine.

anitya/api_v2.py
decathorpe: Makes sense! Thanks for the review.

Zlopez (approved):

decathorpe:
Awesome, thanks for merging!

Issue #1402, opened 2022-07-04
Python versioning parse doesn't return string


Zlopez:
New python versioning scheme returns some special version object instead of string. This is causing the following issue:
```
(psycopg2.ProgrammingError) can't adapt type 'Version'
[SQL: UPDATE projects SET latest_version=%(latest_version)s, updated_on=CURRENT_TIMESTAMP WHERE projects.id = %(projects_id)s]
[parameters: {'latest_version': <Version('0.8.2')>, 'projects_id': 83548}]
(Background on this error at: https://sqlalche.me/e/14/f405)
Traceback (most recent call last):
File "/usr/lib64/python3.10/site-packages/sqlalchemy/engine/base.py", line 1819, in _execute_context
self.dialect.do_execute(
File "/usr/lib64/python3.10/site-packages/sqlalchemy/engine/default.py", line 732, in do_execute
cursor.execute(statement, parameters)
psycopg2.ProgrammingError: can't adapt type 'Version'
```

The `parse` method from `BaseVersion` shouldn't be overloaded and instead these conversions should be done inside the compare methods.

nikitabugrovsky:
@Zlopez could you please review submitted bug fix?

PR #1517, opened 2022-12-11
[#1402] Fix Python versioning parse doesn't return string

nikitabugrovsky:
Bug affects specific version scheme `Python (PEP 440)`

Bug Example

Fix:
Return a string when invoking parse method of the PythonVersion class. When prerelease & postrelease methods are called, Version object is used.

Version handling reference:
https://packaging.pypa.io/en/latest/version.html

### DIFF
--- a/anitya/lib/versions/python.py
+++ b/anitya/lib/versions/python.py
@@ -44,7 +44,7 @@ class PythonVersion(base.Version):
 
     name = "Python (PEP 440)"
 
-    def parse(self):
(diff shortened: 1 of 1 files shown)


nikitabugrovsky:
@Zlopez strangely all the tox tests run successfully on the dev env...

Zlopez:
`POST_FAILURE` isn't a problem with the test, but with CI. Let me rerun it.

PR #1519, opened 2022-12-12
[#1402] Add news bit on bug 1402

nikitabugrovsky:


nikitabugrovsky:
@Zlopez sending a news bit as a separate PR as you requested.

PR #1521, opened 2022-12-12
Improve Vagrant Development Environment

### LINKED ISSUE #1520: Improve Vagrant Development Environment
- Update development environment dependencies
Dependencies should include @virtualization package group

- Docs are not compiling via ansible task

Compilation error
```
Problem in http domain: field is supposed to use role 'obj', but that role is not in the domain
```
-  Exec Paths of librariesio_consumer.py & check_service.py services in systemd service files  are not correct

nikitabugrovsky:
Closes #1520

### DIFF
--- a/ansible/roles/anitya-dev/files/check-service.service
+++ b/ansible/roles/anitya-dev/files/check-service.service
@@ -5,7 +5,7 @@ Documentation=https://github.com/fedora-infra/anitya/
 
 [Service]
 Environment="ANITYA_WEB_CONFIG=/home/vagrant/anitya.toml"
-ExecStart=check_service.py
+ExecStart=/usr/local/bin/check_service
 Type=simple
 
 [Install]
--- a/ansible/roles/anitya-dev/files/librariesio-consumer.service
+++ b/ansible/roles/anitya-dev/files/librariesio-consumer.service
@@ -4,7 +4,7 @@ After=network.target
 
 [Service]
 Environment="ANITYA_WEB_CONFIG=/home/vagrant/anitya.toml"
-ExecStart=librariesio_consumer.py
+ExecStart=/usr/local/bin/librariesio_consumer
 Type=simple
 Restart=on-failure
 
--- a/news/1402.bug
+++ b/news/1402.bug
@@ -1,0 +1,1 @@
+Fix Python versioning parse doesn't return string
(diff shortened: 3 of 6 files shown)


news/1402.bug
```diff
@@ -0,0 +1 @@
+Fix Python versioning parse doesn't return string
```
Zlopez, line 1: This file shouldn't be here.

ansible/roles/anitya-dev/files/librariesio-consumer.service
```diff
@@ -4,7 +4,7 @@ After=network.target
 
 [Service]
 Environment="ANITYA_WEB_CONFIG=/home/vagrant/anitya.toml"
-ExecStart=librariesio_consumer.py
+ExecStart=/home/vagrant/devel/anitya/librariesio_consumer.py
```
Zlopez, line 7: You shouldn't need that the script should be in standard PATH, it is installed as binary

ansible/roles/anitya-dev/files/check-service.service
```diff
@@ -5,7 +5,7 @@ Documentation=https://github.com/fedora-infra/anitya/
 
 [Service]
 Environment="ANITYA_WEB_CONFIG=/home/vagrant/anitya.toml"
-ExecStart=check_service.py
+ExecStart=/home/vagrant/devel/anitya/check_service.py
```
Zlopez, line 8: You shouldn't need that the script should be in standard PATH, it is installed as binary

Zlopez (commented):
I'm not sure why `aniyta/check_service.py` and `anitya/librariesio_consumer.py` are even in this PR and I also added few other comments

ansible/roles/anitya-dev/files/librariesio-consumer.service
nikitabugrovsky: yes, you are right, changed path to scrip, provided by binary install

ansible/roles/anitya-dev/files/check-service.service
nikitabugrovsky: yes, you are right, changed path to the script, provided by binary install

nikitabugrovsky:
@Zlopez could you kindly take another look?

nikitabugrovsky:
Zlopez done!

nikitabugrovsky:
Looks like I did not pick up upstream changes before rebasing. Pushed again.

Zlopez:
It's still not correct, I can see 2 commits in the https://github.com/fedora-infra/anitya/pull/1521/commits, the news file is one of them.

Zlopez:
It's possible that you created this branch from different branch than master?

nikitabugrovsky:
I guess I just need to recreate PR... Locally I o

PR #1528, opened 2023-01-03
Fix Containerfile for Fedora 37 / Python 3.11

der-eismann:
Fedora 37 has Python 3.11, so the hardcoded path won't work. Verified with Podman 4.3.1

```
STEP 19/21: RUN sed -i 's/base64.encodestring/base64.encodebytes/g' /usr/local/lib/python3.*/site-packages/social_sqlalchemy/storage.py
--> Using cache f4243073de8a091c6a777315ef79d5f7da69cd1ff95aaa32d7237d1f719f8464
--> f4243073de8
STEP 20/21: RUN cat /usr/local/lib/python3.11/site-packages/social_sqlalchemy/storage.py | grep base64
import base64
    secret = Column(String(255))  # base64 encoded
        assoc.secret = base64.encodebytes(association.secret).decode()
```

### DIFF
--- a/Containerfile.dev
+++ b/Containerfile.dev
@@ -39,6 +39,6 @@ RUN pip install dist/*.whl
 
 # Hotfix for social_auth-sqlalchemy
 # Should be removed when we switch to something else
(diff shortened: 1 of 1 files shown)


Zlopez:
Thanks for opening this, I will try to migrate to F37 once I will have some free time to spend on it.

PR #1561, opened 2023-04-07
add stable_version as output in PackagesResource

arthurzam:
We at Gentoo plan to use this great infra to find which of our packages are outdated. One very small gripe with current existing APIs that I can't get the latest stable versions across all projects under Gentoo distribution, since it returns only latest version, which include pre-releases.

As suggested in this issue (and I also had similar idea myself before I found this issue), I've added an extra small property which holds the latest stable version. If there are none (for example all versions are pre-release), it will return None, which would be converted to `null` in JSON. I think this is the best way to represent the current state to the consumers.

This is my first contribution to this project, so if I mistook something I'll be happy to fix and improve. Also, the listed example update might not be the best, since for that case latest version is also stable, so if requested I can select better example (I think PyPI/moto is good one).

Resolves: https://github.com/fedora-infra/anitya/issues/1033

### DIFF
--- a/anitya/api_v2.py
+++ b/anitya/api_v2.py
@@ -125,7 +125,8 @@ class PackagesResource(MethodView):
                         "name": "python-requests"
                         "project": "requests",
                         "ecosystem": "pypi",
-                        "version": "2.28.1"
+                        "version": "2.28.1",
+                        "stable_version": "2.28.1"
                     }
                 ],
                 "items_per_page": 25,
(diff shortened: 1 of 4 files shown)


Zlopez:
Hi, I'm glad to see folks from Gentoo contributing to this project, thanks for the PR.

Zlopez (approved):
This looks good and the tests works. Adding a new field to API calls isn't backwards incompatible change, so we can release this in next minor release. :-)

PR #1562, opened 2023-04-11
Fix the arbitrary URL jumping of /login.

lu4nx:
Hi, the /login handler has a vulnerability. 

The `next' parameter does not check the range of the target URL for the jump, so it can jump to any URL, for example

/login/?next=https://fedoraproject.org

I fixed it.


Zlopez (approved):
Thanks for the contribution, that is a nasty vulnerability I missed.

Zlopez:
Could you add a test for the use case you prevented?

lu4nx:
> Could you add a test for the use case you prevented?

Hello, I added ;-)

PR #1580, opened 2023-05-09
Renovate Badge

lmilbaum:


Zlopez:
Thanks for the PR, I didn't know there is a badge for renovate.

lmilbaum:
> Thanks for the PR, I didn't know there is a badge for renovate.

Sure. NP. 
BTW, does it make sense to have both dependabot and renovate activated in parallel?

Zlopez:
@lmilbaum It doesn't, the dependabot should be disabled for Anitya, but we sometimes get security updates that are recommended by GitHub itself and they are using dependabot for that.

PR #1598, opened 2023-05-23
pylint anitya/admin.py

lmilbaum:


lmilbaum:
@Zlopez Would you mind taking a look at the CI error. I can't figure out what have changed in line 269 which produced the coverage error.

Zlopez:
@lmilbaum Looking at it and it seems that the line wasn't covered by tests before and when you changed it it started complaining.

You can either create a test for it or we can merge it as it is.

lmilbaum:
> @lmilbaum Looking at it and it seems that the line wasn't covered by tests before and when you changed it it started complaining.
> 
> You can either create a test for it or we can merge it as it is.

Thanks for looking into it. I'd prefer to merge it as it is if that works for you.

PR #1599, opened 2023-05-24
Sort the list of distributions in the project view

dfandrich:
This make it neater and easier to find a distro when many supply a
package.

### DIFF
--- a/anitya/templates/project.html
+++ b/anitya/templates/project.html
@@ -179,7 +179,7 @@
                   {% endif %}
                 </tr>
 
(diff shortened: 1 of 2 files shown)


dfandrich:
Neither of those two CI failures look like they have anything to do with this PR.

Zlopez:
I agree that both CI issues seems unrelated to your changes and doesn't happen on latest master. So probably a temporary issue.

Zlopez (approved):

PR #1601, opened 2023-05-25
Rewrite the Python versioning scheme implementation

### LINKED ISSUE #1586: Bug in PEP440 sorting?
I've recently discovered a problem when jupyterlab released stable version 4.0.0. The problem is that their versioning upstream looks roughly like this:

```
3.6.2
3.6.3
4.0.0a0
4.0.0a1
4.0.0b0
4.0.0b1
4.0.0b2
4.0.0rc0
4.0.0rc1
4.0.0
```

I think this is the correct versioning as described in: https://peps.python.org/pep-0440/#pre-releases

I've tried all different versioning schemes in anitya but none of them resulted in the same order as upstream project.

* Semantic has all the pre-releases at the very bottom of the list. At least version 4.0.0 is recognized as the newest.
* RPM works better but with that version 4.0.0 is older than alpha 0 and betas are newer than RC.
* With PEP440, all pre-release groups are sorted correctly (RCs, betas, alphas) but alpha 9 is newer than alpha 33 and version 4.0.0 is the oldest version 4 available.

Because I don't expect many pre-releases, I switched to semantic versioning to have the latest released version on top but the rest is not correctly sorted.

What is the best to do here?

Zlopez:
The current solution didn't work well with Anitya and introduced plenty of new issues. This rewrite tries to solve all the issues and hopefully not creating new ones.

Fixes #1586


anitya/tests/lib/versions/test_python.py
```diff
@@ -283,3 +290,79 @@ def test_normalization_ws(self):
...
+            "4.0.0a6",
+            "4.0.0a4",
+            "4.0.0a37",
+            "4.0.0a36",
+            "4.0.0a35",
+            "4.0.0a34",
+        ]
+        python_versions = []
+        for version in versions:
+            python_versions.append(python.PythonVersion(version))
```
hroncok, line 318: This might be a tad easier to read:

```suggestion
        python_versions = [python.PythonVersion(v) for v in versions]
```

PR #1703, opened 2024-01-11
Fix pre-commit workflow

Zlopez:
Running the pre-commit locally didn't caused the issues that I see in the github actions, maybe removing the versions constraints can fix that.


Zlopez:
I couldn't figure out why this is reporting false positives in pylint as there isn't anything directly causing this. I will keep the workflow as it's not blocking and maybe it will fix itself with some dependency upgrade.

PR #1708, opened 2024-01-28
Support odd-numbered pre-release version filtering

jdknight:
This commit introduces a filter pattern `!odds[:n]`, which can be used to help identify pre-release versions for packages using odd-numbered versions for development releases. This convenience filter can be used over having to add explicit exceptions for each odd version entry for projects using odd-numbered development releases.

For example, if a project configures a pre-release filter using the value `!odds`, any odd-numbered minor version for a project will be flagged as a pre-release. For example, versions `1.0` and `1.2` will be considered stable versions, as where `1.1` is considered a pre-release version.

### DIFF
--- a/anitya/lib/versions/base.py
+++ b/anitya/lib/versions/base.py
@@ -26,6 +26,8 @@ import re
 from datetime import datetime
 from typing import Optional
 
+import packaging.version
+
 from anitya.lib.exceptions import InvalidVersion
 
 #: A regular expression to determine if the version string contains a 'v' prefix.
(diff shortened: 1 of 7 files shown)


Zlopez:
@jdknight Thanks for the PR, this is really useful. I'm just not sure if this makes sense to be supported by calendar versioning.

jdknight:
> I'm just not sure if this makes sense to be supported by calendar versioning.

For sure.

My assumption was that if the `!odds` pre-release filter was applied to an unsupported scheme type, it would be more of an invalid filter configuration state (so one may expect unexpected results, such as odd months flagged as pre-releases).

That being said, and if desired, I could tweak the base implementation to accept an `odds` argument:

```
def prerelease(self, odds: Optional[bool]=True) -> bool:
    ...
        if odds and pre_release_filter.startswith("!odds"):
    ...
```

And adjust `anitya/lib/versions/calver.py` to invoke:

```
    return super().prerelease(odds=False)
```

instead. I'm good to use any other suggestions if another approach is desired.

Issue #1627, opened 2023-06-10
Sort search by a "clickable" 'name' 'homepage' 'backend' 'latest version'?


romulasry:
See:
https://release-monitoring.org/projects/search/?pattern=open

So you can sort ascending/descending with the above items?

Zlopez:
That would be a good addition to Anitya.

Freedisch:
I think, we could have a small up-and-down icon for each section to trigger the sort   
![up-and-down-arrow](https://github.com/fedora-infra/anitya/assets/82499435/f7171b08-3180-4a54-840f-3cca4b9ab5a3)

Zlopez:
That is the idea of how this should look, just adding small up-and-down icon for each section.

PR #1724, opened 2024-03-08
[Feature] Implement sorting feature for project listings

Freedisch:
# Description
Fix Issue: #1627 
This PR introduces a dynamic sorting feature to the project listings page, enabling users to sort the projects by Name, Homepage, or Backend in both ascending and descending orders.

## Demo

https://github.com/fedora-infra/anitya/assets/82499435/b2d2d251-34ff-4c41-9c45-00aa89461c6e


Zlopez:
This looks good and I like how it doesn't touch database or don't need javascript as well.

Could you just add tests for the lines you added, so the diff-cover is happy?

Zlopez:
Could you also add this to search page https://github.com/fedora-infra/anitya/blob/master/anitya/templates/search.html as this is what was originally requested in #1627

Zlopez:
And this is also missing a news file, see https://anitya.readthedocs.io/en/latest/contributing.html#release-notes how to add one.

Freedisch:
@Zlopez, Well received, I'm on it

Freedisch:
@Zlopez, I completed the PR

anitya/ui.py
```diff
@@ -181,6 +185,24 @@ def projects():
     )
 
 
+def sort_projects(projects, sort):
```
Zlopez, line 188: I'm thinking if it wouldn't be better to move this to `db/models.py` as it will make more sense to have this as part of [Project object](https://github.com/fedora-infra/anitya/blob/master/anitya/db/models.py#L192)

Freedisch: I agree, this will make code much cleaner, I'm doing rn

Zlopez:
Looks good to me now, could you just squash the commits?

Zlopez (approved):

PR #1725, opened 2024-03-08
Fix AttributeError and HTTP Configuration for Fedora Login Authentication

NyuydineBill:
### Changes Made:
1. Replaced the usage of the deprecated `base64.encodestring` with `base64.encodebytes` to resolve the `AttributeError` encountered during authentication with Fedora login in the Flask application.
2. Updated the Flask application configuration to use HTTP instead of HTTPS to resolve server errors encountered during login.

#### Replacement of `encodestring`:
The `AttributeError` was raised due to the usage of the deprecated `encodestring` attribute of the `base64` module. In Python 3, `encodestring` has been deprecated and replaced with `encodebytes`. This change ensures compatibility with Python 3 and resolves the error.

#### HTTP Configuration Update:
The Flask application was configured to run over HTTPS by default, causing server errors when logging in due to mismatched protocols. To address this issue, the `SERVER_NAME` configuration variable in the `wsgi.py` file was modified to use HTTP instead of HTTPS.

### Steps Taken:
1. Located the file causing the `AttributeError` (`storage.py` within the `social_sqlalchemy` package).
2. Opened the `storage.py` file in a text editor.
3. Replaced instances of `base64.encodestring` with `base64.encodebytes`.
4. Modified the `wsgi.py` file to set the `SERVER_NAME` configuration variable to use HTTP (`http://localhost:5000`).
5. Tested the changes locally to ensure successful authentication with Fedora login and resolved server errors during login.
6. Created this pull request to contribute the fixes back to the original repository.

---
Additionally I have created a file named Auth_ERROR_README.md where you can see full steps to resolve this


server_change.py
Zlopez, line 1: This change shouldn't be done here, but better [here](https://github.com/fedora-infra/anitya/blob/master/ansible/roles/anitya-dev/files/anitya.service), as we only want dev environment to run as HTTP not anything else.

Auth_ERROR_README.md
```diff
@@ -0,0 +1,47 @@
...
+### Description
+The AttributeError occurs because the `encodestring` attribute has been deprecated in Python 3 and replaced with `encodebytes`. Attempting to use `encodestring` results in an AttributeError. Furthermore, server errors occur when logging in due to the Flask application being configured to run over HTTPS.
+
+### Cause
+The root cause of the AttributeError is the usage of the deprecated `encodestring` attribute instead of `encodebytes` in the affected file. The server errors occur because the Flask application is configured to run over HTTPS by default.
+
+### Solution
+To resolve the AttributeError, replace `encodestring` with `encodebytes` in the affected file (`storage.py` within the `social_sqlalchemy` package). To resolve the server errors when logging in, change the Flask application configuration to run over HTTP instead of HTTPS.
+
+### Steps to Resolve
```
Zlopez, line 15: This looks like the issue is with old `social_sqlalchemy` dependency. I will look into that.

Zlopez (changes requested):
Looking at the guide it would be better to fix the issue itself than adding a workaround documentation. I will look at the `social_sqlalchemy` issue and try to fix that.

Also the documentation is in https://github.com/fedora-infra/anitya/tree/master/docs, adding new README will be just confusing.

Auth_ERROR_README.md
Zlopez: I would recommend to remove this file altogether and I need to fix the dev environment as I wasn't aware that there are issues.

NyuydineBill: Okay, I just removed them

Zlopez:
I think we can close it as there are no files added, but you can write your solution to issue you created so it's documented.

PR #1728, opened 2024-03-08
Fix the authentication issue

### LINKED ISSUE #1727: Correctly handle uuid value returned from Fedora Account System
Anitya is currently expects either `None` or `hex` as `UUID` values returned during authentication. But when updating the vagrant environment I encountered an issue when I received the `UUID` object directly.

Zlopez:
Anitya is currently expects either `None` or `hex` as `UUID` values returned during authentication. This is solving the case when `UUID` object is returned directly.

Fixes #1727

### DIFF
--- a/anitya/db/models.py
+++ b/anitya/db/models.py
@@ -915,6 +915,8 @@ class GUID(TypeDecorator):
         """
         if value is None:
             return value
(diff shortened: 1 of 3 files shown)


Zlopez:
The error is not related to changes merging as it is.

Issue #1421, opened 2022-07-29
Automatic filling of details when creating project


Zlopez:
# Description
When creating a new project in Anitya, it would be nice if some of the fields were pre-filled by sane defaults based on the project backend. For instance in PyPi backend we know, that the homepage url will be probably something like `https://pypi.org/project/<name>` or we can read this information from the project itself.

## Requirements
* New button on the Add project page `Fill in details`, which will just call the API call and based on the backend it will fill in the details about the project (homepage, version_url, versioning scheme etc.)
* New API endpoint which will allow the above

## How this will make Anitya better
This will help us preventing duplicate projects and makes it easy for users to fill in new projects.

NyuydineBill:
Hi @Zlopez  I am Nyuydine Bill an applicant at outreachy, is this issue still valuable, let me work on it

Zlopez:
Hi, yes this is still open, feel free to assign the issue to yourself and work on it.

NyuydineBill:
Okay!

NyuydineBill:
@Zlopez  I am currently working on this 
please can i get the backend API url?

Zlopez:
The API is documented [here](https://anitya.readthedocs.io/en/stable/api.html). The URL is `https://release-monitoring.org/api/v2`, but for this issue you need to create a new API call which will do what you need.

PR #1741, opened 2024-03-15
Enhanced Project Creation Functionality: Dynamic Default ValuesNew feature

NyuydineBill:
This PR enhances the project creation and editing functionality by introducing dynamic default values based on the selected backend. With this enhancement, users can now enjoy a more streamlined experience when creating or editing projects, as the form fields are automatically filled with relevant default values tailored to the chosen backend.

Key Changes:

Added a button to fill details in project creation page
Implemented a Flask route '/get_default_values_for_backend' to dynamically generate default values for project creation/editing.
Extended the Python backend logic to calculate and return default values such as project name, homepage URL, version URL, version scheme, version pattern, version prefix, pre-release filter, and version filter.
Integrated JavaScript functionality to handle AJAX requests and populate form fields with default values fetched from the server based on the selected backend.
This enhancement improves usability and efficiency, contributing to a more user-friendly experience for contributors and project maintainers alike.

This is to solve issue #1421


news/1723.other
Zlopez, line 1: This file shouldn't be here.

docs/contributing.rst
Zlopez, line 1: This file shouldn't be here.

anitya/templates/project_new.html
```diff
@@ -13,7 +13,7 @@ <h1>{{ context }} project</h1>
 <form method="POST" action="{{ url_for('anitya_ui.new_project') }}" >
 {%- elif context == 'Edit' -%}
 <form method="POST" action="{{ url_for('anitya_ui.edit_project', project_id=project.id) }}" >
-{%- endif %}
+  {%- endif %}
```
Zlopez, line 16: Don't change indentation of existing files. It will report changes in lines that don't have any changes.

anitya/templates/project_new.html
```diff
@@ -104,16 +105,50 @@ <h1>{{ context }} project</h1>
...
+            <span class="list-group-item">
+                <p id="info_field_title"></p>
+                <ul id="info_field_list"></ul>
+            </span>
+        </div>
     </div>
-  </div>
+</div>
+
+<script src="https://ajax.googleapis.com/ajax/libs/jquery/3.5.1/jquery.min.js"></script>
```
Zlopez, line 120: `jquery` should be already imported, no need to import it again.

anitya/ui.py
```diff
@@ -567,6 +566,44 @@ def new_project():
         400,
     )
 
+from flask import Flask, request, jsonify
+@ui_blueprint.route('/get_default_values_for_backend', methods=['GET'])
+def get_default_values_for_backend():
```
Zlopez, line 571: This isn't what I wanted in this feature.
It should be a new API endpoint, which will scan the project on the server and fills in the values obtained from the backend.

Like for example, user adds URL `https://pypi.org/project/<name>` and you fill the new_project page with all the data obtained by the API call.

To see what I have in mind, look at the https://www.protondb.com/contribute and how they can find the game based on the AppID, URL etc.

Zlopez (commented):
It isn't what I have in mind, but you can use it as base for future work.

anitya/ui.py
NyuydineBill: OKkay if I get You well, then the user should enter the url, then the API is called and retrieves data related to that url entered, 
And which field is responsible for the url here

I will try that out
image

NyuydineBill: could this be a sample url 
https://release-monitoring.org/project/250162/

NyuydineBill: I've read through the docs and I was looking for the API endpoint that could query through the projects and return their data maybe as JSON, but happens not to be there.
Any idea please?

NyuydineBill:
@Zlopez  i am Querying now and getting some data from the API, but is multiple.
Should I consider just the last(most recent) object?


image

NyuydineBill:
@Zlopez  this is the data I am retrieving now, the last object
image

Zlopez:
Looking at the PR, I think that you didn't understood what I want or how complex this feature is. Let me try to explain the workflow and the changes needed for that.

WORKFLOW:

1. When clicking on `Add project` menu entry the user will be redirected to a new page, which will ask him for URL to project he want's to add and click 'Add' button on page
2. User gets redirected to current `Add project` page prefilled with data obtained from project's page

Changes needed:
1. New page that will allow user to put the URL
2. New Flask view that will be called when submitting the URL, this will try to recognize the backend (the best way to do it is probably using domain name checking based on the backends - will probably need some new constant in each backend)
3. New method in each backend (or at least in those where it's possible) that will obtain data from the URL provided
4. The redirect to current `Add project` page with the data obtained in previous step (this will be probably part of the new Flask view)

As you can see there is no need for new `javascript` code or change in any existing HTML template. I will update the description of the issue as well, so it's more clear what I want.

Zlopez:
Now I can see that what I described is actually different feature than the one that is in #1421. My bad in this case. But it's still not what you implemented. The new button on `Add project` page is correct, but how you handle it is not correct. You should have new API endpoint that will return sane defaults based on the backend chosen by user.

For example, if user provides `name` and says it's `PyPi` backend, it will fill in the ecosystem, project homepage, versioning schema etc.

The defaults need to be added to each backend. For inspiration how to handle it, you should look at  `Test check` button.

NyuydineBill:
Okay on it

Zlopez:
Closing due to inactivity.

PR #1748, opened 2024-03-26
Fixing Python virtualenv in development

anshikavashistha:
This is with reference to the [issue](https://github.com/fedora-infra/anitya/issues/1513)


Zlopez:
@anshikavashistha Plenty of tests started failing after the change, could you look into that?

Zlopez:
If you look at the failing tests the issue is obvious `anitya_vestion` doesn't have `get_distribution` method.

anshikavashistha:
@Zlopez How should I solved the failure if `anitya_version` doesn't have `get_distribution` method?

Zlopez:
It's just the string that contains the same information as `pkg_resources.get_distribution`.

anshikavashistha:
@Zlopez I agree with you. Yes that's right `pkg_resources` has `get_ditsribution` method but `anitya_version` don't have.
What should be the alternative to this method for `anitya_version`?

Zlopez:
`anitya_version` should be the same as what is returned by `pkg_resources.get_distribution('anitya').version`. But it needs to be tested first. I'm not sure if we already used poetry at the time this ticket was open. Just try it and see.

anshikavashistha:
Sure @Zlopez I try it and tests it

anshikavashistha:
@Zlopez I have tried to test and run `anitya_version` .Here some of the tests are successful but one of them tox-format is not running.

anshikavashistha:
@Zlopez Sure I will try to format the code as mentioned in the documentation and then try to run

Zlopez:
Closing due to inactivity.

PR #1762, opened 2024-04-19
Fixing python virtualenv

anshikavashistha:


anshikavashistha:
@Zlopez Have a look on this PR?
I have some fault in my VMware due to which I would not be able to make changes in that current PR.

anitya/lib/plugins.py
```diff
@@ -19,7 +19,7 @@
 """Module handling the load/call of the plugins of anitya."""
 
 import logging
-
+import importlib
```
Zlopez, line 22: This import isn't necessary.

createdb.py
```diff
@@ -2,6 +2,7 @@
 # -*- coding: utf-8 -*-
 """createdb"""
 import os
+import importlib
```
Zlopez, line 5: This import isn't necessary.

Zlopez (changes requested):
Some things are still not correct. Also you have a conflict in this PR

PR #1772, opened 2024-05-03
enable deleting distro mapping with slash in package name

arthurzam:
Since we at gentoo plan to start add a lot of mapping for gentoo packages, I'm making sure everything works well with our naming scheme. In gentoo, every package is behind a category, for example `dev-lang/python`. As stated in https://github.com/fedora-infra/anitya/issues/965, this results in inability to easily delete the mapping. The only existing workaround to do it is to edit it, put another "simple" value, and only then delete.

So why does it happen? The flask URL for that expected a single path item, but because of this value, we get 2. So by using `path:` flask annotation, we teach it expect at least one path item, but also multiple. I've tested it by booting a local dev instance, loading from dump, and making myself an admin in postgres, after which I started to play and verify it started to work again. I've also added a test for that thing, since I do agree this might be weird in code, so better test expected behavior.

Sadly this solution isn't enough for cases where the mapping starts or ends with a slash. In this case the browser automatically sanitizes and removes "duplicate" slash in path. But workaround I gave above still works.


Zlopez:
Thanks for the contribution, this is a nice solution to the problem.

Zlopez (approved):

Issue #1900, opened 2025-05-24
You are going to be blocked you from scraping MetaCPAN.org


ranguard:
Hi,

Please could you review your bot, this is traffic for 24 hours hitting MetaCPAN.org

![Image](https://github.com/user-attachments/assets/c0a80667-19dc-418d-934c-7d56db9c72cb)

You are not remembering permanent redirects and are causing load we do not want to handle. 

We will be blocking your scraper tomorrow - please update this ticket once the issue has been rectified and we will re-enable access.

Please also review if scraping is the best approach or if information published by https://pause.perl.org/pause/query?ACTION=pause_04about#indexer would be suffichent

Kind regards

MetaCPAN team

opoplawski:
What's the status here?  Seems like most of the CPAN version checks are failing.

Zlopez:
The patch with the fix is merged, I just need to find time to release new Anitya version.

ranguard:
> [@ranguard](https://github.com/ranguard): [#1904](https://github.com/fedora-infra/anitya/pull/1904) was released, reducing the proportion of requests that will result in a redirection. Is this sufficient for MetaCPAN to unblock Anitya?

Hi All,

Thanks for taking action.

I have removed the block, any issues please email noc @ metacpan.org

All the best

MetaCPAN team

ranguard:
I think this has been resolved - permanently this time!

Please email again if not

PR #1904, opened 2025-06-09
Look for CPAN homepages under https://metacpan.org/dist/

mavit:
This avoids a redirect from https://metacpan.org/release/…, saving traffic for both us and MetaCPAN.

Relates to #1900.


Zlopez (changes requested):
The changes are looking good. Could you just add news file? See https://anitya.readthedocs.io/en/latest/contributing.html#release-notes for more info.

Zlopez:
I will try to release and deploy the new version as soon as possible.

PR #1938, opened 2025-09-12
GNU backend - use ftpmirror

### LINKED ISSUE #1935: Use https://ftpmirror.gnu.org/ for GNU backend
Image

Justinzobel:


Zlopez (changes requested):
So this is missing few things:

* Passing tests
* News file (see https://anitya.readthedocs.io/en/latest/contributing.html#release-notes)

Zlopez:
> I have no idea how to write tests, but I am happy to add something to the News file.

Just look at the failing tests and how they are written, I assume it's just changing the url that is being accessed in the tests.

satmandu:
> > I have no idea how to write tests, but I am happy to add something to the News file.
> 
> Just look at the failing tests and how they are written, I assume it's just changing the url that is being accessed in the tests.

As a long-time Ghostbusters aficionado, my understanding of how to handle Zuul complaining is to just lie to it.

Having said that, these Zuul test failures seem quite opaque, and a simple PR like this tripping them indicates a configuration issue either upstream or somewhere else, no?

I also can't make heads or tails of what to do to try to fix the errors looking at https://fedora.softwarefactory-project.io/zuul/buildset/ff8c0c7c9da24302bcc4461ec2df60a6 .
There are no logs to see there by default...

Zlopez:
@satmandu The logs in zuul are kept only for few days, this explains why the logs aren't there anymore. Doing another run should show the errors. Even running the tests locally should show you what is failing.

The reason why they don't pass is that there is change in code and no change in tests. So they just don't get the correct output for the change. This is how unit tests should work. So failing is correct in this case.

PR #1963, opened 2025-10-29
Make it more obvious to use the Gitea backend for Forgejo/Codeberg

mavit:
I tried to follow the instructions at https://anitya.readthedocs.io/en/latest/contributing.html#docker-podman but couldn't get it working, so these changes are untested.

### DIFF
--- a/anitya/lib/backends/gitea.py
+++ b/anitya/lib/backends/gitea.py
@@ -27,7 +27,7 @@ class GiteaBackend(BaseBackend):
     with gitea api.
     """
 
-    name = "Gitea"
+    name = "Gitea/Forgejo"
     examples = [
         "https://codeberg.org/freedroid/freedroid-src.git",
         "https://codeberg.org/forgejo/forgejo",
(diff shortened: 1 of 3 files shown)


anitya/lib/backends/gitea.py
```diff
@@ -27,7 +27,7 @@ class GiteaBackend(BaseBackend):
     with gitea api.
     """
 
-    name = "Gitea"
+    name = "Gitea/Forgejo"
```
Zlopez: This is not a good idea, as it will invalidate the projects that are already have Gitea set as backend and they will not recognize the backend.

mavit: I'll look into doing something like the following instead:

```html
<select id="backend" name="backend" required="" tabindex="3">
...
  <option value="Gitea">Gitea/Forgejo</option>
...
</select>

Zlopez: Doing that in HTML should be OK

PR #1982, opened 2025-11-27
Format tox instructions for copy/paste

abitrolly:
Makes it easier to copy the line to run tests. There are two problems, though.

1. Looks like `poetry run tox` is not enough
```
$ poetry run tox
The "poetry.dev-dependencies" section is deprecated and will be removed in a future version. Use "poetry.group.dev.dependencies" instead.
Creating virtualenv anitya-L2WRRFYm-py3.13 in /root/.cache/pypoetry/virtualenvs
Command not found: tox
```

2. Docs say "All tests must pass. All new code should have 100% test coverage.", but it is not clear how to see the coverage.


Zlopez (approved):
Thanks for fixing the warning and updating the docs.

PR #2008, opened 2026-02-11
fix(backends): correctly report error when package missing on PyPI

### LINKED ISSUE #2001: Correctly report error when package no longer exists on pypi.org
Currently the pypi.org backend returns 404 when the package no longer exists, we should take this in consideration and report error when chekcing for new version. This will help us with spam that is sometimes created on PyPi.

Currently the check just reports that no new version were found.

YatesMold:
Proposed Change Currently, the PyPI backend logic does not distinguish between a 404 Not Found (missing package) and other errors. This PR updates get_version and get_versions to explicitly check for response.status_code == 404.

    Implementation Details

        Added a check for status code 404 immediately after the API call.

        Raises AnityaPluginException with a clear message ("Package not found on PyPI") if triggered.

    Verification Verified locally by creating a test script that attempts to fetch a non-existent package. Confirmed that the AnityaPluginException is now raised specifically for the 404 case.

    Fixes #2001

### DIFF
--- a/anitya/lib/backends/pypi.py
+++ b/anitya/lib/backends/pypi.py
@@ -45,6 +45,10 @@ class PypiBackend(BaseBackend):
         except Exception as err:  # pragma: no cover
             raise AnityaPluginException(f"Could not contact {url}") from err
 
+        # Handle missing packages explicitly
+        if req.status_code == 404:
+            raise AnityaPluginException(f"Package not found on PyPI: {url}")
+
         # Not modified
         if req.status_code == 304:
             return None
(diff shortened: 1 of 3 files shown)


Zlopez (changes requested):
The change looks good, only few things:

1. test_pypi_404 should be part of test_pypi, no need to create new test file
2. News file is missing, see how to create it https://anitya.readthedocs.io/en/latest/contributing.html#release-notes

Zlopez (approved):

PR #2010, opened 2026-02-15
Fix Bitbucket backend by migrating to REST API (#2000)

### LINKED ISSUE #2000: Fix the bitbucket backend
There was probably some change on bitbucket.org page as the projects with this backend are failing on `ode: no upstream version found. - https://bitbucket.org/odedevs/ode/downloads?tab=tags - class="name">([^<]*[^tip])</td`

See https://release-monitoring.org/project/2532/ as example.

YatesMold:
Fixes #2000.

**Reasoning:**
Bitbucket has recently transitioned to a heavily JavaScript-rendered Single Page Application (SPA). As a result, the raw HTML returned to our `requests.get` call is essentially empty and no longer contains the version tags, causing the regex `class="name">([^<]*[^tip])</td` to fail with `no upstream version found`.

**Changes in this PR:**
* Rewrote `BitBucketBackend` to stop using HTML regex scraping.
* Migrated the version fetching logic to use the official Bitbucket REST API v2.0 (`/2.0/repositories/{owner}/{repo}/refs/tags`).
* Added `?sort=-target.date` to ensure tags are returned in the correct order.
* Updated `test_bitbucket.py` to use `unittest.mock.patch`, preventing live network calls and properly testing the new API JSON parsing logic.
* Added a news fragment.


Zlopez (changes requested):
The bitbucket changes are looking good, but the PyPi changes shouldn't be part of this PR. I assume there is some bad rebase as part of this PR.

Zlopez (approved):

PR #2014, opened 2026-02-24
Strip leading and trailing whitespace from project searches

gene1wood:
For UI-based searches (excluding API searches), when the "exact" query option is disabled, any leading or trailing whitespace is trimmed before wildcard characters are added to the search term.

This prevents accidental copy-paste whitespace from breaking searches. For example, if a user pastes a project name with extra spaces, the system avoids generating a query like "*nginx *", which will return no matches, and instead searches using the trimmed value so relevant results are returned.


gene1wood:
The `fedora/check` failed on the `Get tox envlist config` task in the `fi-tox-python312` environment because of

> Unable to connect to the server: net/http: TLS handshake timeout

which I suspect was some ephemeral network issue during the test run. If a maintainer can re-trigger the test, I suspect it will pass.

Zlopez (changes requested):
Just one small change, otherwise this looks good.

Zlopez (approved):

PR #2023, opened 2026-03-31
feat: add latest_version_created_on to api/v2/versions

rtdegraaf:
add latest_version_created_on to the API call api/v2/versions so that the scraped date of the latest version can be retrieved by using the API.

### DIFF
--- a/anitya/api_v2.py
+++ b/anitya/api_v2.py
@@ -564,6 +564,7 @@ class VersionsResource(MethodView):
 
             {
                 "latest_version": "2.12",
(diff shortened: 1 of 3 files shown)


Zlopez (changes requested):
This looks good and it's a nice addition. Only thing that is missing is the news file (see https://anitya.readthedocs.io/en/latest/contributing.html#release-notes).

FYI I just want to add that `created_on` isn't the same as release date of the version. The version could be retrieved later than it was released.

Zlopez (approved):

PR #2034, opened 2026-05-14
Fix missing escapes in tooltip help

jamesjer:
With the custom backend, the hover text for the Regex field says (in part):
```
For example regex '(d)_(d)_(d)' applied to '1_2_3' will convert version to '1.2.3'.
```
This PR fixes escapes in the python code so they appear in the hover text.  If you prefer, a raw string could be used instead.


Zlopez (changes requested):
Could you add a news file, so it's visible in changelog? See https://anitya.readthedocs.io/en/latest/contributing.html#release-notes for more info

jamesjer:
Can do. How is this?

Zlopez (approved):

PR #2054, opened 2026-07-09
Fix for sort only working on the current page

### LINKED ISSUE #1778: Name, homepage, backend sorting is working only for the current page
The new sorting by name, homepage and backend is not working for whole dataset, but only the current page, which is not helpful.

The behavior could be reproduced on https://stg.release-monitoring.org/projects/?page=1

sheikhlimon:
Fixes #1778

The sort was happening in Python after the DB paginated, so each page sorted itself. Pushed the sort into the SQL query so it runs against the whole dataset before pagination.

Also fixed the pagination links dropping the `sort` param.

### DIFF
--- a/anitya/db/models.py
+++ b/anitya/db/models.py
@@ -541,9 +541,9 @@ class Project(Base):
             return None
 
     @classmethod
-    def all(cls, session, page=None, count=False):
+    def all(cls, session, page=None, count=False, sort=None):
         """all"""
-        query = session.query(Project).order_by(sa.func.lower(Project.name))
+        query = _apply_sort(session.query(Project), sort)
 
         query = _paginate_query(query, page)
 
@@ -635,7 +635,7 @@ class Project(Base):
             return query.all()
 
     @classmethod
-    def search(cls, session, pattern, distro=None, page=None, count=False):
+    def search(cls, session, pattern, distro=None, page=None, count=False, sort=None):
         """Search the projects by their name or package name"""
 
         query1 = session.query(cls)
@@ -665,7 +665,8 @@ class Project(Base):
                 sa.func.lower(Packages.distro_name) == sa.func.lower(distro)
             )
 
-        query = query1.distinct().union(query2.distinct()).order_by(cls.name)
+        query = query1.distinct().union(query2.distinct())
+        query = _apply_sort(query, sort)
 
         query = _paginate_query(query, page)
 
@@ -676,9 +677,7 @@ class Project(Base):
 
     @classmethod
     def sort_projects(cls, projects, sort):
-        """
-        Sorts a list of projects objects based on a sort parameter.
-        """
+        """Sorts a list of project objects based on a sort parameter."""
         sort_attr, _, sort_order = sort.partition("_")
 
         attr_map = {"name": "name", "homepage": "homepage", "backend": "backend"}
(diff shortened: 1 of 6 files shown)


anitya/db/models.py
```diff
@@ -693,6 +692,29 @@ def sort_projects(cls, projects, sort):
...
 
 
+_SORT_ATTRS = {
+    "name": Project.name,
+    "homepage": Project.homepage,
+    "backend": Project.backend,
+}
+
+
+def _apply_sort(query, sort):
```
Zlopez, line 702: Why this is a separate function? This should be modification of existing `sort_projects` method instead as it doesn't make much sense to duplicate it.

`sort_projects` could be also private method as it doesn't need to be called separately (I checked the code and it's only called twice from anitya/ui.py, we can make it part of every `Project.all` and `Project.search` call instead.

anitya/db/models.py
```diff
@@ -541,9 +541,9 @@ def by_name_and_ecosystem(cls, session, name, ecosystem):
             return None
 
     @classmethod
-    def all(cls, session, page=None, count=False):
+    def all(cls, session, page=None, count=False, sort=None):
         """all"""
-        query = session.query(Project).order_by(sa.func.lower(Project.name))
+        query = _apply_sort(session.query(Project), sort)
```
Zlopez, line 546: I don't think we need to call this here, just call `Project.sort_projects` after calling `Project.all`.

anitya/db/models.py
```diff
@@ -665,7 +665,8 @@ def search(cls, session, pattern, distro=None, page=None, count=False):
                 sa.func.lower(Packages.distro_name) == sa.func.lower(distro)
             )
 
-        query = query1.distinct().union(query2.distinct()).order_by(cls.name)
+        query = query1.distinct().union(query2.distinct())
+        query = _apply_sort(query, sort)
```
Zlopez, line 669: I don't think we need to call this here, just call `Project.sort_projects` after calling `Project.search`.

anitya/ui.py
```diff
@@ -379,29 +377,32 @@ def projects_search(pattern=None):
...
+            db.session, pattern=get_extended_pattern(pattern), page=page, sort=sort
         ):
             if proj not in projects:
                 projects.append(proj)
         projects_count = models.Project.search(
             db.session, pattern=get_extended_pattern(pattern), count=True
         )
+        # Two paginated SQL queries are merged here, so SQL-side ordering
+        # no longer holds across the union; sort the combined list instead.
+        projects = models.Project.sort_projects(projects, sort)
```
Zlopez, line 394: We don't need this here at all, let's `Project.search` do the sort for us.

anitya/ui.py
```diff
@@ -176,12 +176,9 @@ def projects():
     except ValueError:
         page = 1
 
-    projects = models.Project.all(db.session, page=page)
+    projects = models.Project.all(db.session, page=page, sort=sort)
     projects_count = models.Project.all(db.session, count=True)
 
-    if sort:
```
Zlopez, line 182: I agree this should go away. `Project.all` should already sort it.

Zlopez (commented):
Thanks for the PR, but this would be better implemented as improvement of existing method in Project class instead of separate function.

Zlopez:
It seems the behavior of mergify changed. In the past it was waiting for the approval from reviewer before merging anything, but it seems that is no longer the case.

@sheikhlimon Could you address the comments in separate PR?

Zlopez:
Hopefully I fixed the mergify config to prevent this from happening in the future.

sheikhlimon:
> It seems the behavior of mergify changed. In the past it was waiting for the approval from reviewer before merging anything, but it seems that is no longer the case.
> 
> @sheikhlimon Could you address the comments in separate PR?

@Zlopez Done! I've addressed the review comments in #2056. PTAL when you have a chance. Thanks!

PR #2056, opened 2026-07-10
Fold _apply_sort into Project.sort_projects

sheikhlimon:
## Summary
Follow-up to #2054.

Updated `Project.sort_projects` to handle both SQL queries and Python lists, removed `_apply_sort` and `_SORT_ATTRS`, and routed `Project.all` and `Project.search` through it.

I kept the final `Project.sort_projects(projects, sort)` call in the extended-search merge. Removing it breaks the ordering because the results of two independently sorted `Project.search` calls are merged in Python.

### Note
If you'd prefer a different approach for the merge, I'm happy to follow up with another implementation.

### DIFF
--- a/anitya/db/models.py
+++ b/anitya/db/models.py
@@ -543,7 +543,7 @@ class Project(Base):
     @classmethod
     def all(cls, session, page=None, count=False, sort=None):
         """all"""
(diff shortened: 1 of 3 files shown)


Zlopez (approved):
LGTM, thanks for the work.

PR #2060, opened 2026-07-26
Add pre-creation project version validation

### LINKED ISSUE #1940: [Feature Request] Require any new project to have a successful test before being added
Projects should require a successful check before they can be added.

That way, people aren't adding things that don't work to start with and eating up resources of all kinds for no reason.

sheikhlimon:
Fixes #1940
                          
This adds a validation step when adding a new project to prevent broken configurations from being saved to the database.                                          
                                                         
Previously, projects were saved first, and if they were broken (e.g., bad URL or backend couldn't find versions), they would sit in the database and waste resources during the periodic check service.           
  
Now, we perform a dry-run test using the backend plugin *before* saving. If no versions can be found, the project is rejected immediately with a helpful error message.

Assisted-By: Claude Opus 4.6

### DIFF
--- a/anitya/lib/utilities.py
+++ b/anitya/lib/utilities.py
@@ -148,6 +148,7 @@ def check_project_release(project, session, test=False):
         project.logs = "No new version found"
 
     if test:
+        session.close()
         return upstream_versions[::-1]
 
     if upstream_versions:
--- a/anitya/lib/utilities.py
+++ b/anitya/lib/utilities.py
@@ -148,7 +148,6 @@ def check_project_release(project, session, test=False):
         project.logs = "No new version found"
 
     if test:
-        session.close()
         return upstream_versions[::-1]
 
     if upstream_versions:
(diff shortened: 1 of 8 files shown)


anitya/lib/utilities.py
```diff
@@ -148,7 +148,6 @@ def check_project_release(project, session, test=False):
         project.logs = "No new version found"
 
     if test:
-        session.close()
```
Zlopez, line 151: Why this is being removed?

news/1940.feature
Zlopez, line 1: This should be just name of the issue being fixed. Not a paragraph. Explanation should be in pull request.

Zlopez (changes requested):
Thanks for the PR, just few small things.

anitya/lib/utilities.py
sheikhlimon: my bad! looks like that line got dropped by accident while refactoring

news/1940.feature
sheikhlimon: done. I'll keep that in mind. The dog(agent) probably went a bit overboard

Zlopez (approved):

PR #2061, opened 2026-07-26
Fix transparent navbar on scroll

sheikhlimon:
## Summary
Fixes a UI bug where the page content was visible *through* the top navigation bar when scrolling down.  
                                                         
When dark mode was added previously, the old solid background class was removed from the navbar but wasn't replaced. This PR simply adds `bg-body-tertiary` back to the navbar so it has a solid background that correctly adapts to both light and dark mode.

**Before**
image

**After**
image

### DIFF
--- a/anitya/templates/master.html
+++ b/anitya/templates/master.html
@@ -61,7 +61,7 @@
   >
     <header>
       <!-- Static navbar -->
-      <nav class="navbar navbar-expand-lg fixed-top">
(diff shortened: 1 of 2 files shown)


Zlopez:
This is a really simple fix. Could you add a news file as well?

sheikhlimon:
> This is a really simple fix. Could you add a news file as well?

Sure. The transparent navbar was bugging me for some time.

Zlopez (approved):

PR #2064, opened 2026-08-11
Apply version filters to historically stored versions

### LINKED ISSUE #1816: Give maintainers the ability to remove tags
Sometimes an upstream will add a weird git tag, which sorts to the top and blocks subsequent tags from being considered as the newest version.  I've had this happen lots of times.  When it happens, I add a rule to filter out the weird tag, but it is too late.  The new filter does not apply to the already fetched tags, so the weird tag stays at the top.

I just had this happen with the cvc5 project: https://release-monitoring.org/project/268448/.  Upstream added tags from the previous cvc4 project, and the higher version numbers for cvc4 are now hiding new releases of cvc5.  (That seems like a bug to me, by the way.  The cvc4 tags are of the form "cvc4-<version>".  They do not contain the string "cvc5" and are also not a bare version number, so why are they considered versions of cvc5?)

Please give maintainers a way to do something about this.  For example, give us a button to throw away all fetched tags.  Then we can press the "Full Check" button again to refetch the tags and apply any new filters.

In the meantime, if you could do that manually for cvc5, that would be very helpful.

### LINKED ISSUE #2062: libvpl not respecting Version Filters
https://release-monitoring.org/project/242664/

`2021;2022;2023` is in the Version Filters field, and yet the current version reported is 2023.4.0 instead of 2.17.0.

sheikhlimon:
### Summary 
Fixes #2062
Fixes #1816

This PR applies the `version_filter` logic directly inside `get_sorted_version_objects`, so older, already fetched versions are retroactively hidden from public view when they match an active filter.

### Changes

- Added `.strip()` to filter values. Accidental spaces in filters such as `2021; 2022` no longer break filtering.
- Filtered versions remain visible in the admin table, but are visually dimmed and marked with a "Filtered" badge to make their inactive status clear.
- The "Latest version" shown at the top of the page now updates to the newest unfiltered version.

https://github.com/user-attachments/assets/eac89e9e-e112-489f-bd16-1023b9de5e19

### DIFF
--- a/anitya/db/models.py
+++ b/anitya/db/models.py
@@ -404,8 +404,41 @@ class Project(Base):
         backend = BACKEND_PLUGINS.get_plugin(self.backend)
         return backend.get_version_url(self)
 
-    def get_sorted_version_objects(self):
+    def is_version_filtered(self, version):
+        """
+        Check if a given version string matches the project's version filter.
+
+        This checks both the raw version string and the version string with the prefix
+        applied, if the project has a version_prefix configured.
+
+        Args:
+            version (str): The version string to check.
+
+        Returns:
+            bool: True if the version is filtered, False otherwise.
+        """
+        if not self.version_filter:
+            return False
+
+        filter_list = self.version_filter.split(";")
+        prefixed_version = (
+            f"{self.version_prefix}{version}" if self.version_prefix else version
+        )
+
+        for filter_str in filter_list:
+            filter_str = filter_str.strip()
+            if not filter_str:
+                continue
+            if filter_str in version or filter_str in prefixed_version:
+                return True
+
+        return False
+
+    def get_sorted_version_objects(self, ignore_filter=False):
         """Return list of all version objects stored, sorted from newest to oldest.
+
+        Args:
+           ignore_filter (bool, optional): If True, do not apply version filters. Defaults to False.
 
         Returns:
            :obj:`list` of :obj:`anitya.lib.versions.Base`: List of version objects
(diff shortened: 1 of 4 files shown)


sheikhlimon:
Hey @zlopez!

While adding the "Filtered" badges, I noticed the Edit/Flag/Delete buttons on the project page were pretty misaligned due to some older Bootstrap float-based styling. I went ahead and cleaned up the `project.html` layout using basic Flexbox, which fixes the alignment and simplifies the existing markup.

The UI cleanup is technically a little outside the scope of this bug fix, so if it feels better to keep the PR focused on the filter logic, I'm happy to revert the style changes.

Zlopez:
> Hey @Zlopez!
> 
> While adding the "Filtered" badges, I noticed the Edit/Flag/Delete buttons on the project page were pretty misaligned due to some older Bootstrap float-based styling. I went ahead and cleaned up the `project.html` layout using basic Flexbox, which fixes the alignment and simplifies the existing markup.
> 
> The UI cleanup is technically a little outside the scope of this bug fix, so if it feels better to keep the PR focused on the filter logic, I'm happy to revert the style changes.

Could you rather move it to separate PR, so it will be easier to rollback if there is any issue?

sheikhlimon:
> Could you rather move it to separate PR, so it will be easier to rollback if there is any issue?

done! I moved the UI cleanup into a separate PR #2065 

Edit: Added a screen recording for reference. The recording was originally included with the combined PR.

Zlopez (approved):
Nice addition to the code.

Zlopez:
You will need to resolve the conflict when #2065 was merged.

sheikhlimon:
> You will need to resolve the conflict when #2065 was merged.

Sure, looking into it

sheikhlimon:
PTAL. I've resolved the conflict. Thank you for reviewing and merging them!

PR #2072, opened 2026-09-04
fix: correct pagination count on the users management page

### LINKED ISSUE #928: No pagination on user management screen
There is no pagination on user management screen, you can only see the first 30 entries. You still can find specific entry by using filters, so this is just inconvenient.

sundaram123krishnan:
Fixes: #928

### DIFF
--- a/anitya/admin.py
+++ b/anitya/admin.py
@@ -546,15 +546,17 @@ def browse_users():
         if active is not None:
             users_query = users_query.filter(models.User.active == active)
 
+        cnt_users = db.session.scalar(
+           select(func.count()).select_from(users_query.subquery())
+        )
+
         if offset > 0:
             users_query = users_query.offset(offset)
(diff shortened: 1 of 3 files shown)


anitya/admin.py
```diff
@@ -546,15 +546,17 @@ def browse_users():
         if active is not None:
             users_query = users_query.filter(models.User.active == active)
 
+        cnt_users = db.session.scalar(
```
Zlopez, line 549: Why this was moved?

sundaram123krishnan: Because the limit query is also being added to cnt_users, so it always displays the limit instead of showing the entire count for users.

sundaram123krishnan: This means that there is no pagination being displayed, as only `limit` number of items are shown

Zlopez (approved):

Issue #2074, opened 2026-09-11
Calendar version pattern is strange - breaks sorting


abitrolly:
I am not sure why there are 0M and MM specifiers at https://release-monitoring.org/static/docs/user-guide.html#version_pattern

The project https://release-monitoring.org/project/392250/ uses calendar YY.MM.DD version, but the sorting is strange.

14.11 appears between 15.5.1 and 15.03

Version|
-- |
16.7.4 |
15.5.2 |
15.5.1 |
14.11 |
15.03 |
15.03.05 |
15.03.17 |
15.04.19 |
15.10.03 |

abitrolly:
The code that does the comparison and fails lives here https://github.com/fedora-infra/anitya/blob/master/anitya/lib/versions/calver.py

The [documentation](https://release-monitoring.org/static/docs/user-guide.html#version_pattern) is too generic and doesn't explain the behavior why 14.11 > 15.03

> The version pattern is used to define how the calendar version should be parsed and how they will be sorted.

Also the docs doesn't explain why 0es matter. They may matter for yeas, but for everything else what is the point?

> Traditional, incremented version numbers are 0-based, whereas date segments are 1-based, and the short and zero-padded years are relative to the year 2000. 

Maybe an example should say what is so special about

sheikhlimon:
I took a look and I believe there are two things happening here.

The `YY.MM.DD` pattern expects months/days without leading zeros, so `15.03` fails to parse and is intentionally placed in the unparseable group at the bottom. So `14.11` being above `15.03` is expected with the current pattern.

The ordering within that unparseable group is a separate bug, though. Those versions should still be sorted consistently, but the comparison logic in `calver.py` can produce an incorrect ordering when both versions fail to parse. PR #2077 should fix that and I've also updated the docs to make the padding behavior clearer.

cc @Zlopez

abitrolly:
> The YY.MM.DD pattern expects months/days without leading zeros

Why? What's the point? Just parse MM as is, convert to int and do tuple comparison. Then zeroes won't matter.

sheikhlimon:
> > The YY.MM.DD pattern expects months/days without leading zeros
> 
> Why? What's the point? Just parse MM as is, convert to int and do tuple comparison. Then zeroes won't matter.

The main place where the padding distinction matters is for delimiter-less patterns like `YYYY0M0D`, where fixed widths are needed so dates like `2024115` aren't ambiguous.

[...]

PR #2077, opened 2026-09-12
Fix CalendarVersion sorting for unparseable versions

sheikhlimon:
## Summary 

Relates to #2074

### What

- Fix `CalendarVersion.__lt__` when both versions fail to parse.
- Fall back to the parent `Version` comparison instead of incorrectly returning `True`.

### Why

When a project's version pattern doesn't match the upstream version format, the version may fail to parse.

Previously, if both versions failed to parse, `CalendarVersion` could consider both `A < B` and `B < A` to be true. This caused invalid versions to be sorted incorrectly.


abitrolly:
> Add a warning to user-guide.rst about strict zero-padded (0M) and non-padded (MM) patterns.

This doesn't explain why are they needed at all.

sheikhlimon:
> > Add a warning to user-guide.rst about strict zero-padded (0M) and non-padded (MM) patterns.
> 
> This doesn't explain why are they needed at all.

I've dropped the doc change from this PR to keep it strictly focused on the `__lt__` fallback fix when versions fail to parse. We can discuss the parser behavior and documentation over in #2074

Zlopez (approved):
This looks like I had a bug in the calendar versioning implementation. Thanks for the fix.

PR #1751, opened 2024-03-27
Adding version pattern to user-guide.rst

vidit-maheshwari:
### Issue

Fixes #1604 

### Description

This pull request adds documentation for the version pattern field in the user guide. The version pattern field is essential for extracting and matching version numbers from project releases, and it allows users to specify the format of version numbers using regular expressions. This documentation clarifies the usage of various pattern tokens such as `YYYY`, `YY`, `MM`, `WW`, `DD`, etc., along with their examples and notes on usage.

### Changes Made

- Added a new section titled "Version Pattern" to the user guide (`user-guide.rst`).
- Provided detailed explanations and examples for each pattern token (`YYYY`, `YY`, `MM`, `WW`, `DD`, etc.).
- Included notes on traditional version numbering, the relative nature of short and zero-padded years, and the mutual exclusivity of weeks with months/days.

### Testing Done

- Reviewed the changes locally to ensure accuracy and clarity.
- Checked for formatting and consistency with existing documentation.

### Additional Notes

- This documentation enhancement aims to improve usability and clarity for users configuring the version pattern field.
- Any feedback or suggestions for improvement are welcome.

---
*Please review and merge this pull request at your earliest convenience. Thank you!*

PR #2078, opened 2026-09-13
user-guide: Define calendar versioning problem

abitrolly:
https://github.com/fedora-infra/anitya/issues/2074#issuecomment-5648500165


abitrolly:
I am still not sure what to do with this Note.

> Traditional, incremented version numbers are 0-based, whereas date segments are 1-based, and the short and zero-padded years are relative to the year 2000. Usage of weeks is usually mutually exclusive with months/days.

Really have no idea what it tries to warn against.

sheikhlimon:
> I am still not sure what to do with this Note.
> 
> > Traditional, incremented version numbers are 0-based, whereas date segments are 1-based, and the short and zero-padded years are relative to the year 2000. Usage of weeks is usually mutually exclusive with months/days.
> 
> Really have no idea what it tries to warn against.

Looking back at #1751 where the table was added, it seems that note was just taken from calver.org. It looks like general CalVer conventions from their site (distinguishing increment counters from dates, and not mixing weeks with months). In `calver.py`, weeks (`WW`) don't seem to be implemented anyway, so the note might just be leftover text that isn't really applicable to Anitya.

On the description change, patterns seem to be needed for any CalVer project so Anitya knows the layout (like `YYYY.MM.DD` vs `YY.MM`) not just delimiter-less dates

abitrolly:
Removed the note and edited text a bit.

sheikhlimon:
looks good to me

PR #2080, opened 2026-09-13
Replace deprecated datetime.utcnow() and arrow.utcnow()

### LINKED ISSUE #2057: Replace deprecated `datetime.utcnow()` / `arrow.utcnow()`
## Summary 

`datetime.datetime.utcnow()` was deprecated in Python 3.12, and `arrow.utcnow()` was deprecated in Arrow 1.3.0. Anitya currently uses both. It may be worth updating these usages to their non-deprecated equivalents.

**Occurrences:**

* `arrow.utcnow()` ×4: `anitya/check_service.py:97,200`, `anitya/db/models.py:265,268`, `anitya/lib/utilities.py:83`
* `datetime.datetime.utcnow()` ×7 (all SQLAlchemy `default=`): `anitya/db/models.py:276,381,724,766,872,1110`

There are also ~25 occurrences in test fixtures.

**Proposed approach:**

* Replace `arrow.utcnow()` with `arrow.now("UTC")`.
* Replace `datetime.datetime.utcnow()` with a non-deprecated equivalent. Since the affected SQLAlchemy columns are tz-naive, it's worth deciding whether to preserve the current column semantics or treat this as part of a broader migration to timezone-aware timestamps.

Happy to send a PR.

sheikhlimon:
Replaces deprecated datetime.datetime.utcnow() and arrow.utcnow() usages across the codebase.

- Adds an internal `_utcnow()` helper in `anitya.db.models` returning naive UTC datetime to preserve existing `sa.DateTime` column semantics
- Updates defaults for `Project.created_on`, `ProjectVersion.created_on`, `ProjectFlag.created_on`, `Run.created_on`, and `ApiToken.created` to use `_utcnow`
- Updates `create_version_objects()` to set `created_on=_utcnow()`
- Replaces `arrow.utcnow()` with `arrow.now("UTC")` across `check_service`, `utilities`, migration `1bf8aead6179_add_check_times`, and test fixtures
- Updates date calculation in `test_models.py` to use `datetime.now(timezone.utc).date()`
- Adds towncrier news fragment `2057.dev`

Fixes #2057


anitya/db/models.py
```diff
@@ -262,18 +267,22 @@ class Project(Base):
...
+        index=True,
     )
 
     updated_on = sa.Column(
         sa.DateTime,
         server_default=sa.sql.functions.now(),
         onupdate=sa.sql.functions.current_timestamp(),
     )
-    created_on = sa.Column(sa.DateTime, default=datetime.datetime.utcnow)
+    created_on = sa.Column(sa.DateTime, default=_utcnow)
```
Zlopez, line 285: Wouldn't this work with `arrow.now("UTC").datetime`

anitya/db/models.py
```diff
@@ -378,7 +387,7 @@ def create_version_objects(self, versions):
                     version=version if isinstance(version, str) else version["version"],
                     prefix=self.version_prefix,
                     pre_release_filter=self.pre_release_filter,
-                    created_on=datetime.datetime.utcnow(),
+                    created_on=_utcnow(),
```
Zlopez, line 390: Same as above.

anitya/db/models.py
```diff
@@ -755,7 +764,7 @@ class ProjectVersion(Base):
         primary_key=True,
     )
     version = sa.Column(sa.String(50), primary_key=True)
-    created_on = sa.Column(sa.DateTime, default=datetime.datetime.utcnow)
+    created_on = sa.Column(sa.DateTime, default=_utcnow)
```
Zlopez, line 767: Same as above.

anitya/db/models.py
```diff
@@ -797,7 +806,7 @@ class ProjectFlag(Base):
     reason = sa.Column(sa.Text, nullable=False)
     user = sa.Column(sa.String(200), index=True, nullable=False)
     state = sa.Column(sa.String(50), default="open", nullable=False)
-    created_on = sa.Column(sa.DateTime, default=datetime.datetime.utcnow)
+    created_on = sa.Column(sa.DateTime, default=_utcnow)
```
Zlopez, line 809: Same as above.

anitya/db/models.py
```diff
@@ -902,9 +911,7 @@ class Run(Base):
     error_count = sa.Column(sa.Integer)
     ratelimit_count = sa.Column(sa.Integer)
     success_count = sa.Column(sa.Integer)
-    created_on = sa.Column(
-        sa.DateTime, default=datetime.datetime.utcnow, primary_key=True
-    )
+    created_on = sa.Column(sa.DateTime, default=_utcnow, primary_key=True)
```
Zlopez, line 914: Same as above.

anitya/db/models.py
```diff
@@ -1141,7 +1148,7 @@ class ApiToken(Base):
     __tablename__ = "tokens"
 
     token = sa.Column(sa.String(40), default=_api_token_generator, primary_key=True)
-    created = sa.Column(sa.DateTime, default=datetime.datetime.utcnow, nullable=False)
+    created = sa.Column(sa.DateTime, default=_utcnow, nullable=False)
```
Zlopez, line 1151: Same as above.

Zlopez (commented):
Just one thing that I want to clear.

anitya/db/models.py
sheikhlimon: I just gave it a try, but it looks like it might not work.

When I use `arrow.now("UTC").datetime`, the test suite starts failing with `TypeError`s (can't compare offset-naive and offset-aware datetimes). It seems the original `datetime.utcnow()` returned a naive datetime, and since the database columns don't use `timezone=True`, introducing an aware datetime here breaks the downstream comparisons.

I used `_utcnow()` just to preserve that naive behavior.

Edit: Ah, wait, let me correct myself. I just double-checked and I was slightly off about the test suite.

The tests actually *do* pass with `arrow.now("UTC").datetime` because SQLite (which the test suite uses) silently normalizes the aware datetimes on insertion.

However, the core issue remains for production: since Anitya uses PostgreSQL and those columns are `sa.DateTime` (without `timezone=True`), passing aware datetimes violates the offset-naive semantics the old code relied on. The `_utcnow()` helper just keeps it strictly naive to match the exact behavior of `datetime.utcnow()` and avoid any psycopg2 warnings.

Let me know if you'd still prefer to go the arrow route, though!

Zlopez:
In that case, let me approve it.

Zlopez (approved):

PR #2084, opened 2026-09-16
Raise exception on HTTP errors when parsing versions by regex

sheikhlimon:
As discussed in https://github.com/fedora-infra/anitya/issues/2075#issuecomment-5697841590, backends using `get_versions_by_regex` currently swallow HTTP errors and pass 4xx/5xx error pages directly to the regex engine, resulting in a generic `no upstream version found` failure.

This PR adds `req.raise_for_status()` to catch and surface the actual HTTP error. As suggested by @Zlopez, it raises a standard `AnityaPluginException`, so it plugs non-intrusively into the existing error logging instead of the regex failure.


anitya/lib/backends/__init__.py
```diff
@@ -381,6 +381,15 @@ def get_versions_by_regex(url, regex, project, insecure=False):
         # Not modified
         if req.status_code == 304:
             return []
+        
+        try:
+            req.raise_for_status()
+        except requests.exceptions.HTTPError as err:
+            _log.debug("%s ERROR: %s", project.name, str(err))
```
Zlopez: We should use `_log.error` for this.

Zlopez (commented):
Also missing the news file.

anitya/lib/backends/__init__.py
sheikhlimon: Thanks!! updated

Zlopez (approved):

PR #2085, opened 2026-09-17
Add version details with prerelease flags to API v2

### LINKED ISSUE #1448: Get release date with API
If you visit a project on release-monitoring.org, you see the list of all versions. This list contains the date, on which the version was released. But I have not found a way to get this date with the API in the [documentation](https://release-monitoring.org/static/docs/api.html). Could this be added?

sheikhlimon:
When clients query the v2 API for project versions, they previously only received a flat list of version strings. This required making additional assumptions to figure out whether a version was a prerelease or exactly when it was created. 

This update introduces a `version_details` field that exposes the full version objects directly in the response. I've also extracted the data generation into a shared helper to avoid duplicating the same logic across the GET and POST endpoints.

Fixes #1448


anitya/api_v2.py
```diff
@@ -49,6 +49,20 @@ def _page_validator(arg):
...
 
+def _get_version_details(project):
+    """
+    Helper function to get the version details for a project.
+    """
+    return [
+        {
+            "version": str(v_obj.version),
+            "created_on": v_obj.created_on.isoformat() if v_obj.created_on else None,
+            "prerelease": v_obj.prerelease(),
```
Zlopez: Just a small thing. Would be probably better to name it `pre_release`, so it's easier to read.

Zlopez (changes requested):
Just a small change.

anitya/api_v2.py
sheikhlimon: Updated. Thanks!

Zlopez (approved):

PR #2086, opened 2026-09-17
Fix version validation logic and highlight unparseable versions

### LINKED ISSUE #2081: Report failed parsing in `Test Check` dialog
From https://github.com/fedora-infra/anitya/issues/2074#issuecomment-5648058468:

> The YY.MM.DD pattern expects months/days without leading zeros, so 15.03 fails to parse and is intentionally placed in the unparseable group at the bottom. 

Errors should never pass silently (c) https://peps.python.org/pep-0020/

Would be nice to:

1. See failed versions highlighed in red with explanation
2. Make error messages actionable - propose a schema that doesn't fail

sheikhlimon:
The base version validation previously swallowed parse failures silently, making it difficult to debug broken patterns.

To fix this, the parsing logic now explicitly defines an `is_valid` property on all version classes. This surfaces unparseable versions by highlighting them in the project versions UI table, making it clear to maintainers when upstream tags do not fit the chosen scheme.

Fixes #2081

screenshot-2026-09-17_22-45-01


anitya/api_v2.py
Zlopez, line 1: I'm not sure if this wouldn't be confusing for user. New field in API that is looking like some versions on the projects are not valid.

Zlopez:
Also it seems that we didn't understood each other. I don't want this to happen during the test check, it's nice to show this, but it could end up with really long modal window, also the change in api_v2 is not great. I rather wanted it to be shown in the version table, just to highlight the versions that are not sortable with current version scheme, which could be done when the page is being rendered.

sheikhlimon:
> Also it seems that we didn't understood each other. I don't want this to happen during the test check, it's nice to show this, but it could end up with really long modal window, also the change in api_v2 is not great. I rather wanted it to be shown in the version table, just to highlight the versions that are not sortable with current version scheme, which could be done when the page is being rendered.

Ah, that makes total sense! You're right - if a project has hundreds of unparseable tags, dumping them all into the test check modal would completely break the layout.

To fix this, I can revert the changes to the api_v2 endpoint and the test check modal entirely. Instead, I can just use the new is_valid property directly in the Jinja template to visually highlight the unsortable versions in the main "Versions" table on the project page.

Does this align with what you had in mind? If so, I'll go ahead and update the PR

Zlopez:
Yeah, that is what I had in mind.

sheikhlimon:
> Yeah, that is what I had in mind.

Updated. Here's how it looks in the table now:

screenshot-2026-09-17_22-45-01

Zlopez (approved):
This looks good now, let's merge it.

PR #2087, opened 2026-09-19
Prevent `openh264` from installing in dev image

abitrolly:


abitrolly:
Do we really needs ancient Python versions to bloat the image for a one-of-a-kind web service? As I understand there are no libs and it is not a framework. And even if it is framework, people can build own images to port it to other OS and Python versions.

```
# Install helpful development packages (for tests and setup)
RUN dnf install -y git poetry python3-tox which python3.10 python3.11 nmap-ncat
```

```
anitya-base:latest                  20ffbde2acb9       2.62GB          671MB    U   
```

sheikhlimon (approved):
LGTM

sheikhlimon:
> Do we really needs ancient Python versions to bloat the image for a one-of-a-kind web service? As I understand there are no libs and it is not a framework. And even if it is framework, people can build own images to port it to other OS and Python versions.

We install multiple versions because `tox` uses them to run our test suite across different environments.

However, I just checked `tox.ini`, and we no longer test against `python3.10`. Probably safe to remove it.

sheikhlimon:
On a second note, do we even need the `upgrade` line here?

https://github.com/fedora-infra/anitya/blob/f04146d48f913d61da3a311955106e3aa377a9da/Containerfile.dev#L9

Since we are already pulling the latest `fedora:44` base image, it seems like running a full OS upgrade on top of it just bloats the container and causes random hangs

cc @Zlopez

Zlopez:
> On a second note, do we even need the `upgrade` line here?
> 
> https://github.com/fedora-infra/anitya/blob/f04146d48f913d61da3a311955106e3aa377a9da/Containerfile.dev#L9
> 
> Since we are already pulling the latest `fedora:44` base image, it seems like running a full OS upgrade on top of it just bloats the container and causes random hangs
> 
> cc @Zlopez

This is there just to be sure we are having the latest versions of everything.

PR #2091, opened 2026-09-19
Place <a> inside <label>

abitrolly:
Otherwise it is invalid markup and is unclickable.

See https://release-monitoring.org/project/392250/edit

I can not test that the fix works, because dev environment is broken for me - https://github.com/fedora-infra/anitya/issues/2090 If it doesn't, then maybe put some info icon there..


abitrolly:
The link will probably come up escaped https://wtforms.readthedocs.io/en/3.2.x/fields/#wtforms.fields.Label

abitrolly:
May need to add `|safe` if escape occurs - https://jinja.palletsprojects.com/en/stable/templates/#working-with-automatic-escaping

anitya/templates/functions.html
```diff
@@ -1,8 +1,10 @@
 {% macro render_field_in_row(field, after="", docs="") %}
     <td>
-      {%- if docs -%}<a href="{{ docs }}">{%- endif -%}
-      {{ field.label }}
-      {%- if docs -%}</a>{%- endif -%}
+      {%- if docs -%}
+        {{ field.label('<a href="{{docs}}" target="_blank">' + field.label.text + '</a>') }}
+      {%- else -%}
+        {{ field.label }}
+      {%- endif -%}
```
sheikhlimon: I was testing this out locally and it seems we might run into a couple of issues with this specific approach. It looks like WTForms automatically escapes the HTML passed into `field.label()`, and Jinja won't interpolate the `{{docs}}` variable when it is inside a literal string, so the link ends up rendering as raw text.

I was actually thinking about your comment regarding adding an info icon. That aligns perfectly with modern accessibility guidelines, since wrapping labels in links can sometimes conflict with the label's default click behavior.

What do you think about bypassing `field.label` entirely and rendering a Bootstrap badge next to the text instead? Here is a suggestion that uses some inline flex styling to ensure the badge doesn't wrap to a new line on narrow tables. Let me know if you think this works!

```suggestion
      {%- if docs -%}
          <span style="display: inline-flex; align-items: center; white-space: nowrap;">
            <label for="{{ field.id }}" style="margin-bottom: 0; margin-right: 4px; white-space:
  nowrap;">{{ field.label.text }}</label>
            <a href="{{ docs }}" target="_blank" aria-label="Documentation" title="Documentation"
  class="text-decoration-none">
              <span class="badge rounded-pill bg-secondary">?</span>
            </a>
          </span>
        {%- else -%}
          {{ field.label }}
        {%- endif -%}
```

abitrolly:
I can not test it right now, but it looks like a better solution.

Zlopez (approved):

PR #2092, opened 2026-09-19
Fix missing static files in docker dev environment

### LINKED ISSUE #2090: Development Environment missing static files with `docker`
https://anitya.readthedocs.io/en/latest/contributing.html#development-environment

After going through
```
docker compose -f container-compose.yml up -d
docker compose -f container-compose.yml exec -it anitya-web bash -c "poetry run python3 createdb.py"
```

CSS, JS and other static resources are 404.

sheikhlimon:
Fixes #2090.

The local development environment (`make up`) returned 404s for static assets because the host repository bind mount (`.:/app:z`) overshadowed the `node_modules` directory built into the container image.

This adds a named volume for `/app/anitya/static/node_modules` to preserve the built frontend assets while still allowing host mounts for the rest of the application code. It also corrects the path in `Containerfile.dev` to build the assets in `anitya/static` where the application expects them.


Containerfile.dev
```diff
@@ -43,8 +43,8 @@ RUN mkdir -p /etc/fedora-messaging
 COPY .container/web/config.toml /etc/fedora-messaging
 
 # Install javascript dependencies
-COPY anitya/static /app/static
-RUN pushd static && npm install && popd
+COPY anitya/static /app/anitya/static
+RUN pushd anitya/static && npm install && popd
```
abitrolly, line 47:
```suggestion
COPY anitya/static $STATIC_FILES_PATH
RUN (cd $STATIC_FILES_PATH && npm install)
```

Containerfile.dev
abitrolly, line 3:
```suggestion

ENV STATIC_FILES_PATH=/srv/anitya_static

```

news/2090.dev
```diff
@@ -0,0 +1 @@
+Fix missing static files in local Docker dev environment by mounting node_modules as a named volume.
```
abitrolly, line 1:
```suggestion
Fix missing static files in local Docker dev environment by moving them out of volume shadow.
```

abitrolly:
Also need to set `static_folder` (https://flask.palletsprojects.com/en/stable/api/#flask.Flask) from ENV here:

https://github.com/fedora-infra/anitya/blob/f04146d48f913d61da3a311955106e3aa377a9da/anitya/app.py#L44

sheikhlimon:
@abitrolly thanks for the suggestions!

I was looking at the idea of moving the static folder and modifying `app.py` to route via an env variable, but I think modifying the application code might be a bit out of scope for this specific PR.

I'd also be a little worried that moving the static folder just for local docker setups might create a difference between how the app runs locally and how it runs in production. since the named volume handles the docker shadowing automatically, it feels like the simplest option for now since it keeps the environments identical.

maybe we could track the dynamic routing setup in a separate issue? so we can keep this PR scoped

abitrolly:
@sheikhlimon production setup doesn't use containers - https://anitya.readthedocs.io/en/latest/admin-guide.html - and containers are already bloated with development dependencies, so I wouldn't care about it.

I will do separate PR to test moving static folder works. Why named volume works in this case?

sheikhlimon:
> @sheikhlimon production setup doesn't use containers - https://anitya.readthedocs.io/en/latest/admin-guide.html - and containers are already bloated with development dependencies, so I wouldn't care about it.
> 
> I will do separate PR to test moving static folder works. Why named volume works in this case?

oh my bad, I just meant the python code itself, not the containers.. I just lean towards fixing docker issues inside the docker config rather than modifying the python code to work around it.

as for why the volume works: when compose mounts our local repo (`.:/app`), it completely covers up the `node_modules` folder that the image built. adding the named volume just tells the container daemon to preserve the original `node_modules` folder instead of hiding it... it lets us fix the missing files purely through the compose config without ever having to touch the python code

out of curiosity, what advantage do you think we'd get by moving the static folder?

PR #2094, opened 2026-09-20
Update deprecated Fedora infra Ansible repository links to Forgejo

sheikhlimon:
Updated the hardcoded Pagure links in the admin and contributing docs to point to new Forgejo instance now that the Ansible repo has migrated.


Zlopez (approved):
Thanks for catching this, it will probably pop-up on multiple places after the migration from pagure.io.

PR #2095, opened 2026-09-21
Remove the deprecated Pagure backend

sheikhlimon:
This removes the pagure backend since it no longer has any valid upstream targets to monitor. With pagure.io officially read-only as projects migrate to Forgejo, upstreams can no longer push new tags there for Anitya to detect, and src.fedoraproject.org is strictly a downstream dist-git that we don't monitor anyway.

I've completely stripped out the pagure.py backend implementation, its test suite, and the outdated documentation references to clean out the dead weight.

> [!NOTE]
> This includes an Alembic migration that safely moves any existing pagure projects in the database over to the `custom` backend, generating a valid URL for them so they don't fail validation when the plugin is removed.


anitya/db/migrations/versions/cb692d5e35a9_remove_pagure_backend.py
```diff
@@ -0,0 +1,23 @@
...
+from alembic import op
+
+# revision identifiers, used by Alembic.
+revision = "cb692d5e35a9"
+down_revision = "ebc827e80373"
+
+
+def upgrade():
+    op.execute(
+        "UPDATE projects SET backend = 'custom', version_url = 'https://pagure.io/' || name || '/releases' WHERE backend = 'pagure'"
```
sheikhlimon: for a quick gut-check, since I'm not super familiar with Alembic. I wrote this raw SQL to move the orphaned projects to the custom backend and used string concat to build valid URLs so they don't fail validation. Is this the right approach we want to take here?

anitya/db/migrations/versions/cb692d5e35a9_remove_pagure_backend.py
```diff
@@ -0,0 +1,24 @@
...
+
+from alembic import op
+
+# revision identifiers, used by Alembic.
+revision = "cb692d5e35a9"
+down_revision = "ebc827e80373"
+
+
+def upgrade():
+    op.execute(
```
sheikhlimon, line 16: for a quick gut-check, since I'm not super familiar with Alembic. I wrote this raw SQL to move the orphaned projects to the custom backend and used string concat to build valid URLs so they don't fail validation. Is this the right approach we want to take here?

Zlopez:
I don't think we want to do this until there is no project using the pagure backend at all. We can archive those projects on release-monitoring.org to prevent them from being checked, but I don't want to remove the backend completely as it would cause issues for projects that are already using it.

sheikhlimon:
> I don't think we want to do this until there is no project using the pagure backend at all. We can archive those projects on release-monitoring.org to prevent them from being checked, but I don't want to remove the backend completely as it would cause issues for projects that are already using it.

Makes sense. I had a migration here to forcefully migrate the remaining projects, but I like the idea of archiving them too. Closing this out.

